PR #14, opened 2015-09-30
Add wpdb injector for filtering out 'OPTIMIZE TABLE' queries

onnimonni:
This fixes unnessary load which is caused by OPTIMIZE TABLE queries.
This fix injects some logic which is done before queries are handled to wpdb

### DIFF
--- a/modules/wpdb-injector.php
+++ b/modules/wpdb-injector.php
@@ -1,0 +1,108 @@
+<?php
+/*
+Plugin Name:  WPDB-injector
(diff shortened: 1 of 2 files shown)


ottok:
I wonder how much load increases as every single DB query does now through an additional layer of regex check?

onnimonni:
I didn't do Any comparison but The vagrant box didn't feel slower at all

ottok:
Closing away deprecated issue. The optimize runs are not likely to be harmful and having extra code to filter them out would just add unnecessary complexity.

PR #15, opened 2015-10-07
Redirect 404 uploads in vagrant to production

onnimonni:
I realised that because uploads folder isn't (and shouldn't be) in git
and it can became really big pain when I try to pull production code for
a small change and all you can see is empty spaces for missing images.

This change helps developers so that they won't need to pull all uploads
into their local development.

wp-palvelu-plugin will just redirect missing uploads into production.

This parses the production url from config.yml (if available) so it
needs yaml. Yaml is now included in requirements of this plugin."

This is similiar feature as https://github.com/H1FI/h1-auto-image-download but just for our system.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
     "wordpress", "wp", "wp-palvelu"
   ],
   "require": {
-    "composer/installers": "v1.0.12"
+    "composer/installers": "v1.0.12",
+    "symfony/yaml": "v2.7.5"
   }
 }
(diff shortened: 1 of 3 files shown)


ottok:
The equivalent feature was already impelemted via https://github.com/Seravo/wp-palvelu-vagrant/commit/0dcbd94284633e9fb7136ad175bd19d5d7bffdbc so this issue can be closed.

PR #19, opened 2016-01-18
Fix incorrect reference to `domain` in `Loader::loadTextdomain`

### LINKED ISSUE #18: PHP Notice:  Undefined variable: domain
The plugin text domain loading method `loadTextdomain` throws said PHP notices on all page loads when the plugin is active (essentially always when used as a mu-plugin).

```

  public static function loadTextdomain() {

    // Load translations first from the languages directory
    $locale = apply_filters( 'plugin_locale', get_locale(), $domain );
    load_textdomain( 
            $domain, 
            WP_LANG_DIR . '/my-plugin/' . self::$domain . '-' . $locale . '.mo' 
    );

    // And then from this plugin folder
    load_muplugin_textdomain( 'wpp', basename( dirname(__FILE__) ) . '/languages' );
  }
```

`$domain` is not defined anywhere. Should those two instances be `self::$domain` or is something missing? Plugin version is 1.3.1 and a fresh `vagrant up` on the `seravo/wordpress` stack.

rask:
`Loader::loadTextdomain` threw PHP notices because the `Loader::domain`
property was referenced without the `self` reference inside the method.

This is fixed and the `domain` property is referenced properly now.

Fixes #18.

### DIFF
--- a/wp-palvelu-plugin.php
+++ b/wp-palvelu-plugin.php
@@ -55,10 +55,11 @@ Class Loader {
   public static function loadTextdomain() {
 
     // Load translations first from the languages directory
-    $locale = apply_filters( 'plugin_locale', get_locale(), $domain );
+    $locale = apply_filters( 'plugin_locale', get_locale(), self::$domain );
+
     load_textdomain(
-            $domain,
-            WP_LANG_DIR . '/my-plugin/' . self::$domain . '-' . $locale . '.mo'
+      self::$domain,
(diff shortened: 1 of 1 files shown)


rask:
Let me know if I should leave out the whitespace and formatting changes from this PR. Adjusted them to match indents and enhance general readability.

anttiviljami:
Thank you for the Fix! :)

PR #23, opened 2016-11-29
Extend Reports page

ottok:
* Improve stats listing with summary bar graphs.
* Report disk size, as it is important for the customer to know their resource usage.
* Show multiple site health checks.

screencapture-wp-palvelu-local-wp-admin-tools-php-1480460967733

### DIFF
--- a/lib/reports-ajax.php
+++ b/lib/reports-ajax.php
@@ -5,4 +5,80 @@ function wpp_report_disk_usage() {
   return $output;
 }
 
-echo json_encode(wpp_report_disk_usage());
+function wpp_report_wp_core_verify() {
+  exec("wp core verify-checksums", $output);
+  array_unshift($output, '$ wp core verify-checksums');
+  return $output;
+}
+
+function wpp_report_git_status() {
+  exec("git -C /data/wordpress status", $output);
+
+  if ( empty($output) ) {
+    return ['Git is not used on this site. To start using it, read our documentation for WordPres developers at <a href="https://seravo.com/docs/">seravo.com/docs</a>.'];
+  }
+
+  array_unshift($output, '$ git status');
+  return $output;
+}
+
+function wpp_report_redis_info() {
+  exec("redis-cli info | grep -e keyspace -e db0", $output);
+  return $output;
+}
+
+function wpp_report_front_cache_status() {
+  exec("curl -ILk ". get_site_url(), $output);
+  array_unshift($output, '$ curl -IL '. get_site_url());
+
+  if ( preg_match('/X-Proxy-Cache: ([A-Z]+)/', implode("\n", $output), $matches) ) {
+
+    switch ($matches[1]) {
+      case 'HIT':
+      case 'EXPIRED':
+        $result = 'Front cache is working correctly.';
+        break;
+
+      case 'MISS':
+        $result = 'Front page is not cached due to cookies or expiry headers emitted from the site.';
+        break;
+
+      default:
+        $result = 'Unable to detect front cache status.';
+        break;
+    }
+
+  } else {
+    $result = 'No front cache available in this WordPress instance.';
+  }
+
+  array_unshift($output, $result, '');
+
+  return $output;
+}
+
+switch ($_REQUEST['section']) {
+  case 'disk_usage':
+    echo json_encode(wpp_report_disk_usage());
+    break;
+
+  case 'wp_core_verify':
+    echo json_encode(wpp_report_wp_core_verify());
+    break;
+
+  case 'git_status':
+    echo json_encode(wpp_report_git_status());
+    break;
+
+  case 'redis_info':
+    echo json_encode(wpp_report_redis_info());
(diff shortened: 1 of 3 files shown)


anttiviljami:
<3

lib/reports-ajax.php
```diff
@@ -0,0 +1,84 @@
...
+  exec("wp core verify-checksums", $output);
+  array_unshift($output, '$ wp core verify-checksums');
+  return $output;
+}
+
+function wpp_report_git_status() {
+  exec("git -C /data/wordpress status", $output);
+
+  if ( empty($output) ) {
+    return ['Git is not used on this site. To start using it, read our documentation for WordPres developers at <a href="https://seravo.com/docs/">seravo.com/docs</a>.'];
```
anttiviljami: Typo here "WordPres"

lib/reports-page.php
```diff
@@ -1,27 +1,122 @@
...
   }
 }
 
 // List months in reverse order with newest first
 rsort($months);
 
 foreach ($months as $month) {
-    echo "<li><a href='?report=$month.html' target='_blank'>$month</a></li>";
+
+  $bar_size = $month['requests'] / $max_requests * 300;
```
anttiviljami: There should probably be a minimum bar width since there's white text displayed inside the bar. 

I suggest if the bar size goes below 40px, we say `width: auto;` which automatically makes the bar fit the text inside of it. We also need the bar div display mode to be `inline-block` for that to work.

anttiviljami: Also, we should round the number to an integer. On tuomioja.org for example a comma was used instead of a full stop as a decimal separator for the output style for some odd reason which made the css not be able to read the value properly.

lib/reports-page.php
```diff
@@ -1,27 +1,122 @@
...
 
 ?>
-</ul>
+  </tbody>
+</table>
+
+
+<h2>Disk usage</h2>
+
+<p>Total size of <code>/data</code> is <?php system("df -h /data | tail -n 1 | cut -f 11 -d ' '"); ?></p>
```
anttiviljami: This is unfortunately incorrect. Docker shows the usage of the host /ldata here. The only way to properly get the size of /data is by using `du` like for other directories below.

lib/reports-ajax.php
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function wpp_report_redis_info() {
+  exec("redis-cli info | grep -e keyspace -e db0", $output);
+  return $output;
+}
+
+function wpp_report_front_cache_status() {
+  exec("curl -ILk ". get_site_url(), $output);
+  array_unshift($output, '$ curl -IL '. get_site_url());
```
anttiviljami: the k flag missing from this line. `-ILk`

lib/reports-page.php
```diff
@@ -1,27 +1,122 @@
...
+
+<h3>Git</h3>
+
+<div id="git_status_loading"><img src="/wp-admin/images/loading.gif"></div>
+<pre id="git_status"></pre>
+
+
+<h2>Cache status</h2>
+
+<h3>Redis</h3>
```
anttiviljami: Should we always show redis cache status on the reports page? Even if `object-cache.php` is disabled? Quite often enabling it doesn't make a net-positive effect on the page load times, so should we encourage its usage like this? It shouldn't be enabled by default either.

We could simply just check to see whether `class_exists( 'WP_Object_cache' )`

lib/reports-page.php
```diff
@@ -1,27 +1,122 @@
...
+
+<div id="redis_info_loading"><img src="/wp-admin/images/loading.gif"></div>
+<pre id="redis_info"></pre>
+
+<h3>Front cache</h3>
```
anttiviljami: Front cache? How about HTTP cache?

lib/reports-page.php
```diff
@@ -1,27 +1,122 @@
-<h1>HTTP Request Reports</h1>
+<h1>Reports</h1>
```
anttiviljami:
```
<div class="wrap"></div>
```

The main wrap used on all WordPress admin pages is missing from this page. Should be wrapped around all content on the page.

lib/reports-page.php
```diff
@@ -1,27 +1,122 @@
...
 // List months in reverse order with newest first
 rsort($months);
 
 foreach ($months as $month) {
-    echo "<li><a href='?report=$month.html' target='_blank'>$month</a></li>";
+
+  $bar_size = $month['requests'] / $max_requests * 300;
+
+  echo "<tr>".
+         "<td><a href='?report=". $month['date'] .".html' target='_blank'>". $month['date'] ."</a></td>".
```
anttiviljami: It isn't that clear this link leads to the goaccess report. It's a really cool feature and should be made available to the client as clearly as possible.

How about adding a button column like this?

screen shot 2016-11-30 at 10 56 22

PR #24, opened 2016-12-11
Add Backups page - first, very basic version

ottok:


modules/backups.php
```diff
@@ -0,0 +1,40 @@
...
+/*
+ * Plugin name: Backups
+ * Description: Enable users to list and create backups
+ * Version: 1.0
+ */
+
+namespace WPPalvelu;
+
+if (!class_exists('Backups')) {
+  class backups {
```
anttiviljami: Should probably be capitalised ?

PR #40, opened 2017-05-24
Fix Travis CI error with opening the wp-login.log file

l3ku:


ottok:
Looking more closely into the issue, it is actually not about the file not found, but that `ini_get('error_log')` returning a path that is write protected.

So the actual fix is to use a different path instead. Implemented in https://github.com/Seravo/seravo-plugin/commit/c674748c07024a4e140769d0c1ef166059b11e45

PR #41, opened 2017-05-26
Use vanilla DOMContentLoaded

k1sul1:
You're not even using jQuery, might as well remove the dependency for it?

We remove jQuery from the client side of themes when we're not using it, which is most of the time nowadays, since it's not 2006 anymore.

### DIFF
--- a/js/instance-switcher.js
+++ b/js/instance-switcher.js
@@ -1,4 +1,4 @@
-jQuery(document).ready(function($){
+document.addEventListener('DOMContentLoaded', function() {
     var links = document.querySelectorAll('#wp-admin-bar-instance-switcher li > a');
     var listener = function(e){
     e.preventDefault();


ottok:
JQuery is always loaded when the WP admin bar is loaded, so it is safe to depend JQuery. Using raw JS is of course better, but some customer corporations still use IE8 or similar which does support 2006 JQuery but perhaps not always the latest EMCA standard. What is the oldest browser you have tested this with?

k1sul1:
Not when you deregister it. Deregistering it competely is the only way I've been able to get rid of it, dequeueing isn't enough because some badly coded plugin will "depend" on it and load it anyway.

DOMContentLoaded doesn't work with IE versions below 9,  but if someone is seriously using IE8 when even IE11 is EOL they deserve broken webpages, but that's just my opinion.  WP core doesn't support IE8 neither should any plugins. With 4.8 that's going to happen.

Another alternative is to move the script to the footer which renders DOMContentLoaded or $(document).ready() unnecessary.

ottok:
We don't support IE8 or anything older, it will not load any of our pages at all due to it having a broken/unsecure HTTPS configuration.

So if you say this is IE9+, then it is fine.

PR #48, opened 2017-06-08
Fix the error when opening docs link 

### LINKED ISSUE #47: Instance switcher javascript code assumes all links in dropdown point to a shadow
At least in situations where there are no shadows, the toolbar dropdown menu for the instance switcher shows a link "Shadows explained at seravo.com/docs". Clicking this item does not take you to seravo.com/docs as one would assume, but instead sets the `wpp_shadow` and `seravo_shadow` cookies with the value `https://seravo.com/docs/deployment/shadows/` and then reloads the page, which in my case resulted in a scary 500 error, presumably because the server was trying to show me a shadow instance that didn't exist or was broken. 

The culprit is instance-switcher.js, which assumes that all items in the dropdown point to shadow-related tasks (i.e. either point to a shadow or exiting one).

You should definitely fix this as it can be quite a scary experience for someone who clicks on that link by mistake and doesn't have pre-existing knowledge of how your shadows work.

elguitar:
Fixes #47 .

As it is documented in instance-switcher.php:140, the last item in the admin bar's instance switcher menu is always the docs link. The loop can be stopped before the link.

### DIFF
--- a/js/instance-switcher.js
+++ b/js/instance-switcher.js
@@ -1,12 +1,13 @@
 document.addEventListener('DOMContentLoaded', function() {
     var links = document.querySelectorAll('#wp-admin-bar-instance-switcher li > a');
     var listener = function(e){
-    e.preventDefault();
     var instance = e.target.getAttribute('href').substr(1);
     if (instance === 'exit') {
+      e.preventDefault();
       document.cookie = "wpp_shadow=;path=/";
       document.cookie = "seravo_shadow=;path=/";
-    } else {
+    } else if (instance.length === 6){
+      e.preventDefault();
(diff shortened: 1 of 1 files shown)


ottok:
Thanks for looking into the issue.

A positional solution is problematic. What if we change the many layout, who will remember to update that -1? Could you find a solution that includes reading the contents of the link, and assuming it is a shadow link only if it is 6 chars long?

Also, think about what comment you but in the commit message and what in code. Change should be described in commit message, while permanent code function should be described in code comments

elguitar:
I confirmed that the shadow hrefs are always 6 chars long.
instance-switcher.php:125:
`'href' => '#' . substr($instance['name'], -6),`

PR #56, opened 2017-07-20
Travis CI: Actually fail if php syntax checking fails

onnimonni:
Current implementation returns true always.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ php:
 
 script:
   # Syntax check all php files and fail for any error text in STDERR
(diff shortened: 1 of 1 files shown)


ottok:
Thanks!

Next time squash your commits. Fixes of fixes within the PR are not relevant for the project history.

ottok:
Similar check is done in https://github.com/Seravo/wordpress/blob/master/scripts/hooks/pre-commit#L69, but the result is evaluated line per line and not just the last line of a list, so that pre-commit hook does not need updating.

onnimonni:
You can squash the commits in github too :)! More information here: https://github.com/blog/2141-squash-your-commits

ottok:
Yes, that is what I did.

PR #60, opened 2017-08-11
Add checks to limit the rights of other than superadmins on multisites

elguitar:


lib/search-replace-ajax.php
```diff
@@ -34,6 +34,8 @@ function seravo_search_replace_set_flags( $options ) {
   }
   if ( $options['network'] === 'true' ) {
     $flags .= '--network ';
+  } else if ( is_multisite() ){
+    $flags .= '--url="' . get_site_url() . '" ';
```
ottok: What if a network super-admin wants to run search-replace for the entire network? Now they won't be able or..?

elguitar: Super-admins can tick "network" box. If it's enabled, the "--url" -flag isn't set (it's in else if).

PR #61, opened 2017-08-15
Remove the permission check from instance switcher (Closes #55)

### LINKED ISSUE #55: Enable instance switcher for non-admins (or provide an easy way to do it)
There was a request for enabling instance switcher for non-admins.

elguitar:
If the permissions are not checked inside instance-switcher, users can run instance-switcher from somewhere else with their own permissions.

### DIFF
--- a/modules/instance-switcher.php
+++ b/modules/instance-switcher.php
@@ -83,11 +83,6 @@ if ( ! class_exists('InstanceSwitcher') ) {
         return;
       }
 
-      // check permissions
-      if ( ! current_user_can( 'activate_plugins' ) ) {
-        return;
-      }
-
       $id = 'instance-switcher';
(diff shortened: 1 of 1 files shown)


ottok:
Isn't this counter-productive to solving #55 ?

elguitar:
If the permission check is not done in instance-switcher's public function add_switcher, but in seravo-plugin's loadAllModules, the users can require instance-switcher.php somewhere else and manage the permissions themselves.
**Example**
somewhere in their code:
```
if ( current_user_can( 'edit_pages' ) ) {
    require_once( PATH . '/folder/folder/instance-switcher.php' );
}
```

elguitar:
Code updated according to the discussion we had with @ottok

PR #70, opened 2017-10-23
Add shadow query string to link if in shadow

k1sul1:
When working in a staging environment, I often have to link to a page inside that environment, but that is surprisingly hard, I have to remember what the query param was, get the instance ID from somewhere, and so on. 

With this, I can just click the big red button, and copy the URL from my browser, or if I'm tech savvy, I can right click the link and copy the address.

Tested and "developed" live @ wp-palvelu.fi instance.


ottok:
Thanks! Code looks good. I will check the UI and test this tomorrow.

PR #74, opened 2017-11-06
Make purge cache capability filterable

k1sul1:
I told a client to hit the "purge cache" button whenever they have problems, and got "What button, where?" as a response. Naturally I don't want to let them manage all the options, but flushing a cache is something they need to be able to do. 

```php
add_filter('seravo_purge_cache_capability', function () {
  return 'edit_posts';
});
```

### DIFF
--- a/modules/purge-cache.php
+++ b/modules/purge-cache.php
@@ -4,6 +4,14 @@
  * Description: Purges the Seravo cache
  */
 
+/** 
+ * Make capability filterable
+ */
+
+function _seravo_purge_capability() {
+  return apply_filters( 'seravo_purge_cache_capability', 'manage_options' );
+}
+
 /**
  * Add a purge button in the WP Admin Bar
  */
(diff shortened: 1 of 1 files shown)


ottok:
What is we just make it by default available to everybody everywhere with capability `edit_posts`?

k1sul1:
That would work for this case, but it would be way more flexible if it's filterable.

ottok:
Allowed now purging for users with edit capability in https://github.com/Seravo/seravo-plugin/commit/06d8703261a94a4f96c36c15d2177e184e26e0c5

In fact, the purge button endpoint was misconfigured, this button has never worked :( I fixed it now and cleaned up an array of smaller bugs around it as well.

I will still leave this issue open if it gets more votes.

ottok:
This PR had another feature I didn't realize at first. The filter is a nice mechanism to also refactor the capability check to one single place, so there is never a mismatch between the three different checks in current code design. Thus I merged the original commit and changed the default role requirement to `Allow any user with 'edit_posts' capabilities to purge cache (related #74)`.

This also closes #82 correctly.

PR #75, opened 2017-11-20
Feature find cruftfiles

simoke:
Cruft file finding feature is now ready for testing.


modules/cruftfiles.php
```diff
@@ -34,13 +34,24 @@ public static function ajax_delete_file() {
...
         $result['success'] = (bool) $unlink_result;
         echo json_encode($result);
       }
       wp_die();
     }
 
+    public static function rmdir_recursive($dir, $recursive) {
+      foreach(scandir($dir) as $file) {
+        if ('.' === $file || '..' === $file) continue;
+        if (is_dir("$dir/$file")) rmdir_recursive("$dir/$file", 1);
```
ottok: These shorthand if-else constructs are hard to read and fail easily if a line is added alter. Please don't use them.

seravo-plugin.php
```diff
@@ -178,6 +178,13 @@ public static function loadAllModules() {
         require_once(dirname( __FILE__ ) . '/modules/backups.php');
       }
 
+      /*
+       * Cruft view for Seravo customers
+       */
+      if ( apply_filters('seravo_show_cruftfiles_page', true) && getenv('CONTAINER') ) {
```
ottok: We could warn about crut also in Vagrant, no need to limit this functions to production/staging containers.

modules/cruftfiles.php
```diff
@@ -26,6 +27,17 @@ public static function load_cruftfiles_page() {
       require_once(dirname( __FILE__ ) . '/../lib/cruftfiles-page.php');
     }
 
+    public static function ajax_delete_file() {
+      if ( isset($_POST['deletefile']) && ! empty($_POST['deletefile']) ) {
+        $file = $_POST['deletefile'];
```
ottok: This looks very dangerous. Should we maybe limit the deletes to files that are in the cruft_found array? Or validate somehow else what is done? Imagine is somebody manages to send '/data/wordpress' to $_POST['deletefile']

ottok:
Ping @simoke

seravo-plugin.php
simoke: Fixed by removing the second condition from if clause

ottok:
Looks good!

Issue #28, opened 2017-03-10
Show a link to Adminer


ottok:
...for easier access. Maybe even bootstrap the session automatically so that users don't need to manually copy-paste the password?

l3ku:
The link to Adminer was added in https://github.com/Seravo/seravo-plugin/pull/80, but the database credentials were not added to the database tools page due to security concerns. Thus, bootstrapping the session automatically could come in handy, as fetching the credentials via SSH causes extra steps for developers and may be too difficult for non-technical users – should we consider implementing it?

Moppa5:
As far as I've found out now, automatically authenticating link to Adminer would require our own implementation for Adminer's credentials() function. After that we could generate working link by adding "&username=db_user" to URL.

Moppa5:
After trying to implement this there were quite many concern about the security. I managed to implement unsafe version by following these https://sourceforge.net/p/adminer/discussion/960418/thread/a0e14f27/ and extending the Adminer class. However the main concern is that how we validate people trying to log on Adminer. Automatically authenticating link would require our own implementation.

Implementing this might be too difficult taking in account the security effort.

ypcs:
You can use WordPress to check if user is authenticated and is either super admin (on network/multisite installations) or admin (on "regular"/single site installations). These users have anyways permission to install plugins/themes that can read the database credentials.

If necessary, we can provide eg. a WordPress filter for site developer to block this feature.

(ie. something like: load WP core, check permissions, provide Adminer modification and then load adminer code)

PR #80, opened 2017-12-21
Database tools

TeemuSuoranta:
New Database page in Tools:

- Only for admins and can be disabled with filter
- Shows database credentials (if this is too risky, I understand)
- Link to Adminer (needs probably more logic with multisites)
- Combines Search and Replace functionality in same page
- Enhances Search and Replace UI
- Shows database size and each table's size
- Localization for these

I wanted to make Adminer button one-click sign-in but found out that it can't be done by just sending credentials to it with POST request (or at least I failed to do it for an hour). Easiest option would be probably to modify the adminer with extension https://www.adminer.org/en/extension/

But then again, there might be other use cases that need those database credentials like using external DB application to connect to database so even if you can create one-click login, they might still be useful.

Linked issues: https://github.com/Seravo/seravo-plugin/issues/28

screen shot 2017-12-21 at 17 31 50


ottok:
Impressive! i will try to review this tomorrow.
:champagne:

ottok:
After discussing this internally, we think that exposing the DB password here is not acceptable currently. Could you replace the credentials with a short help text telling how to see the credentials via SSH and wp-list-env ?

Also a short text that remote access to the database is possible via a SSH tunnel.

ottok:
Here is a snippet I use to figure out what is making the wp_postmeta so big (it often is): `select substring(meta_key, 1, 12) as s, count(*) as c from wp_postmeta group by s order by c desc limit 20;`

Should this or something similar be included? I like the fact that you list the size of the database tables. It would be good to show even more info about database contents. Currently no tool (not even Adminer) is has any automagic feature for this type of debugging.

TeemuSuoranta:
I'll try to make the commit for removing credentials out today. I understand there are many risks for just showing them and it might be better to not show them even if it makes developers take a few extra steps.

For debugging slow database maybe we could separate plain size and debuggin to different sections. The "Database size" could run always but maybe the "Debug database" could run after user has clicked a button.

I'd separate the two because the debugging will probably grow to have many more tests in future. In addition to post_meta the options table could be one thing to check as well. I know one case at least when WPML stuffed one option value with many MB of stuff by faulty string registration.

TeemuSuoranta:
Credentials are now removed and instructions added. I renamed "Database info" already to "Database size" ready for future debug tools.

screen shot 2017-12-22 at 14 56 01

ottok:
What tool are you using to generate the .pot file? In your commit you remove all the plugin desc fields from the end. You new fi.po also introduces many fuzzy lines. If you use the wp-makepot in our Vagrant image, you will automatically get perfect WordPress standard pot files.

TeemuSuoranta:
I was using Poedit that has the ability to generate .pot files as well as edit translations. I wasn't even aware of wp-makepot command as it's not really documented in docs. When I regenerated the .pot with that command as wp-plugin the missing info returned. I also noticed a few warnings in existing translations for minor inconsistencies with translations like original string ending with space and translation not. All this is fixed now.

ottok:
Thanks!

ottok:
You can deploy this with `wp-seravo-plugin-update --dev` on your sites where you want to see it in action.

PR #89, opened 2018-02-13
Add undefined variable

k1sul1:
Fixes Undefined variable: siteurl in some cases. 

Ran into this on a multisite that doesn't require HTTPS using Seravo/wordpress.


ottok:
Thanks!

PR #99, opened 2018-05-17
Fix exit shadow bar not being visible in staging environments

mikkoaf:


l3ku:
Tested this and works as intended. After this change also users that are not logged in can clearly distinguish that they are looking at a shadow instance without viewing the cookies for "wpp_shadow" or "seravo_shadow", and exit the shadow with the click of a link. Good job!

PR #100, opened 2018-05-18
Make logs-tool more readable

nasusa:
- Change log rows hover-effect
- Fix log-views overlay-effect to full-width
- Add log-view-active box under log-view-table
- Add a script when log is clicked, display it in the log-view-active box
- Fix log-viewer.css code
- Fix log-viewer.js indentation

### DIFF
--- a/modules/logs.php
+++ b/modules/logs.php
@@ -165,6 +165,8 @@ if ( ! class_exists('Logs') ) {
         <p><?php _e('Scroll to load more lines from the log.', 'seravo'); ?></p>
       <?php endif;
     endif; ?>
+    <div class="log-view-active">
+    </div>
   </div>
 
   <?php
--- a/js/log-viewer.js
+++ b/js/log-viewer.js
@@ -1,41 +1,42 @@
 'use strict';
 
 jQuery(document).ready(function($) {
-    $(window).on('load', function() {
-        // auto-scroll to bottom of log viewers on page load
-        $('.log-table-view').each(function() {
-        $(this).scrollTop($('table', this).height())
+  $(window).on('load', function() {
+    // auto-scroll to bottom of log viewers on page load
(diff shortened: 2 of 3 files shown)


ottok:
Please add a screenshot as a comment to this PR, as the PR is mostly about UI changes which are not visible directly from the code. Thanks!

nasusa:
Okay!

In the first image you can see logs-tools click-feature. If you click one log, it will appear under the log-view-table. That way it is easier to copy/read.
logs

And in the second image you will see the overlay effect, which was there already but was not fullwidth.
logs_overlay

modules/logs.php
```diff
@@ -165,6 +165,17 @@ public function render_log_view( $logfile, $regex = null, $max_num_of_rows ) {
         <p><?php _e('Scroll to load more lines from the log.', 'seravo'); ?></p>
       <?php endif;
     endif; ?>
+    <div class="log-view-active">
+    </div>
+    <script>
+        // Script for showing clicked row on log-view-active box
+        jQuery(".logrow").click(function () {
```
l3ku: This script should be moved to `log-viewer.js`, as we already have that file for JavaScript required for viewing logs. Also, we want to use the wp_enqueue_script for including scripts and thus keep JS separated from PHP as much as possible.

l3ku:
Couple of small notices: 
- When fixing some functionality in a PR which was created earlier in the same branch, it makes no sense to introduce a new commit, as we want to keep the commit history in the master branch as relevant as possible. Please run `git rebase -i master` so that there is only the original commit `"Make logs-tool more readable"`, but that it contains to functionality added in `"Move javascript from php-file to js-file"`.
- There seems to be some indentation problem in the `log-viewer.js` file as the added JS does not indent properly (see lines 33-34).

After these small fixes I will merge this PR.

PR #111, opened 2018-05-28
Use ready() to make a function available after the document is loaded

nasusa:
Pull request "Update Database-tool layout to WordPress layout" didn't work correctly.

Here is fix to that.

```
'use strict';

jQuery(document).ready(function($) {
});
```


l3ku:
Great! Next time you can push small fixes like this straight to master.

PR #112, opened 2018-05-29
Show stderr when reporting WP data integrity

### LINKED ISSUE #103: Show Error -message if WordPress installation doesn't verify against checksums
Currently the Reports -page doesn't print out standard error messages if WordPress data integrity is compromised. When printing out the results for `wp core verify-checksums` the `stderr` should be printed along side `stdout`.

When WordPress files have not been changed:
nominal

When WordPress installation doesn't verify against checksums:
fix_this

mikkoaf:
Closes #103

### DIFF
--- a/lib/reports-ajax.php
+++ b/lib/reports-ajax.php
@@ -31,7 +31,7 @@ function seravo_report_folders() {
 }
 
 function seravo_report_wp_core_verify() {
-  exec('wp core verify-checksums', $output);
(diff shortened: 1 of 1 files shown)


l3ku:
Good job! Next time, you could maybe name the branch more relevantly: if you fix something, you could use e.g. `fix-reports-verify-checksums`. The current branch name would indicate that you are adding a new page as a feature.

PR #113, opened 2018-05-29
 Include new Seravo PHP_CodeSniffer ruleset for unified code style

l3ku:
This PR includes a ruleset that is created by keeping in mind that we want to primarily
use WP Coding Guidelines, but also have some customizations:
- Use 2 spaces instead of tab indentation
- Multiple assignments don't need to be aligned
- Don't require whitespace around brackets
- No Yoda Conditions
- Don't require class files to be named "class-*"
- Allow usage of PHP exec()
- Don't require all output to be escaped
- Warn about lines that exceed 100 characters

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,4 +26,4 @@ script:
   # Syntax check all php files and fail for any error text in STDERR
   - '! find . -type f -name "*.php" -exec php -d error_reporting=32767 -l {} \; 2>&1 >&- | grep "^"'
   # More extensive PHP Style Check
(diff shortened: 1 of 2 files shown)


ottok:
Related https://seravo.com/coding-wordpress-in-style-with-phpcs/

l3ku:
Rebased on latest master. Also, changed the .travis.yml file to sniff all files instead of only `seravo-plugin.php` as from now on we should fix our existing code base style. This caused the build to fail as expected. We should probably also introduce a pre commit or push hook to run `phpcs` automatically for developers locally to avoid the failure of builds.

l3ku:
Ping @ottok, rebased on master and made the changes mentioned above. Is the review approved?

ottok:
Normally I would never merge something when the CI tests are failing, but I guess I'll do an exception this time as I anticipate that you will fix the style errors ASAP.

Issue #101, opened 2018-05-18
Cruft remover suggest improper removables


simoke:
Cruft Remover suggests "migration" directories to be removed from places that could break a theme or plugin if they were removed. For example, Divi theme has the directory /data/wordpress/htdocs/wp-content/themes/Divi/includes/builder/module/settings/migration containing php files which are necessary for the theme to work.

mikkoaf:
This can be fixed for example by not listing files identified as cruft from `/themes`, `/mu-plugins` and `/plugins` to the user.

PR #114, opened 2018-05-29
Remove files in plugins and themes from suggested cruft

mikkoaf:
Removing files from plugins and themes is generally not advisable. The cruft remover in it's current state can suggest the user such files and directories which might have unintended effects as mentioned in #101 . In this PR such items are not suggested to the user, making the cruft remover safer for the user.

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -60,18 +60,23 @@ function seravo_ajax_list_cruft_files() {
   switch ( $_REQUEST['section'] ) {
     case 'cruftfiles_status':
 
-      // List of known types of cruft files
-      $list_files = array( '*.sql', '.hhvm.hhbc', '*.wpress' );
-      // List of known cruft directories
-      $list_dirs = array( 'siirto', 'palautus', 'vanha', '*-old', '*-copy', '*-2', '*.bak', 'migration',
-                          '*_BAK', '_mu-plugins', '*.orig', '-backup', '*.backup' );
-      $list_known_files = array();
-      $list_known_dirs = array(
-        '/data/wordpress/htdocs/wp-content/plugins/all-in-one-wp-migration/storage',
-        '/data/wordpress/htdocs/wp-content/ai1wm-backups',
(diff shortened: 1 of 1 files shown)


lib/cruftfiles-ajax.php
```diff
@@ -35,6 +35,11 @@ function list_known_cruft_dir( $name ) {
       '/data/wordpress/htdocs/wp-content/uploads/backupbuddy_backups',
       '/data/wordpress/htdocs/wp-content/updraft',
     );
+    $white_list_dirs = array(
+      "/data/wordpress/htdocs/wp-content/plugins",
```
l3ku: Please use single quotes if you are not evaluating anything in the string: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#single-and-double-quotes

lib/cruftfiles-ajax.php
```diff
@@ -35,6 +35,11 @@ function list_known_cruft_dir( $name ) {
...
+      "/data/wordpress/htdocs/wp-content/mu-plugins",
+      "/data/wordpress/htdocs/wp-content/themes"
```
l3ku: Add trailing comma to last array item: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation

lib/cruftfiles-ajax.php
```diff
@@ -55,6 +60,17 @@ function list_known_cruft_dir( $name ) {
         $crufts = array_merge($crufts, $cruft_found);
       }
     }
+    // This should be performed right after cruftfile search and before wp core
+    foreach ( $white_list_dirs as $dirname ) {
+      // Some directories are whitelisted and their files should not be deleted
+      $keep = array();
+      foreach ($crufts as $filename) {
```
l3ku: Add whitespace to foreach loop, see other foreach loops for reference:
```
foreach ( $crufts as $filename ) {
```

mikkoaf:
Code style issues have been addressed.

PR #115, opened 2018-05-30
Improve access restrictions and capability checks

l3ku:


ypcs:
Haven't tested yet, code & changes look good.

Especially changes made in ca30cc0 (ie. use `add_action()` instead of directly accessing file) = :+1: :+1: :+1:, this should be expanded to `lib/backups-ajax.php`.

Commit c524242 has too long subject line in the commit message, recommended maximum is 72 characters, current one is 75 characters. Longer lines cause various tools, like GitHub, to wrap the line.

l3ku:
Thanks for reviewing! Made the requested changes:
- Refactored AJAX functionality also for `lib/backups-ajax.php`
- Shortened commit message for ec55ee9

ypcs:
Results from initial testing:

 - OK: I see site error count @ dashboard only if logged in as admin
 - FAIL: I don't see login notifications in any situations, however this doesn't work from `master` either
 - OK: ajax queries work
 - FAIL: Tools -> Tests doesn't work, submitted separate report about this (JS error)

PR #116, opened 2018-06-04
Update Cruft Remover layout to WordPress layout

mikkoaf:
As stated by @nasusa the postbox-style of layout is better than what was used before. For consistency and usability, in this PR cruft remover uses the same style as most other features in seravo-plugin.
image
image

### DIFF
--- a/js/cruftfiles.js
+++ b/js/cruftfiles.js
@@ -67,4 +67,21 @@ jQuery(document).ready(function($) {
 
   // Load on page load
   seravo_load_report('cruftfiles_status');
+    // Postbox toggle script
+  jQuery('.ui-sortable-handle').on('click', function () {
+    jQuery(this).parent().toggleClass("closed");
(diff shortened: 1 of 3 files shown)


js/cruftfiles.js
```diff
@@ -67,4 +67,21 @@ jQuery(document).ready(function($) {
...
+    } else {
+      jQuery(this).parent().parent().parent().height('auto');
+    }
+  });
+  jQuery('.toggle-indicator').on('click', function () {
+    jQuery(this).parent().parent().toggleClass("closed");
+    if (jQuery(this).parent().hasClass("closed")) {
+      jQuery(this).parent().parent().parent().parent().height(60);
+    } else {
+      jQuery(this).parent().parent().parent().parent().height('auto');
```
l3ku: For overall better code readability and style, please use `$('#element').parents().eq(3);` instead of having multiple `parent()` calls in succession (https://api.jquery.com/parents/).

PR #119, opened 2018-06-05
Fix code style errors

mikkoaf:
In this PR all the issues creating errors with PHP_CodeSniffer are resolved.

### DIFF
--- a/lib/backups-page.php
+++ b/lib/backups-page.php
@@ -21,7 +21,7 @@ if ( ! defined('ABSPATH') ) {
                 <span><?php _e('Backups', 'seravo'); ?></span>
               </h2>
               <div class="inside">
-                <p><?php _e('Backups are made automatically every night and preserved for 30 days. The data can be accessed on the server at <code>/data/backups</code>.', 'seravo') ?></p>
+                <p><?php _e('Backups are made automatically every night and preserved for 30 days. The data can be accessed on the server at <code>/data/backups</code>.', 'seravo'); ?></p>
               </div>
           </div>
           <!-- first ends -->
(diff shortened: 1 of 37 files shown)


lib/backups-page.php
```diff
@@ -56,7 +56,10 @@
               <span><?php _e('Files excluded from backups', 'seravo'); ?></span>
             </h2>
             <div class="inside">
-              <?php echo wp_sprintf( __('Below is the content of the file %s.', 'seravo'), '<code>/data/backups/exclude.filelist</code>' ); ?>
+              <?php
+                // translators: %s
```
l3ku: The PHPCS error regarding the "translators:" comment implies that you should try to explain to the translators what "%s" means in the context of the sentence, see https://developer.wordpress.org/themes/functionality/internationalization/#variables for details. Please imply this logic also to the other code blocks.

lib/helpers.php
```diff
@@ -16,24 +16,25 @@ class Helpers {
...
     public static function is_staging() {
-      return (getenv('WP_ENV') && getenv('WP_ENV') === 'staging');
+      return ( getenv('WP_ENV') && getenv('WP_ENV') === 'staging' );
     }
 
     public static function human_file_size( $size, $precision = 2 ) {
-      for ( $i = 0; ($size / 1024) > 0.9; $i++, $size /= 1024 ) {}
+      for ( $i = 0; ( $size / 1024 ) > 0.9; $i++, $size /= 1024 ) {
         return round($size, $precision) . [ 'B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB' ][ $i ];
+      }
```
l3ku: The return statement is not probably meant to be inside the for loop, as the for loop is meant to increase the `$i` variable. Now the loop stops on the first run and functionality is probably broken.

modules/fixes.php
```diff
@@ -84,46 +84,47 @@ public static function maybe_clear_alloptions_cache( $option ) {
...
       // control alert type
       $type = '';
       if ( isset($response->type) ) {
         $type = $response->type;
       }
       if ( ! empty($message) ) {
-      ?>
-        <div class="<?php esc_attr_e($type) ?> notice is-dismissible">
+        ?>
+        <div class="<?php esc_attr($type); ?> notice is-dismissible">
```
l3ku: Using `esc_attr_e` here means that the content is also outputted. If you change it to `esc_attr`, it should be transformed into `echo esc_attr()`.

PR #120, opened 2018-06-06
Add new mail forwards-tool

nasusa:
- Add mails.php, mails-forward.php and mails-page.php -files
- Load module 'modules/mails.php' in seravo-plugin.php -file

Mails -tool purpose is to show mail forwards, and later show also mailboxes. Tool was requested by @elguitar.


l3ku:
Please rebase with newest master so the PHPCI check will not fail on the errors in the already existing codebase.

elguitar:
This feature depends on the backend API. There's no need to merge this before the API has reached production.

l3ku:
Is it necessary to list the all the site domains repeatedly both on the domains and mail forwards submenu page? 

screen shot 2018-06-07 at 8 38 29

Could this be redesigned in such a way, that either the mail forwards are listed on the domains page, or that only the mail forwards are listed on the mail forwards submenu page without having to click each domain to access them?

l3ku:
Backend API support is now available, so this can be implemented.

ottok:
This was implemented in https://github.com/Seravo/seravo-plugin/commit/2ed13eaff28ae631aeaddde89284aa4c25868f82

PR #122, opened 2018-06-07
Update tools language translations

nasusa:
There was couple problems with finnish-translations in reports- and optimize images-tool.

I fixed those and also updated .po-file.


l3ku:
When https://github.com/Seravo/seravo-plugin/pull/123 is merged to master, please rebase on master and update the translations based on the new UI, as it is not sensible to update translations before UI changes.

PR #131, opened 2018-06-14
Introduce feature to reset shadows

mikkoaf:
In this PR the user is given an option of resetting shadow instances from the WordPress user interface. User can track their shadow instances from a new Shadows page in their Tools menu.

image

Pressing the ``` Reset Shadow ```-button begins the resetting operation, which is indicated to the user by disabling the button. Successful operation is indicated to the user by changing the text in the button to ```DONE```

### DIFF
--- a/js/shadows.js
+++ b/js/shadows.js
@@ -2,16 +2,20 @@
 
 jQuery(document).ready(function($) {
   $('.shadow-reset').click(function(event) {
+    var is_user_sure = confirm(seravo_shadows_loc.confirm);
+    if ( ! is_user_sure) {
+      return;
+    }
     seravo_ajax_reset_shadow($(this).attr("data-shadow-name"),
       function( status ){
-        if( status == 'progress' ) {
+        if ( status == 'progress' ) {
           event.target.disabled = true;
-        }
-        else if( status == 'success' ){
-          event.target.innerHTML = 'DONE';
-        }
-        else {
-          console.log('Unknown status');
+        } else if ( status == 'success' ) {
+          event.target.innerHTML = seravo_shadows_loc.success;
+        } else if ( status == 'failure' ) {
+          event.target.innerHTML = seravo_shadows_loc.failure;
+        } else {
+          event.target.innerHTML = seravo_shadows_loc.error;
         }
       });
   });
@@ -25,10 +29,31 @@ jQuery(document).ready(function($) {
         'resetshadow': shadow },
         function( rawData ) {
           var data = JSON.parse(rawData);
-          console.log(data);
-          animate('success');
-      }
+          // If the last row of rawData does not begin with SUCCESS:
+          if ( data[data.length - 1].search('Success') ) {
+            animate('success');
+          } else {
+            animate('failure');
+          }
+        }
     );
   }
+
+  //Postbox animations
+  jQuery('.ui-sortable-handle').on('click', function () {
+    jQuery(this).parent().toggleClass("closed");
+    if (jQuery(this).parent().hasClass("closed")) {
+      jQuery(this).parents().eq(3).height(60);
+    } else {
+      jQuery(this).parents().eq(3).height('auto');
+    }
+  });
+  jQuery('.toggle-indicator').on('click', function () {
+    jQuery(this).parent().parent().toggleClass("closed");
+    if (jQuery(this).parent().hasClass("closed")) {
+      jQuery(this).parents().eq(4).height(60);
+    } else {
+      jQuery(this).parents().eq(4).height('auto');
+    }
+  });
 });
-    
--- a/js/shadows.js
+++ b/js/shadows.js
@@ -1,0 +1,34 @@
+'use strict';
+
+jQuery(document).ready(function($) {
+  $('.shadow-reset').click(function(event) {
+    seravo_ajax_reset_shadow($(this).attr("data-shadow-name"),
+      function( status ){
+        if( status == 'progress' ) {
+          event.target.disabled = true;
+        }
+        else if( status == 'success' ){
+          event.target.innerHTML = 'DONE';
+        }
+        else {
+          console.log('Unknown status');
+        }
+      });
+  });
+
+  function seravo_ajax_reset_shadow(shadow, animate) {
+    animate('progress');
+    $.post(
+      ajaxurl,
+      { type: 'POST',
+        'action': 'seravo_reset_shadow',
+        'resetshadow': shadow },
+        function( rawData ) {
+          var data = JSON.parse(rawData);
+          console.log(data);
+          animate('success');
+      }
+    );
+  }
+});
+    
--- a/lib/shadows-ajax.php
+++ b/lib/shadows-ajax.php
@@ -4,7 +4,7 @@ function seravo_reset_shadow() {
   if ( isset($_POST['resetshadow']) && ! empty($_POST['resetshadow']) ) {
     $shadow = $_POST['resetshadow'];
     $output = array();
-    exec('. /etc/container_environment.sh; wp-shadow-reset ' . $shadow . ' --force 2>&1', $output);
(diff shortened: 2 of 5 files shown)


nasusa:
I would add `class="button"` to reset button so It would not look like Windows XP button.
Also, small fix:

image

I would make it more similar like HTTP-request statistics postbox in Reports-tool. But looks great 👍

mikkoaf:
I've consulted @tlxo on-site about the UI and UX. @nasusa and I fixed a scalability issue that arose when testing the site.

lib/shadows-ajax.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+function seravo_reset_shadow() {
+  if ( isset($_POST['resetshadow']) && ! empty($_POST['resetshadow']) ) {
+    $shadow = $_POST['resetshadow'];
+    $output = array();
+    exec('. /etc/container_environment.sh; wp-shadow-reset ' . $shadow . ' --force 2>&1', $output);
```
l3ku: The container environment should be available in the newest version of the `wp-shadow-reset` script. I updated your site so the new script version should be there, so please remove the command `. /etc/container_environment.sh;`.

js/shadows.js
```diff
@@ -0,0 +1,49 @@
...
+  function seravo_ajax_reset_shadow(shadow, animate) {
+    animate('progress');
+    $.post(
+      ajaxurl,
+      { type: 'POST',
+        'action': 'seravo_reset_shadow',
+        'resetshadow': shadow },
+        function( rawData ) {
+          var data = JSON.parse(rawData);
+          console.log(data);
```
l3ku: Console logging and other debugging measures should not be done in production JavaScript.

js/shadows.js
```diff
@@ -0,0 +1,49 @@
...
+          animate('success');
```
l3ku: What is done if the shadow reset process fails? Is the success message still displayed? Please provide some error handling mechanism so the user knows something went wrong. You could ask the UI guys for some advice on their suggested practices of displaying errors to users.

js/shadows.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+jQuery(document).ready(function($) {
+  $('.shadow-reset').click(function(event) {
+    seravo_ajax_reset_shadow($(this).attr("data-shadow-name"),
+      function( status ){
+        if ( status == 'progress' ) {
+          event.target.disabled = true;
+        } else if ( status == 'success' ) {
+          event.target.innerHTML = 'DONE';
```
l3ku: I think displaying "Success" would provide more information about the actual result than "DONE". Also, the JavaScript should be localized by WP's mechanims instead of hard coding the strings to JavaScript. See https://codex.wordpress.org/Function_Reference/wp_localize_script and existing code in e.g. https://github.com/Seravo/seravo-plugin/blob/2eb11f8d2115e5b729062cd74e8b85063c53f3a4/modules/tests.php#L33 for implementation details.

modules/shadows.php
```diff
@@ -0,0 +1,37 @@
...
+      // Load only in production
+      if ( Helpers::is_production() ) {
+        add_action( 'admin_menu', array( __CLASS__, 'register_shadows_page' ) );
+        wp_register_script( 'seravo_shadows', plugin_dir_url( __DIR__ ) . '/js/shadows.js' );
+        wp_enqueue_script( 'seravo_shadows' );
+        add_action( 'wp_ajax_seravo_reset_shadow', 'seravo_reset_shadow' );
+      }
+    }
+    public static function register_shadows_page() {
+      add_submenu_page( 'tools.php', __('Shadows', 'seravo'),
```
l3ku: The following lines of multiline function calls should be indented once instead of being at the same level as the actual function call:
```
add_submenu_page(
  'some_param1'
   some_param2,
   some_param3
);
```
I might add some code sniff for this later on.

seravo-plugin.php
```diff
@@ -247,6 +247,13 @@ public static function load_all_modules() {
     if ( defined( 'WP_CLI' ) && WP_CLI ) {
       require_once dirname( __FILE__ ) . '/modules/wp-cli.php';
     }
+
+    /**
+     * Shadows page
+     */
+    if ( apply_filters('seravo_show_shadows_page', true) && current_user_can( 'administrator' ) ) {
+      require_once dirname( __FILE__ ) . '/modules/shadows.php';
+    }
```
l3ku: I think that this functionality should be hidden from normal multisite admins, as reseting a shadow should not be something every multisite admin should be able to do. So please move the loading of the shadows module inside the if statement inside https://github.com/Seravo/seravo-plugin/blob/cf3194b26dbd0b7931ffd7ded34212cfb607fc04/seravo-plugin.php#L172.

lib/shadows-page.php
```diff
@@ -0,0 +1,87 @@
...
+                      if ( is_wp_error($shadow_list) ) {
+                        die($shadow_list->get_error_message());
+                      }
+                      // Order the shadow data so they correspond to the table labels.
+                      $shadow_data_ordered = array( 'info', 'name', 'ssh', 'created' );
+                      ?>
+                      <table class="widefat striped fixed"
+                        style="width=100%; border:none" cellspacing="0">
+                        <thead>
+                          <th><?php _e('Name', 'seravo'); ?></th>
```
l3ku: If there are no shadows, should the table headers be hidden?
screen shot 2018-06-15 at 11 51 24

Also, I think that having all the WP admin notifications in the post boxes makes pages look messy, but that is a separate issue to address.

lib/shadows-page.php
```diff
@@ -0,0 +1,87 @@
...
+                        <tbody>
+                          <?php if ( ! empty($shadow_list) ) : ?>
+                            <?php foreach ( $shadow_list as $shadow_data ) : ?>
+                              <tr>
+                                <?php foreach ( $shadow_data_ordered as $data_key ) : ?>
+                                  <?php if ( isset($shadow_data[ $data_key ]) ) : ?>
+                                    <td><?php echo $shadow_data[ $data_key ]; ?></td>
+                                  <?php endif; ?>
+                                <?php endforeach; ?>
+                                <td><button data-shadow-name="<?php echo $shadow_data['name']; ?>" class="shadow-reset button" type="button">Reset Shadow</button></td>
```
l3ku: If you hard code "Reset Shadow" inside the button, there is no possibility for us to translate it with gettext. Use `__()` or `_e()`, see existing code for details.

lib/shadows-ajax.php
mikkoaf: Thanks!

l3ku:
Tested this again and put the emphasis on trying to think like a first-time user who views the page, and found a couple of notices: 
- I think that you should probably add some JavaScript "Are you sure" type dialog there, as the shadow reset is an operation that should not be done on accident. Now clicking the button immediately starts the procedure.
- You should probably add the description of the command somewhere (either the popup dialog or the postbox description), where it would be clarified that it replaces the current shadow files and database. Currently the user would have no idea what the "Reset shadow" button does.
- Add a very brief description about our shadows functionality and include link to the shadows documentation at https://seravo.com/docs/deployment/shadows/ for more in-depth documentation.

After the changes above, rebase with the newest master.

mikkoaf:
I've fixed the user experience issues you have addressed and rebased to the current master.

l3ku:
Please make some of the descriptions more precise for the user, for example:

> Allow easy access to site shadows. Replaces the current shadow files and database. For more information, visit our Developer documentation

=>

> Allow easy access to site shadows. Resetting a shadow copies the production state to a shadow by replacing files and importing the production database. For more information, visit our Developer documentation.

Also

> Are you sure? This removes all information in the selected environment.

=> 

> Are you sure? This replaces all information in the selected environment.

PR #139, opened 2018-06-29
Show more information on suggested cruft-files

mikkoaf:
Removing cruft-files reduces clutter from the users' websites reducing disk usage. However, removing files should be done carefully, and this can be achieved by giving the user more information.

In this PR the cruft remover is enhanced by showing the user the last file modification date and file size.

image

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -5,10 +5,11 @@ if ( ! defined('ABSPATH') ) {
 }
 
 function add_file_information( $file ) {
-  exec( 'du ' . $file . ' -h', $output );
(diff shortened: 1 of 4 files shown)


lib/cruftfiles-ajax.php
```diff
@@ -133,6 +144,9 @@ function seravo_ajax_list_cruft_files() {
       }
       $crufts = array_unique($crufts);
       set_transient('cruft_files_found', $crufts, 600);
+
+      $crufts = array_map('add_file_information', $crufts);
+      error_log(json_encode($crufts));
```
l3ku: Please remove debug logging from production code.

mikkoaf:
Debugging output has now been removed from the PR.

Issue #142, opened 2018-07-03
Sort the tools provided by Seravo-plugin into concise categories


mikkoaf:
Currently Seravo's  tools take up 10 rows Tools-tab
![image](https://user-images.githubusercontent.com/16817737/42205752-f4b40b6c-7ead-11e8-81bd-05a8bcd6d21a.png)
I've come up with a suggestion to decrease the number of tabs to 3, making the plugin easier to use, decreasing clutter and combining similar functionalities together.
![image](https://user-images.githubusercontent.com/16817737/42206282-5744d3f0-7eaf-11e8-941a-4c2498d6d087.png)
_a sketch of what the end result might look like_

mikkoaf:
I suggest that the categories could resemble the tasks that can be preformed and the information provided with the tools.
For example the layout could be

1. Data Usage or Content Management or similar

- Any backup-related features
- Any database-related features
- Cruft-file and -plugin features
- Image optimization

2. Maintenance or administrative tasks

- Any reports-related features
- Logs
- Updates and site status features
- Tests
- Data integrity

3. Developer tools

- Shadows-feature

mikkoaf:
This is a duplicate of #98

mikkoaf:
WordPress postbox elements usually are reserved for simple functionalities and showing information. Settings are traditionally shown as their own pages.

After discussing with @tlxo we came to the conclusion that Logs and Domains -pages should be left as their own submenu-pages as they do not translate to postboxes well. The Image optimization tool is such a small functionality which can be considered turning into a postbox, even though it is technically a setting.

mikkoaf:
I've managed to reduce the menu to the following:
![image](https://user-images.githubusercontent.com/16817737/42678860-dfe3e184-8688-11e8-83a5-ce90c973747f.png)
"Päivitykset" and "Optimoi kuvia" are the next ones to be refactored to use AJAX and , after that this is pretty much a case of UX optimizing.

l3ku:
Now that we have our own postboxes in the plugin instead of hard-coded pages, the tools can be easily swapped between different admin screens.

ottok:
@nuuttitoivola @elguitar @JoosuaKoskinen This issue needs an status update

nuuttitoivola:
The current status in the version `1.9.20` is as follows: 
[...]

Issue #98, opened 2018-05-16
Redesign plugin menu and page structure


l3ku:
Until now when we have added new features to Seravo Plugin, we have simply added a new page under "Tools". This has resulted in the menu becoming increasingly bloated with features, e.g.:

![screen shot 2018-05-16 at 14 57 56](https://user-images.githubusercontent.com/24781811/40115580-a1f4445a-5919-11e8-90af-de5c9029673a.png)

Before we continue to implement new features like the functionality described in #32, we should redesign the plugin menu structure by merging the content of different pages into similar categories, or even introducing an entirely new top-level WP admin menu page.

simoke:
Also the upcoming GDPR features will be located under Tools menu, so relocating of Seravo plugin menu items makes sense.

l3ku:
Note that if we choose to introduce a new WP admin menu page, we must also design a proper filter functionality: e.g. changing the name of the menu page or completely hiding it should be possible by adding filters to the site theme. This way, the site can also be branded with Seravo being merely the hosting provider (not showing up immediately in the customer's UI).

elguitar:
Closing in favor of https://github.com/Seravo/seravo-plugin/issues/233

PR #143, opened 2018-07-05
Refactor existing Tests and Reports -pages to current architecture

mikkoaf:
In this changes have been made to the Tests and Reports -pages in order to make future changes easier. The pages are now more similar to how the Cruft remover and Shadows -pages work. With these changes it's easier to redesign the plugin menu layout as described in #98 and #142.

### DIFF
--- a/lib/tests-page.php
+++ b/lib/tests-page.php
@@ -7,36 +7,53 @@ if ( ! defined('ABSPATH') ) {
 
 <div class="wrap">
 
-    <h1><?php _e('Tests', 'seravo'); ?> (beta)</h1>
-
-    <p>
-        <?php
-            _e('Here you can test the core functionality of the WordPress installation on your site.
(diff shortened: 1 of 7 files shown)


lib/tests-page.php
```diff
@@ -1,42 +1,59 @@
 <?php
 // Deny direct access to this file
 if ( ! defined('ABSPATH') ) {
-  die('Access denied!');
+    die('Access denied!');
```
l3ku: Please revert this change, as the indentation was already correct with two spaces.

mikkoaf:
In the current iteration the "Show more" link behaves better and the unnecessary modification on line 4 is now reverted.
I have planned on improving upon how the postboxes are presented to the user at a later time, as with this PR the intention is to make it easier to rearrange the menu.

PR #147, opened 2018-07-12
Add nonces to WP ajax-requests

mikkoaf:


modules/shadows.php
```diff
@@ -38,18 +43,20 @@ public static function register_shadows_page() {
     }
 
     public static function register_shadows_scripts( $page ) {
-      wp_register_style('seravo_shadows', plugin_dir_url( __DIR__ ) . '/css/shadows.css' );
+      wp_register_style('seravo_shadows', plugin_dir_url( __DIR__ ) . '/style/shadows.css' );
```
l3ku: I happened to notice that the file `shadows.css` has never existed, so you could as well remove this line.

ypcs:
Code looks good, if it works (didn't test / don't have time to test), its good :)

Issue #148, opened 2018-07-13
Make Updates-page fail gracefully when in development environment


mikkoaf:
Updates page when it is working normally
![image](https://user-images.githubusercontent.com/16817737/42677746-65662230-8685-11e8-8cae-7a01b2a373ce.png)
Due to the nature how the Updates page is currently setup, it only returns the following in local development environment:
![image](https://user-images.githubusercontent.com/16817737/42677816-a4a7fe82-8685-11e8-98c3-3fdf4adf1dec.png)
This makes developing quite difficult due to the disruptive nature of the message and if such an event would occur in production, I would rather show postboxes with "Service temporarily unavailable" -message or such.

l3ku:
Fixed in e521d2f. Next time, use GitHub's automation (Closes: xx) so issues don't have to be manually closed afterwards.

PR #149, opened 2018-07-16
Allow updates -tool-page fail gracefully

mikkoaf:
As described in #148 updates-tools currently handle exceptions so that when API can not be reached, the page is not loaded at all.
In this PR the way how such events are handled is changed so that when encountering an error the postboxes are shown. This makes further development easier.
image

### DIFF
--- a/modules/updates.php
+++ b/modules/updates.php
@@ -57,10 +57,6 @@ if ( ! class_exists('Updates') ) {
 
     public static function seravo_admin_get_site_info() {
       $site_info = API::get_site_data();
-      if ( is_wp_error($site_info) ) {
-        die($site_info->get_error_message());
-      }
-
       return $site_info;
     }
 
--- a/lib/updates-page.php
+++ b/lib/updates-page.php
@@ -29,47 +29,51 @@ if ( ! defined('ABSPATH') ) {
               <h2 class="hndle ui-sortable-handle">
                 <span><?php _e('Seravo updates', 'seravo'); ?></span>
               </h2>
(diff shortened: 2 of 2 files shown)


modules/updates.php
```diff
@@ -58,7 +58,7 @@ public static function load_updates_page() {
     public static function seravo_admin_get_site_info() {
       $site_info = API::get_site_data();
       if ( is_wp_error($site_info) ) {
-        die($site_info->get_error_message());
+        $site_info->get_error_message();
```
l3ku: This line now acheives nothing, as `$site_info->get_error_message();` returns a string, but it is not assigned to any variable. Previously, the error message was printed with `die()`, but this also caused the PHP to exit. Maybe the error message should be printed somewhere on the updates page by checking `is_wp_error()` there and outputting the WP_Error message created in https://github.com/mikkoaf/seravo-plugin/blob/56b8ff3674a4f42d727740fb87c316eec320761d/lib/api.php#L61 there.

mikkoaf: I solved this part by printing out the returned string on appropriate rows https://github.com/mikkoaf/seravo-plugin/blob/e521d2f894af6aafc03a8609f65a7cb049611b59/lib/updates-page.php#L74
and
https://github.com/mikkoaf/seravo-plugin/blob/e521d2f894af6aafc03a8609f65a7cb049611b59/lib/updates-page.php#L102

l3ku:
In the future, we could also apply the graceful fail also to the updation of settings, so e.g. an admin notice would be shown when the updation fails (currently `die()` is used).

Issue #153, opened 2018-07-23
Cruft-file remover shows folders and the contained files as separate items


mikkoaf:
This currently how cruft-files are searched, presented and removed can have unintended effects. 
![image](https://user-images.githubusercontent.com/16817737/43075167-8e025c94-8e88-11e8-9ecc-9dbe5e214541.png)
Selecting only the last cruft item on the list removes the whole folder with only a single prompt and on the UI only the last item is removed. This also makes estimating the actual used storage space difficult, as files inside folders that are listed are essentially twice - once as their own items and once included in the storage space used by the folder.

l3ku:
Fixed in bead5f1. Next time, use GitHub's automation (Closes: xx) so issues don't have to be manually closed afterwards.

PR #157, opened 2018-07-24
Fix cruft file remover information retrieval

mikkoaf:
This commit fixes the issue #153 by removing sub-directories and files from the listings. This also changes how the cruft remover shows file size information. Previously the shown directory size information was about the first directory within the directory, if applicable.

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -143,6 +143,16 @@ function seravo_ajax_list_cruft_files() {
           $crufts = array_merge($crufts, $cruft_found);
         }
       }
+
+      $crufts = array_filter($crufts, function( $item ) use ( $crufts ) {
+        foreach ( $crufts as $substring ) {
+          if ( strpos($item, $substring) === 0 && $item !== $substring ) {
+            return false;
+          }
+        }
+        return true;
+      });
+
       $crufts = array_unique($crufts);
       set_transient('cruft_files_found', $crufts, 600);
 


lib/cruftfiles-ajax.php
```diff
@@ -6,10 +6,11 @@
...
   exec( 'du ' . $file . ' -h --time', $output );
-  $size = explode("\t", $output[0]);
-
-  $data['size'] = $size[0];
-  $data['mod_date'] = $size[1];
+  $size = explode("\t", end($output));
+  $mod_date = array_values(array_slice($size, -2, 1 ))[0];
+  $size = array_values(array_slice($size, -3, 1 ))[0];
+  $data['size'] = $size;
+  $data['mod_date'] = $mod_date;
```
l3ku: Could you please explain why this change is done? I tried this in PHP interactive and couldn't find any scenario in which the already existing code would not work:
```
php > exec('du /data/wordpress/nginx/ -h --time', $output);
php > $size = explode("\t", end($output));
php > print_r($size);
Array
(
    [0] => 12K
    [1] => 2018-07-11 15:07
    [2] => /data/wordpress/nginx/
)
php > exec('du /data/wordpress/ -h --time', $output);
php > $size = explode("\t", end($output));
php > print_r($size);
Array
(
    [0] => 120M
    [1] => 2018-08-06 10:39
    [2] => /data/wordpress/
)
php > exec('du /data/wordpress/* -h --time', $output);
php > $size = explode("\t", end($output));
php > print_r($size);
Array
(
    [0] => 652K
    [1] => 2017-08-10 19:09
    [2] => /data/wordpress/vendor
)
php > exec('du /data/wordpress/htdocs/wp-content/mu-plugins -h --time', $output);
php > $size = explode("\t", end($output));
php > print_r($size);
Array
(
    [0] => 6.9M
    [1] => 2018-08-06 10:39
    [2] => /data/wordpress/htdocs/wp-content/mu-plugins
)
```
The above would in my opinion imply that the file size would always be in $size[0], the modification date in $size[1] and filename in $size[2].

mikkoaf: During testing and developing I recall running into a situation where this change was solved a problem I ran into.
However, I can not reproduce root cause for this change at this point in time.

mikkoaf:
Rebased and addressed the requested changes

PR #159, opened 2018-07-24
Fix Shadows-page handling of missing data

mikkoaf:
In this PR, when handling staging environments any situations where data fields were missing an empty cell is printed instead of nothing at all.


l3ku:
Just for a notice, but in my opinion a more elegant solution could be:
```
 <td>
  <?php if ( isset($shadow_data[ $data_key ]) ) : ?>
    <?php echo $shadow_data[ $data_key ]; ?>
  <?php endif; ?>
</td>
```
The functionality is the same, but I am somehow obsessed with omitting else-blocks... :)

This can be merged after rebasing with master.

mikkoaf:
I've rebased this PR to master

l3ku:
My aplogies for letting this PR hang for so long, but could you rebase with master again? After that I will merge the PR.

mikkoaf:
I've re-rebased this PR to master.

PR #164, opened 2018-07-31
Update coding standards

mikkoaf:
This PR updates Seravo-plugin coding standards.

### DIFF
--- a/lib/helpers.php
+++ b/lib/helpers.php
@@ -31,6 +31,10 @@ if ( ! class_exists('Helpers') ) {
       return ( getenv('WP_ENV') && getenv('WP_ENV') === 'staging' );
     }
 
+    public static function seravo_plugin_version() {
+        return get_plugin_data( WP_CONTENT_DIR . '/mu-plugins/seravo-plugin/seravo-plugin.php' )['Version'];
+    }
+
     public static function human_file_size( $size, $precision = 2 ) {
       for ( $i = 0; ( $size / 1024 ) > 0.9; ) {
         $i++;
--- a/modules/backups.php
+++ b/modules/backups.php
@@ -38,8 +38,8 @@ if ( ! class_exists('Backups') ) {
      * @param string $page hook name
      */
     public static function register_backups_scripts( $page ) {
-      wp_register_script('seravo_backups', plugin_dir_url(__DIR__) . '/js/backups.js');
-      wp_register_style('seravo_backups', plugin_dir_url(__DIR__) . '/style/backups.css');
+      wp_register_script('seravo_backups', plugin_dir_url(__DIR__) . '/js/backups.js', '', Helpers::seravo_plugin_version());
+      wp_register_style('seravo_backups', plugin_dir_url(__DIR__) . '/style/backups.css', '', Helpers::seravo_plugin_version());
 
       if ( $page === 'tools_page_backups_page' ) {
         wp_enqueue_script( 'seravo_backups' );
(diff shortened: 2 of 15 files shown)


lib/helpers.php
```diff
@@ -31,6 +31,15 @@ public static function is_staging() {
       return ( getenv('WP_ENV') && getenv('WP_ENV') === 'staging' );
     }
 
+    public static function seravo_plugin_version() {
+      if ( is_admin() ) {
+        $var = '1.0';
+        $var = get_plugin_data( WP_CONTENT_DIR . '/mu-plugins/seravo-plugin/seravo-plugin.php' );
+        $var = $var['Version'];
+        return $var;
```
l3ku: Can you explain in words what is going on here? The same variable is assigned multiple times for some reason.

modules/backups.php
```diff
@@ -7,6 +7,7 @@
 
 namespace Seravo;
 
+use Seravo\Helpers;
```
l3ku: If you're inside the `Seravo` namespace, to my understanding you could just use the `Helpers` class directly without using `Seravo\Helpers` inside `Seravo`.

phpcs.xml
```diff
@@ -52,7 +52,9 @@
     assessed separately. Also we can't always use nonces e.g. for APIs... -->
     <exclude name="WordPress.CSRF.NonceVerification.NoNonceVerification" />
     <exclude name="WordPress.XSS.EscapeOutput" />
-
+    <exclude name="WordPress.Security.EscapeOutput.UnsafePrintingFunction" />
+    <exclude name="WordPress.Security.EscapeOutput.OutputNotEscaped" />
+    <exclude name="WordPress.Security.NonceVerification.NoNonceVerification" />
```
l3ku: Please provide an explanation for each of these as to why they are excluded (see other excludes for example). I bet that excluding them is reasonable but it is always good to provide some reference.

l3ku:
Also as a notice, note that your branch is named `origin/update-coding-standards`, if not intended, this might cause some confusion as the default remote is named `origin` in git :).

lib/helpers.php
elguitar: I apologize for jumping in like this but I would suggest that the $var should be named in a more descriptive way.

phpcs.xml
mikkoaf: These exclusions are related to the comment on rows 51-52.

lib/helpers.php
l3ku: I agree. No need to apologize, the more eyes we get on code reviews the better :)

phpcs.xml
l3ku: Ok 👍

lib/helpers.php
mikkoaf: @elguitar thanks for the input! I'll keep this in mind - even if this variable is used for only a few rows, is better to describe it well for better readability.

modules/backups.php
mikkoaf: For some reason I could not replicate what you described.

mikkoaf:
The requested changes have been made and the branch has been rebased.

l3ku:
The changes made to the static asset query strings are not most likely correct:
screen shot 2018-08-01 at 16 09 43

Also, you can remove the lines `use Seravo\Helpers;` when inside the`Seravo` namespace (tested and works).

mikkoaf:
Latest changes address the requested changes.

lib/helpers.php
```diff
@@ -31,6 +31,12 @@ public static function is_staging() {
       return ( getenv('WP_ENV') && getenv('WP_ENV') === 'staging' );
     }
 
+    public static function seravo_plugin_version() {
+      if ( is_admin() ) {
```
l3ku: I was already about to merge, but could you explain why this function only returns something when the user is in WP admin? We load e.g. `instance-switcher.js` and `instance-switcher.css` in the frontend side, and this results in the version query strings missing as the function returns `null`.

PR #166, opened 2018-08-06
Feature Seravo custom admin postboxes

l3ku:
This PR provides a module for creating WP core dashboard-like postboxes, that can be dragged in custom orders and opened/closed. The postbox location and opened/closed state is saved to the site database for each individual user to enable custom settings per-user. 

As the current display mechanism is identical to the dashboard widget-like mechanism in WP, there is no way for us to have wider/narrower boxes than what WP dashboard CSS defines. If someone has a good idea on how to enable this whilst still keeping the drag/drop functionality, I am open to suggestions.

This can be freely tested at seravo.com and wp-palvelu.fi production, and I hope that there would be some discussion especially about the current usage mechanism:
- Register an admin page that uses `Seravo\seravo_postboxes_page` as callback function:
```add_submenu_page( 'tools.php', __('Backups', 'seravo'), __('Backups', 'seravo'), 'manage_options', 'backups_page', 'Seravo\seravo_postboxes_page' );```
- Add postboxes to the page in a similar mechanism as WP's `add_meta_box()`:
```
seravo_add_postbox(
  'backups-list',
  __('Current backups', 'seravo'),
  array( __CLASS__, 'backups_list_postbox' ),
  'tools_page_backups_page',
  'side'
);
```

- If there is a need to add content before or after the postbox listing (e.g. admin notices), they can be appended by using the filters `'before_seravo_postboxes_' . $current_screen` or `'after_seravo_postboxes_' . $current_screen`, where `$current_screen` is the wanted admin screen (e.g. `tools_page_reports_page`).

Note that there is currently minimal CSS applied to these boxes, but we can in the future personalize these postboxes more in such a way that the user can separate them from WP core postboxes (I believe @nasusa has a design for this).

### CODE REVIEW

**Comment by mikkoaf** [comment]
PR looks good to me.
- Moving and closing postboxes works as intended and having notifications on the page works as well.

- This PR brings better maintainability, as separate *-page.php, *.js and *.css are not required for every tool-page.

- Hooking add_submenu_page appropriately we can move tools that are intended for network admins or equivalent to correct locations, which is a good thing.

- I think it would be a slight quality of life improvement if the postbox title used in `seravo_add_postbox` could be defined with the rest of the content, but this not necessary.

- Could you elaborate on why you added `before_seravo_postboxes_` . $current_screen or `after_seravo_postboxes_` . $current_screen filters?

**Comment by l3ku** [comment]
The idea for adding those filters is if we want to display some content above or below the postboxes, e.g. our own custom admin notifications. This way we could register the hook in a separate module (or even theme/plugin) that adds notifications above the postbox content on some page.

Of course, another option would be to provide a wrapper function in the call, e.g.
```
add_submenu_page( 'tools.php', __('Backups', 'seravo'), __('Backups', 'seravo'), 'manage_options', 'backups_page', array( this, 'seravo_postboxes_page_custom_wrapper' ) );
```
and call the `seravo_postboxes_page()` inside there, but this restricts the content to be added only from the same `seravo_postboxes_page_custom_wrapper()` function. 

I don't know which of these solutions is better, depends widely on future use cases. If it is not needed to view anything else apart from the postboxes on these pages, then the filters can be as well removed from the code.

l3ku:
Could @ottok also review this? I think this is a rather large architectural change and it would be nice to have some additional eyes look into the code.

ottok:
Small visual glitch:
image

Otherwise UI is good and this is an improvement both in UI terms and in code structure.

Question: can any of these custom postboxes be easily registered to be on the actual WordPress dashboard, or do they require additional code changes? According to https://codex.wordpress.org/Dashboard_Widgets_API the dashboard widget is just a function call that returns something. Can our custom postboxes functions be called in the dashboard registration context?

And should they be called instead of seravo_add_postbox() simply seravo_add_widget() to follow the WP terminology of calling them widgets?

ottok:
For the hooks regarding showing extra content before/after the widgets we could follow the example set by WordPress core itself, of having a hook to add full width elements before the widgets, eg.
image

l3ku:
> Small visual glitch:
Seems like some small CSS problem, fixing this.

> Question: can any of these custom postboxes be easily registered to be on the actual WordPress dashboard, or do they require additional code changes? According to https://codex.wordpress.org/Dashboard_Widgets_API the dashboard widget is just a function call that returns something. Can our custom postboxes functions be called in the dashboard registration context?

I think it is not possible to add our own custom postboxes there, as the current architecture requires registering a page that outputs a WP dashboard-like page with `Seravo_Postbox_Factory ::display_postboxes_page`. However, as we now have separated the functions that output the actual content of our postboxes, it should be simple to add them also as dashboard widgets by just passing the same content function to `add_dashboard_widget`.

> And should they be called instead of seravo_add_postbox() simply seravo_add_widget() to follow the WP terminology of calling them widgets?

This is probably a matter of taste, widget could be a better naming convention as our postboxes can be moved around and closed just like WP's dashboard widgets.

l3ku:
I have now fixed the HTTP request stats UI bug, and will merge this PR as "Seravo Postboxes", as all the word "widget" is not used anywhere in the HTML classes/IDs and would most likely cause too much of a hassle and bugs if renamed. I actually also like the name "Seravo Postbox".

PR #168, opened 2018-08-20
Update crufttheme remover

### LINKED ISSUE #167: JavaScript error is caused when no cruft themes are found
If there are no cruft themes found, nothing is displayed inside the cruft themes postbox, and the following JS error is caused:

screen shot 2018-08-15 at 16 35 15

mikkoaf:
This PR fixes #167 
Context for following images: TwentySeventeen is the active theme.
There is only one theme on the page, which is active.
image
There are some themes on the page. The active theme is not presented to the user for removal.
image
Avant-X as the active theme
image
Removing themes...
image
Removing themes...
image
SpicePress as the active theme
image
image
The page needs to be reloaded at this point to achieve the next view.
image

### DIFF
--- a/js/cruftthemes.js
+++ b/js/cruftthemes.js
@@ -1,6 +1,69 @@
 'use strict';
+
+function removeValue(list, value) {
+  list = list.split(',');
+  list.splice(list.indexOf(value), 1);
+  return list.join(',');
+}
+
+class CruftTheme {
+  constructor(name, title, active) {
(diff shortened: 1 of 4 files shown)


l3ku:
Now the cruft remover takes the active theme into consideration. UI-wise, an enhancement would be to view the removing of the parent-child relationship without reloading the page when removing the child theme. But the current functionality is good for now.

PR #173, opened 2018-08-27
Log view improvements

ottok:


l3ku:
Great! Now we can also link the dashboard "Site error count" widget directly to `php-error.log`. I also realised that there is a JavaScript bug when clicking the newly appended log line entries: they are not added to the log view preview box.

PR #174, opened 2018-08-28
Show screenshots from last update attempt on the Updates page

ottok:
image

### DIFF
--- a/lib/updates-page.php
+++ b/lib/updates-page.php
@@ -107,3 +107,70 @@ $site_info = Seravo\Updates::seravo_admin_get_site_info();
     </div>
   </div>
 </div>
+
+<h2 class="clear"><?php _e('Screenshots', 'seravo'); ?></h2>
+<?php
+$screenshots = glob( '/data/reports/tests/debug/*.png' );
+
+if ( count($screenshots) > 0 ) {
+
+  echo '
+<table>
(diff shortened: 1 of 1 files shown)


ottok:
Screenshot of UI when visual diff failed:
image

Commit rebased on latest master and added some more minor improvements.

l3ku:
Seems like a very neat feature! Some suggestions:
- Don't show area if screenshots are not available, or show indication to user that no screenshots were found and they will appear during the next update cycle
- Use gettext for outputting strings (e.g. lines 111-116 on updates-page.php)
- Maybe somehow in the UI separate that the middle row is the difference of the two, now it kind of looks like there are screenshots from three different instances instead of two
- Maybe add some borders between the images for better area separation, and so that it can be clearly seen that the diff percentage applies to the above image
- Add the screenshots inside full-width postbox hard-coded into the page template? I think it looks a little bit out of place when the other features above are in postboxes

Otherwise looks good. I tested this at dnstesti.ams-test.seravo.com, where the screenshots already existed, but it is not related to this PR how swd generates the images.

ottok:
@mikkoaf Great idea to make a slider UI! This first version if this page is very rudimentary and servers mostly as a proof of concept. I am glad if you can further improve this.

PR #176, opened 2018-09-06
Added missing filter that exists in documentation

tlehtimaki:
Added the missing filter. In seravo docs there is filter named seravo_use_purge_cache, but it's missing from the actual plugin file.

### DIFF
--- a/seravo-plugin.php
+++ b/seravo-plugin.php
@@ -132,8 +132,8 @@ class Loader {
      * Add a cache purge button to the WP adminbar
      */
     if ( apply_filters('seravo_use_purge_cache', true) ) {
-		  require_once dirname( __FILE__ ) . '/modules/purge-cache.php';
-	  }
+      require_once dirname( __FILE__ ) . '/modules/purge-cache.php';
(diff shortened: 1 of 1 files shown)


ottok:
Thanks for noticing that code and docs are not in sync! This filter was removed in ec1e18e2de2e93d565d92f5c327c07af373d5c3d. 
Now we have this filter instead which makes much more sense:
`modules/purge-cache.php:  return apply_filters( 'seravo_purge_cache_capability', 'edit_posts' );`.

Does the filter above fit your use case?

tlehtimaki:
It did cross my mind to look the history to check this. In my case I would've wanted to hide it away, but the case is quite "edgy" anyway, so this will have to do for now.  Thanks for the quick reply!

ottok:
@tlehtimaki Fixed docs in https://github.com/Seravo/docs/commit/35c5321c3e4ce07bfd85f008fccee71b6126474f

Note that our docs is also public on Github, and you can file issues or pull requests if you find something. We appreciate feedback!

PR #178, opened 2018-09-07
Fix leaky seravo-postbox padding styles

tnottu:
Padding from `.inside` class affects all postboxes, adding unnecessary padding to multiple places in the admin, the Publish-postbox for example. These styles are probably meant for seravo-postboxes only. Or maybe this css-file should only be enqueued on seravo-plugin pages?

There are similar risky styles in `seravo-plugin/style/updates.css` but apparently those styles are loaded only on the Updates-page.

### DIFF
--- a/style/seravo-postbox.css
+++ b/style/seravo-postbox.css
@@ -4,7 +4,7 @@
   }
 }
 @media only screen and (max-width:1445px) {
-  .inside {
+  .seravo-postbox .inside {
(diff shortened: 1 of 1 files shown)


ottok:
Thanks! We will look into this.

l3ku:
Thanks for noticing and making a PR! We will skim through the plugin code and fix possible other occurrences of this problem as described in https://github.com/Seravo/seravo-plugin/issues/180.

PR #181, opened 2018-09-21
Make "Purge Cache" button run full wp-purge-cache with AJAX (closes #83)

### LINKED ISSUE #83: Purge cache -button only purges front proxy cache, should run full wp-flush-cache instead
Currently the UI button does not do run wp-purge-cache (alias wp-flush-cache). All of our command line tools should have WP-admin equivalents, so users would expect this button and cli command to do the same thing.

l3ku:
Changes to the current mechansim:
- Run `wp-purge-cache` to also empty other caches than the Nginx front cache
- Also display failure message instead of merely success when in WP Admin, log errors to php-error.log
- Use AJAX instead of `wp_nonce_url` => Redirect from JS instead of PHP
- Include a neat cache spinning animation with keyframes :)

***Note:*** for this to work properly, the recent change involving the removation of `wp-restart-nginx` from `wp-purge-cache` must be available.

I haven't myself performed testing this on other sites than my own test site (where it works flawlessly), but will soon do that to find all the scenarios where the mechanism may fail.

EDIT: remember to also test this with other browsers than Chrome.

### DIFF
--- a/js/purge-cache.js
+++ b/js/purge-cache.js
@@ -1,0 +1,23 @@
+'use strict';
+jQuery(document).ready(function($) {
+
+  // Add event listener for clicking the purge cache button
+  $('#wp-admin-bar-nginx-helper-purge-all .ab-item').click(function(evt) {
+    evt.preventDefault();
+
+    // Show spinning animation
+    $('.seravo-purge-cache-icon', this).addClass('spin');
+
+    // Ask for server to empty cache via AJAX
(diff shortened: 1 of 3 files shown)


mikkoaf:
I've tested this locally and on my personal website using Firefox 61.0.1 and I didn't notice any major issues when `wp-purge-cache` was set up correctly.
Looks good to me!

PR #185, opened 2018-10-16
Install specific versions of PHPCS and WP Coding Standards in Travis

l3ku:
There have been now a couple of incidents in which the PHP code sniff has
been successful locally, but then failed in the remote Travis CI build to
to different versions of PHPCS and WP Coding Standards. Instead of cloning
the repos directly from GitHub, we should stick to specific releases and
update them periodically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,13 @@ matrix:
       env: SNIFF=1
 
 before_install:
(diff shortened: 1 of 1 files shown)


l3ku:
@simoke you should now be able to pass the Travis CI build on your branch https://github.com/Seravo/seravo-plugin/commits/fix-typo after rebasing with master.

PR #188, opened 2018-11-02
Whitelist our infrastructure in Jetpack

ypcs:
See commits for more details.


ypcs:
Note, code hasn't been tested at all after refactoring as part of Seravo plugin.

ypcs:
Did some testing outside WordPress using this (you can change `$ip` to confirm that non-whitelisted IPs return false, and whitelisted return true).

```php
<?php
define("ABSPATH", "x");
require_once("thirdparty-fixes.php");

function wp_cache_get($key, $group='', $maybe, &$found) {
   $found = false;
   return false;
}
function wp_cache_add($key, $group='', $data, $ttl) {
    return false;
}
function wp_remote_get($url) {
    return file_get_contents($url);
}
function wp_remote_retrieve_body($response) {
    return $response;
}
function esc_url_raw($url) {
    return $url;
}
function add_filter() {
    return false;
}


$m = new Seravo\ThirdpartyFixes();

$ip = '94.237.24.17';

$res = $m->jetpack_whitelist_seravo($ip);

var_dump($res);
echo chr(10);
?>
```

Still needs to be tested inside WordPress.

ypcs:
Latest version has been deployed to couple sites, now just waiting for results.

ypcs:
It seems that `$ip` that is passed to filter function is always empty. Why? Don't know yet. However, we use the filter [exactly as in Jetpack example](https://developer.jetpack.com/tag/jpp_allow_login/):

```php
<?php
function jetpackcom_let_me_in( $ip ) {
    if ( 'xxx.xx.xx.x' === $ip ) {
        return true;
    }
    return false;
}
add_filter( 'jpp_allow_login', 'jetpackcom_let_me_in', 10, 1 );
```

ypcs:
It seems that `apply_filters` in Jetpack source doesn't pass IP correctly to our filter function. So, I refactored filter to read the IP using Jetpack's internal function, if it exists. This patch has now been deployed to all sites we've noticed issues with.

PR #190, opened 2018-11-06
Add a GUI for editing zones

elguitar:
It may not be pretty, but it should work as expected!

Before merging, the internal API should be updated first.


ottok:
Skimmed through quickly, looks good!

ottok:
Crude but works!
image

PR #191, opened 2018-11-07
Add seravo.com to disallowed hosts

serter:


ottok:
This change would cover addresses like *.fi.seravo.com, *.de.seravo.com, *.se.seravo.com, *.fi-s.seravo.com but also www.seravo.com, but the last one is not in use so it should not matter. Also the other *.seravo.com we have are not WordPress sites and don't have the Seravo Plugin, right?

serter:
@ypcs comments?

PR #195, opened 2018-11-15
Use transients for caching API responses

ypcs:
This is more reliable than WP cache functions, that store responses only
temporarily if no object cache has been configured.


ypcs:
I've now tested this, proposed fix seems to fix the issue, ie. sites load the JSON payload only once per day.

PR #196, opened 2018-11-19
Fix 'pagenow not defined'

elguitar:
There was a notice that undefined pagenow broke Visual Composer.
This simple fix solves the problem.

### DIFF
--- a/js/seravo-postbox.js
+++ b/js/seravo-postbox.js
@@ -283,7 +283,9 @@ jQuery(document).ready(function($) {
   };
   ajaxPopulateWidgets();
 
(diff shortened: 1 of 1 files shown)


js/seravo-postbox.js
```diff
@@ -283,7 +283,9 @@ jQuery(document).ready(function($) {
   };
   ajaxPopulateWidgets();
 
-  seravo_postboxes.add_postbox_toggles(pagenow, { pbshow: ajaxPopulateWidgets } );
+  if ( typeof pagenow !== 'undefined ) {
```
ottok: Syntax error, ' missing?

elguitar: Fixed in the new commit.

PR #199, opened 2018-11-21
Fix domain zone editing

JoosuaKoskinen:
Fixes the following problems with domain zone editing:

- Editing does nothing (Because of 2 nested forms)
- API errors with subdomains (Subdomains don't have zones)
- Styling not updated everywhere
- No proper support for multiple pages (Always returned to 1st page)


ottok:
Skimmed through code. Looks good!
@elguitar Do you want to review this?

PR #201, opened 2018-11-26
Add PHP version checker

VeetiYliHemmo:
The check-php-version.php will check if the PHP version is below the recommendation (7.2.11). If the version is below the recommendation, it'll show a warning message on the WordPress dashboard.

### DIFF
--- a/modules/check-php-version.php
+++ b/modules/check-php-version.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * Plugin name: Seravo Check PHP version
+ * Description: Checks that the PHP version is supported. If the version is lower than recommended,
+ * it displays a warning on the dashboard.
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+
+  die('Access denied!');
+
+}
+
+if ( ! class_exists('CheckPHPVersion') ) {
+
+  class CheckPHPVersion {
+
+    public static function load() {
+
+      add_action('admin_notices', array( __CLASS__, '_seravo_check_php_version' ));
+
+    }
+
+    public static function _seravo_check_php_version() {
+
+      // Get the php version and check if it is supported, if not, show a warning
+
+      $recommended_version = '7.2';
+
+      if ( version_compare( PHP_VERSION, $recommended_version, '<' ) ) {
+
+        self::_seravo_show_php_warning( $recommended_version );
+
+      }
+
+    }
+
+    public static function _seravo_show_php_warning( $recommended_version ) {
+
+	  ?>
+		<div class="notice notice-error">
+	  <?php
+
+        printf(
+			// translators: %1$s: current php version, %2$s: recommended php version
+			__('The PHP-version %1$s currently in use is lower than the recommended %2$s. The security support for versions below the recommendation is ending.
+			<a target="blank" href="https://help.seravo.com/en/knowledgebase/13-server-configuration/docs/107-set-your-site-to-use-newest-php-version">More information</a>', 'seravo'),
+			PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION, $recommended_version
+		);
+
+	  ?>
+		</div>
+	  <?php
+
+    }
+
+  }
+
+  CheckPHPVersion::load();
+
+}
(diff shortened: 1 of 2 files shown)


modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
...
+
+    public static function _seravo_check_php_version() {
+
+      // Get the php version and check if it is 7.2, if not, show a warning
+
+      preg_match('#^\d+(\.\d+)*#', PHP_VERSION, $current_version);
+
+      $recommended_version = '7.2.11';
+
+      if ( version_compare( $current_version[0], $recommended_version ) == -1 ) {
```
ypcs: It seems that you just compare if current version is (exactly) the recommended version? Perhaps you should just check if current version is less than recommended version?

See PHP documentation for `version_compare()`, there is some good examples.

modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
...
+
+    }
+
+    public static function _seravo_check_php_version() {
+
+      // Get the php version and check if it is 7.2, if not, show a warning
+
+      preg_match('#^\d+(\.\d+)*#', PHP_VERSION, $current_version);
+
+      $recommended_version = '7.2.11';
```
ypcs: We don't care about patch versions, so probably you should compare either against `7.2` or `7.2.0`.

modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
...
+
+      add_action('admin_notices', array( __CLASS__, '_seravo_check_php_version' ));
+
+    }
+
+    public static function _seravo_check_php_version() {
+
+      // Get the php version and check if it is 7.2, if not, show a warning
+
+      preg_match('#^\d+(\.\d+)*#', PHP_VERSION, $current_version);
```
ypcs: Is this really required?

See PHP documentation for `version_compare()`, https://secure.php.net/manual/en/function.version-compare.php#example-570 and please try results when just using `PHP_VERSION` directly as parameter for forementioned function. It should work.

modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public static function _seravo_show_php_warning( $current_version, $recommended_version ) {
+
+	  ?>
+		<div class="notice notice-error">
+	  <?php
+
+        printf(
+			// translators: %1$s: current php version, %2$s: recommended php version
```
ypcs: :+1:

modules/check-php-version.php
ypcs: By skimming through full code, it seems that you display version to user, so parsing just `MAJOR.MINOR.PATCH` might be justified & good thing to do. However, you can still use `PHP_VERSION` for the comparison itself.

modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
+ * Plugin name: Seravo Check PHP version
+ * Description: Checks that the PHP version is 7.2. If the version is lower than 7.2
```
ypcs: Perhaps instead of using `7.2` everywhere, you could just say something like "checks that PHP version is supported one" or something like that... This is however very minor issue, as later we can just use `sed` to replace the version with another one.

modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
...
+ * it displays a warning on the dashboard.
+ */
+
+namespace Seravo;// Deny direct access to this file
+
+if ( ! defined('ABSPATH') ) {
```
ypcs: :+1:

modules/check-php-version.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
+ * Plugin name: Seravo Check PHP version
+ * Description: Checks that the PHP version is 7.2. If the version is lower than 7.2
+ * it displays a warning on the dashboard.
+ */
+
+namespace Seravo;// Deny direct access to this file
```
ypcs: This comment is related to `!defined` block below.

ypcs:
Finally tested this, and seems to work. If PHP version is 5.6 or 7.0, nag is displayed, and if version is 7.2, nag is gone. Merging now.

screenshot from 2018-11-29 23-08-15

PR #204, opened 2018-11-30
Prevent committing executable PHP files

ypcs:
We don't really want executable files into master. These have slipped through couple times, so lets prevent these now.


ypcs:
I did testing for Travis test in f4bf081: if I try to add commit with executable PHP file after that, tests will fail. If I remove execute bit, tests will succeed again.

Issue #203, opened 2018-11-30
Domain zone edit should detect if zone isn't manageable


ypcs:
Plugin should perhaps tell explicitly why editing isn't possible. In this case, DNS isn't managed by Seravo, but plugin shows view/edit links for the zone anyways.

![screenshot from 2018-11-30 11-22-16](https://user-images.githubusercontent.com/97568/49280238-5566c080-f492-11e8-8970-c1a7c273e54d.png)

It would be preferred to tell the real status to the user, "DNS not managed by Seravo, see more details" with link to some help article. Also, if editing isn't possible, edit/view links should be hidden.

elguitar:
Fixed via f2c34308af7e7b84f1fe978671c8ac851668ecd7. Closing

PR #206, opened 2018-12-05
Zone edit improvements

JoosuaKoskinen:
Adds primary domain indicator.

Removes zone actions for domains that can't use them (subdomains and domains not managed by Seravo) and shows message instead. Suggested in #203 

Replaces the default zone edit submit button with a proper one.

### DIFF
--- a/lib/domains-page.php
+++ b/lib/domains-page.php
@@ -51,7 +51,7 @@ class Seravo_Domains_List_Table extends WP_List_Table {
 
   public function column_domain( $item ) {
 
-  	$actions = array();
+    $actions = array();
 
     /*
     // Domains managed by Seravo can be added, edited or deleted
@@ -62,7 +62,7 @@ class Seravo_Domains_List_Table extends WP_List_Table {
     // Domains managed by customers themselves can only be added, viewed or deleted
     // $actions['delete'] = sprintf( $action_request, 'delete', 'Delete');
     */
-        
+
     $action_request = '<a href="?page=' . $_REQUEST['page'] . '&domain=' . $item['domain'] . '&paged=' . $_REQUEST['paged'] . '&action=%s">%s</a>';
 
     $actions['view'] = sprintf( $action_request, 'view', 'View' );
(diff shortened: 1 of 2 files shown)


lib/domains-page.php
```diff
@@ -62,18 +62,20 @@ public function column_domain( $item ) {
...
+
     $action_request = '<a href="?page=' . $_REQUEST['page'] . '&domain=' . $item['domain'] . '&paged=' . $_REQUEST['paged'] . '&action=%s">%s</a>';
 
     $actions['view'] = sprintf( $action_request, 'view', 'View' );
     $actions['edit'] = sprintf( $action_request, 'edit', 'Edit' );
 
     $primary_str = ! empty( $item['primary'] ) ? ' — ' . __( 'Primary Domain', 'seravo' ) : '';
 
-    return sprintf(' <strong class="row-title">%1$s<small>%2$s</small></strong> %3$s',
+    $fetch_result = $this->dns->managed_by_seravo( $item['domain'] );
```
ottok: Are you sure this is a good idea If you do an API request for every domain like this it will snowboll into tens or hundreds of requests in the background and create unnecessary domain zone updates on every pagelod. We need to postpone this feature until the SWD API actually returns a list of domains that are managed.

JoosuaKoskinen: I have thought about that too but I haven't found any other way so far.

Does anyone know any less intensive way to do a temporary fix for that?

elguitar: The internal API is now implemented in a PR in another repo.
The domains page fetches it already when populating the domains table. Maybe it should be saved as an transient and then reused when needed?

elguitar:
Looks good to me! This is a really big step when thinking about UX.
The only concern I have is the same as otto said. The internal API is coming soon which could be used to limit the amount of API requests. Maybe this should wait for it?

JoosuaKoskinen:
This uses the 'management' field from the API now so these changes won't make any new requests.

Issue #198, opened 2018-11-20
Optimize images -page instructions unclear


ottok:
The current UI does not explain enough how it works:
![image](https://user-images.githubusercontent.com/668724/48777165-8993f180-ecda-11e8-8274-01cd3fef4e9d.png)

- Link to docs is missing
- Does not explain that recompressing images is possible even without restricting resolution
- Does not explain that recompression does not happen unless box is ticked
- Does not explain that optimization is done on files on disk and only once per file

ottok:
WIP: https://github.com/Seravo/seravo-plugin/tree/dev-maf

mikkoaf:
![image](https://user-images.githubusercontent.com/16817737/49847910-224df680-fddb-11e8-9ac2-bd39a71b99a6.png)
I have improved the usability of the module with UI-changes in conjunction with the Helpy-articles written.

ottok:
At a first look, this text does not explain why, what and how. It is better than before, but please but a bit more thought into it. Maybe ask others for ideas on what should be said here if you yourself don't have new ideas?

The article https://help.seravo.com/fi/knowledgebase/23-wordpressin-hallinta/docs/119-seravo-plugin-optimoi-kuvia does a slightly better job, but there is still relevant details missing. Try to put yourself in the position of a website owner trying to decide whether or not to tick that box, and is is risky for them or not.

Also we need some kind of feature description page on seravo.com about what image optimization is, that we have it and why users should choose it. And then there are the man pages in the command itself, and seravo.com/docs. Many places that all need to tell the same story, but not have duplicate content..

mikkoaf:
Revised the description text according to given feedback. 
![image](https://user-images.githubusercontent.com/16817737/50088002-6e999c00-020a-11e9-8ac3-97c6560c1dc2.png)
Added less disrupting logic for handling empty input.
Previously empty input fields would return values of 500, which is the agreed minimum for image optimization.
With the current iteration the default value when user begins to use the tool is 4000 and when giving empty input the value is 4000

Helpy article has been also revised so that information relevant to website owners should be better available.

A blog post has has been written which describes the benefits of image optimization.

tlxo:
That much text feels a bit heavy for a UI. I'd limit the text to the amount we currently feature and link to user documentation for details.

mikkoaf:
Closed in #211

PR #211, opened 2018-12-18
Improve Optimize images -page instructions

### LINKED ISSUE #160: Optimize Images-tool should be more informative
In the optimize Images-tool should be information, why you can't optimize images if you do not agree to restrict the resolution of images.

is a bit confusing when the page does not show inputs at all, maby inputs should be disabled, or there could be text "Notice: you can't optimize images without restricting image resolution."
---- OR EVEN BETTER: both ----

image

mikkoaf:
Currently the Optimize images page is ambiguous to users.
This PR will
- Improve the UX of the option by disabling the textboxes rather than hiding them.
- Add a more understandable description text to the option
- Return 4000px in case user inputs a value lower than 500px or none
- Omit other than Multisite Super Admins choose the option

Closes #160, #198


ottok:
Seems translations are missing.

mikkoaf:
Updated the .pot and added translations.

PR #212, opened 2018-12-18
Improve UI on mobile devices

VeetiYliHemmo:
The commit will make the purge cache and instance switcher features available for mobile devices. It'll also remove the text and only show the icon for the features when the screen size goes below a certain amount.


ottok:
Screenshot?

VeetiYliHemmo:
Here you can see what the admin bar looks on mobile with the changes:
image

Here is the menu open:
image

The icon is temporary and can be changed to a more suitable one easily.

ottok:
Looks good!

Skimmed though code, but did not test by running it. Merging now however based on screenshot.

PR #213, opened 2018-12-19
Add image comparison slider to Update Attempt images

mikkoaf:
Purpose of this pull request is to enhance the user experience of the Seravo-plugin Updates page.
Currently the page has three images side by side, with little logic regarding missing image handling.
Current behaviour:
image

With this PR if only one of the image pair is found, no comparison is made. If all images pairs are partially missing, the following is shown to the user:
image
Visually inspected the comparison should look something like this:
image


ottok:
The pictures are now huge and one picture does not fit the screen at one, so the comparison is not as convenient as it could be. Now also the images that emphasize the differences in yellow are not used at all anywhere, maybe they should be somewhere as a reference image to highlight what changed?

PR #214, opened 2018-12-21
Add mail forwards page

JoosuaKoskinen:
Adds a page that lists all email addresses with forwards for a specific domain.

mails-page


mikkoaf:
In a situation that the site does not have any domains, there should be some sort of message that the process was completed successfully, but no domains were found.
image

Handling empty results is handled correctly here:
image

Otherwise the tools seems good and well designed.
Results such as described in the PR:
image

Issue #180, opened 2018-09-07
Review plugin CSS for leaky styles and unnecessary usage of !important


l3ku:
Some of our clients reported in e.g. https://github.com/Seravo/seravo-plugin/pull/178/commits/c9ba30f4f031e5bf7c01ddfd1f1566ba75b19964 and https://github.com/Seravo/seravo-plugin/issues/179 that there was a too broad style definition in `seravo-postbox.css`, which conflicted with the styles of other WP admin functionalities. While this is a single case that was reported, we should while at it review the CSS for the whole plugin and check for too broad / leaky style definitions so that this will not happen in the future.

@nasusa and @tlxo you are probably way more competent with CSS than I am, so assigning this issue to you guys... :)

PR #216, opened 2018-12-29
Clean the code (Closes: #180)

nasusa:
- I deleted unnecessary !important tags
- I added more specific classes when necessary
- I deleted conflicting classes/styles
- I deleted unnecessary tabs
- I cleaned the code also by adding line breaks after every class, so style-files could be readed more easily


ottok:
Thanks for cleaning up the code. All changes look sensible, so I will merge this without testing (also because I don't have time to test everything).

Good to see you squashed all commits and the final commit message is in imperative format "Clean" instead of the WIP commits that had "I cleaned..". You are on a good path to become a professional developer with high standards. Keep it up!

Issue #218, opened 2019-01-06
PHP version warning has wrong link


ottok:
I noticed that this PHP version warning that is shown to user links to generic site http://php.net/supported-versions.php
![image](https://user-images.githubusercontent.com/668724/50739014-9d3fdc80-11e2-11e9-8c74-70d6fadaec32.png)

There is nothing indicating to the user that this notification stems from the Seravo Plugin, nor info on what to do about this. We even had customers contacting us asking if we support PHP 7.2 or not, because this message looks like a built-in WP message.

ottok:
Any takers, @mikkoaf @ypcs

ottok:
New version:
![image](https://user-images.githubusercontent.com/668724/51101779-c9f88300-17e4-11e9-9f07-0e02b44da2f5.png)

PR #219, opened 2019-01-07
Update PHP version warning message

mikkoaf:
Currently the PHP version checking module in the Seravo-plugin is reaching users effectively, but the link in the notification and the link itself are far too generic.
image

To ensure that users know where the notification stems from and what actions they have available, the message could be as follows: (PHP version recommendation exaggerated for demonstration purposes)
image

The link leads to the [knowledgebase](https://help.seravo.com/en/knowledgebase/13/docs/107-uusimpaan-php-versioon-siirtyminen) article in the respective language.

The image used is an inline svg image for added performance.

The added notification style loading functionalities are somewhat similar to the existing postbox-module loading. This could be implemented later into a more generic notification building.

Suggestions and help on styling is much appriciated.

Related to #218

### DIFF
--- a/lib/seravo-notification.php
+++ b/lib/seravo-notification.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * File for Seravo custom notification functionality.
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+if ( ! class_exists('Seravo_Notification') ) {
+  class Seravo_Notification {
+
+    /**
+     * Instance of this class.
+     */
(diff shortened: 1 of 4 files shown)


ottok:
LGTM!

nasusa:
Could the icon be on top, in mobile? Now there is much unused space above and below the icon. Also, in some screen sizes it has different width than postboxes and other notifications. Could it be the same width?

**Notification after this pr:**
image


**My suggestion:**
image

ottok:
Closed via https://github.com/Seravo/seravo-plugin/commit/16beb6546358988682ab0501c3fd555af577a6db which is a simpler, to-the-point solution without any logo placement issue drawbacks.

PR #223, opened 2019-01-26
Mails page improvements

JoosuaKoskinen:
Makes the following improvements to the mails page:

- Fixes problem told by @mikkoaf in #214 
- Adds a better design using postbox to make the page better looking and more responsible

mails-page

### DIFF
--- a/lib/mails-page.php
+++ b/lib/mails-page.php
@@ -79,11 +79,11 @@ class Seravo_Mails_Forward_Table extends WP_List_Table {
         return;
     }
 
-      // Fetch the mail data
-      $api_query = '/domain/' . $url . '/mailforwards';
+		// Fetch the mail data
+  	$api_query = '/domain/' . $url . '/mailforwards';
 	  $fetch_data = Seravo\API::get_site_data($api_query);
-    if ( is_wp_error($data) ) {
-      die($data->get_error_message());
+    if ( is_wp_error($fetch_data) ) {
+      die($fetch_data->get_error_message());
     }
 	  $data = $fetch_data['forwards'];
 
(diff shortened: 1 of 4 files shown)


lib/mails-page.php
```diff
@@ -127,22 +127,32 @@ function list_domains() {
...
 
   // Render the list
-  echo '<select name="domain" style="min-width: 150px;">';
-  foreach ( $data as $row ) {
-  	$management = $row['management'];
-  	if ( $management == 'Seravo' || $management == 'Customer' ) {
-  		$domain = ! empty( $_GET['domain'] ) ? $_GET['domain'] : '';
-  		printf('<option value="%1$s" %2$s>%1$s</option>', $row['domain'], $domain == $row['domain'] ? 'selected' : '');
-  	}
+  if( ! empty($valid_data) ) {
```
ottok: Space missing between `if` and `(` - in general

lib/mails-page.php
```diff
@@ -127,22 +127,32 @@ function list_domains() {
...
+  	echo '<select name="domain" style="min-width: 150px;">';
+	foreach ( $valid_data as $row ) {
+	  $domain = ! empty( $_GET['domain'] ) ? $_GET['domain'] : '';
+	  printf('<option value="%1$s" %2$s>%1$s</option>', $row['domain'], $domain == $row['domain'] ? 'selected' : '');
+	}
+	echo '</select>';
+	echo '<input type="submit" value="' . __('Fetch Forwards', 'seravo') . '" class="button" style="margin-left: 15px;">';
+  } else {
+  	echo __("No valid domains were found!");
```
ottok: Language domain 'seravo' missing.

JoosuaKoskinen: This has been fixed.

lib/mails-page.php
```diff
@@ -127,22 +127,32 @@ function list_domains() {
     die($data->get_error_message());
   }
 
+  // Parse valid domains
```
ottok: What are the cases when the domain are invalid? Should we maybe fix something in our API? I am a bit afraid that this will hide problems that should be addressed and not ignored.

ottok:
Looks good!
image

It seems the "Parse valid domains" part considers valid domains that are real top-level domains, eg. example.com but not sub.example.com.

lib/mails-page.php
JoosuaKoskinen: The API answers with error when curl fails or gives invalid HTTP code. In this case it should only happen if the given domain is not part of the site.

Probably doesn't need changes in the API but I think the error should be handled better than with die() as the requested domain can easily be changed as URL parameter etc.

JoosuaKoskinen:
@ottok Should subdomains be considered valid? Currently the API responds with error 'Subdomains don't have their own forwards'.

ottok:
Why? Can subdomains have emails?

JoosuaKoskinen:
> Looks good!
> image
> 
> It seems the "Parse valid domains" part considers valid domains that are real top-level domains, eg. example.com but not sub.example.com.

I didn't quite understand from this that is it correct now now. But I think overall, subdomains can have emails.

ottok:
OK, now I understood your questions. Yes, in general any domain that has a MX record can have emails, including subdomains. However in this UI we use the Gandi API as the backend and there subdomain emails are not supported.

Did a test and the result was negative:
> 554 5.7.1 <test@mailtest.seravo.net>: Relay access denied

PR #225, opened 2019-02-14
Add documentation on postboxes

### LINKED ISSUE #224: Document Seravo postboxes
I tried creating a new Seravo postbox but the result looks very ugly. Apparently adding normal <p> tags and content does not result in what I would expect. There was also an awful lot of boilerplate code.

I was thinking maybe I am using it incorrectly? Thus I hope somebody would document in the project README how the Seravo postboxes are supposed to be used so that next time I create a new one the HTML markup, CSS and PHP are correct and end result is something nice out-of-the-box instead of this:
image

mikkoaf:
### Changes in this PR
This PR adds Seravo-plugin documentation describing the use of Postboxes. This documentation enables developers to create cleaner looking admin tools easily using Postboxes.

### Reviewing this PR
The documentation should be clear, understandable and following it should allow developers to create new Postboxes.

Closes #224

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,76 @@
+# Seravo-plugin developer documentation
+
+This readme is to help in developing, maintaining and updating Seravo-plugin.
+
+## Modularity
+
+Features and tools in the Seravo-plugin are divided into modules. This simplifies developing new features and maintaining existing ones.
+Tools provided by the Seravo-plugin are divided into modules within the /modules directory.
+
+### Adding a new page
+
+If you want to use postboxes in a submenu page, use `'Seravo\seravo_postboxes_page'` as the `$function` parameter for `add_submenu_page` when registering the submenu page. With the following example, you can add a submenu under `Tools`.
+
+```
+add_submenu_page('tools.php',
+                 __('Example Tools', 'seravo'),
+                 __('Example Tools', 'seravo'),
+                 'manage_options',
+                 'example_tools_page',
+                 'Seravo\seravo_postboxes_page' );
+```
+
+## Postboxes
+
+Postboxes were introduced to the Seravo-plugin in [#166](https://github.com/Seravo/seravo-plugin/pull/166).
+Postboxes are elements used in the Seravo-plugin Admin tools that behave similarily to the Dashboard widgets in WordPress. They can be dragged in custom orders and opened/closed. The postbox location and opened/closed state is saved to the site database for each individual user to enable custom settings per-user.
+Using postboxes gives the plugin a more unified, professional look.
+
+If there is a need to add content before or after the postbox listing (e.g. admin notices), they can be appended by using the filters `'before_seravo_postboxes_' . $current_screen` or `'after_seravo_postboxes_' . $current_screen`, where `$current_screen` is the wanted admin screen (e.g. `tools_page_reports_page`).
+
+
+### Adding a postbox
+
+Adding a postbox to the Example Tools -page which uses `'Seravo\seravo_postboxes_page'` as a function for outputting the page content can be done using
+
+`modules/example-tool.php`
+```
+  seravo_add_postbox('example_tool_id',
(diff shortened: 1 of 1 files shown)


ottok:
Looks good to me!

Please ask Samuli and Leo for comments or feel free to merge as is.

docs/README.md
```diff
@@ -0,0 +1,99 @@
+# Seravo-plugin developer documentation
+
+This readme is to help in developing, maintaining and updating Seravo-plugin
```
l3ku: Missing period at the end of this sentence.

docs/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Modularity
+
+Features and tools in the Seravo-plugin are divided into modules. This simplifies developing new features and maintaining existing ones.
+Tools provided by the Seravo-plugin are divided into modules within the /modules directory.
+
+### Adding a new page
+
+If you want to use postboxes in your new submenu page under `Tools`, use `'Seravo\seravo_postboxes_page'` as the `$function` parameter for `add_submenu_page` when registering your submenu page.
```
l3ku: Maybe emphasize that the Postboxes can be used outside `Tools` too, but this example just happens to add one to `tools.php`? This way the reader doesn't get an impression that they are only restricted to tools.php.

docs/README.md
```diff
@@ -0,0 +1,99 @@
...
+                 __('Example Tools', 'seravo'),
+                 __('Example Tools', 'seravo'),
+                 'manage_options',
+                 'example_tools_page',
+                 'Seravo\seravo_postboxes_page' );
+```
+
+## Postboxes
+
+Postboxes were introduced to the Seravo-plugin in #166
```
l3ku: Add a period here, also there could be a link to the PR/commit (the hashtag doesn't automatically generate one at least for me).

docs/README.md
```diff
@@ -0,0 +1,99 @@
...
+Postboxes are elements used in the Seravo-plugin Admin tools that behave similarily to the Dashboard widgets in WordPress. They can be dragged in custom orders and opened/closed. The postbox location and opened/closed state is saved to the site database for each individual user to enable custom settings per-user.
+Using postboxes gives the plugin a more unified, professional look.
+
+If there is a need to add content before or after the postbox listing (e.g. admin notices), they can be appended by using the filters `'before_seravo_postboxes_' . $current_screen` or `'after_seravo_postboxes_' . $current_screen`, where `$current_screen` is the wanted admin screen (e.g. `tools_page_reports_page`).
+
+### Postboxes creation
+
+Postboxes are
```
l3ku: Maybe still WIP?

docs/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+### Adding a postbox
+
+Adding a postbox to the Example Tools -page which is a `'Seravo\seravo_postboxes_page'` can be done using
```
l3ku: Maybe rephrase, e.g. "which uses `'Seravo\seravo_postboxes_page'` as a function for outputting the page content".

docs/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+    <!-- Postbox content -->
+    <div class="inside">
+      <div class="seravo-section">
+        <?php call_user_func_array($postbox_content['callback'], $postbox_content['callback_args']); ?>
+      </div>
+    </div>
+
+```
+It's not adviseable to use class `inside` in your Postbox content.
```
l3ku: What if we modify the CSS selectors as `.seravo-postbox > .inside`, so that our CSS specification only applies to the elements with `.inside` class that have the `.seravo-postbox` as immediate parent (`.inside` is directly under `.seravo-postbox` in DOM structure)? Then this side note could be removed, and there would be less confusion amongst developers.

mikkoaf: Sounds reasonable. Using CSS selectors that are more specific would make development easier.

PR #226, opened 2019-02-23
Add null-checking to suppress PHP-warnings

JoosuaKoskinen:
Adds null-checking to domain and log -pages. These
pages otherwise give PHP-warnings with WP_DEBUG enabled.

Logs-page has been designed so that it doesn't try to
use log files that don't exists. But always using
'php-error.log' as default didn't verify its existence.

### DIFF
--- a/lib/domains-page.php
+++ b/lib/domains-page.php
@@ -246,8 +246,12 @@ class Seravo_Domains_List_Table extends WP_List_Table {
         $this->dns->display();
       } else if ( 'edit' === $this->current_action() ) {
         $this->dns->display_edit();
-      } elseif ( $_REQUEST['zone-updated'] ) {
-        $this->dns->display_results( $_REQUEST['modifications'], $_REQUEST['error'] );
+      } elseif ( isset( $_REQUEST['zone-updated'] ) ) {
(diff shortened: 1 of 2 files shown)


lib/domains-page.php
```diff
@@ -246,8 +246,12 @@ public function single_row( $item ) {
...
       } else if ( 'edit' === $this->current_action() ) {
         $this->dns->display_edit();
-      } elseif ( $_REQUEST['zone-updated'] ) {
-        $this->dns->display_results( $_REQUEST['modifications'], $_REQUEST['error'] );
+      } elseif ( isset( $_REQUEST['zone-updated'] ) ) {
+        if ( $_REQUEST['zone-updated'] === 'true' ) {
+          $modifications = isset( $_REQUEST['modifications'] ) ? $_REQUEST['modifications'] : false;
+          $error = isset( $_REQUEST['error'] ) ? $_REQUEST['error'] : false;
+          $this->dns->display_results( $modifications, $error );
+        }
```
ottok: Is this intentionally part of this pull request? The commit message does not mention zone editing.

JoosuaKoskinen: It is. I only said 'domain page' because there wasn't anything special about that. Just missing the isset() part.

PR #230, opened 2019-03-29
New security page

ottok:
A picture tells more than a thousand words..
image

### DIFF
--- a/lib/security-restrictions.php
+++ b/lib/security-restrictions.php
@@ -1,0 +1,71 @@
+<?php
+/*
+ * Description: Custom redirects to enforce primary domain or https
+ *
+ * This module should be called as early as possible so that the redirect
+ * happens as quickly as possible and no other parts of the WordPress are
+ * initialized in vain.
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+if ( ! class_exists('Security_Restrictions') ) {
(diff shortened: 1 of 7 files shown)


lib/security-restrictions.php
```diff
@@ -0,0 +1,69 @@
...
+          if ( isset($endpoints['/wp/v2/users']) ) {
+            unset($endpoints['/wp/v2/users']);
+          }
+          // Disable fetching a single user
+          if ( isset($endpoints['/wp/v2/users/(?P<id>[\d]+)']) ) {
+            unset($endpoints['/wp/v2/users/(?P<id>[\d]+)']);
+          }
+          return $endpoints;
+        }
+        add_filter('rest_endpoints', 'disable_user_endpoints', 1000);
```
JoosuaKoskinen: Should probably be:
`add_filter('rest_endpoints', 'Seravo\disable_user_endpoints', 1000);`

Currently the `disable_user_endpoints` function is not called at all and the error from it causes all the endpoints to break, not only `/wp/v2/users`.

ottok: Thanks for noting! I did not test the site with wp-cli at all.

Namespacing is now fixed.

ottok:
@JoosuaKoskinen OK now? Will you finish the review with merge?

JoosuaKoskinen:
> @JoosuaKoskinen OK now? Will you finish the review with merge?

Yep, seems to be working now.

PR #236, opened 2019-05-08
Remove WP Core Site Health update check

j-falk:
Closes wpd #390.


ypcs:
Looks good! I pushed updated version, fixed those two tabs (=replaced with spaces), but code itself seems fine. :) ... Will test soon & then merge.

ypcs:
Does what it's intended to do! :) ... Now just waiting for tests to pass.

Screenshot from 2019-05-08 14-05-45

Issue #182, opened 2018-09-21
Implement ajax for domain zone editting


elguitar:
This is a subtask #12.

This should be done before making the frontend. It should be kind of a MVP of the final product which @nasusa can finnish after :) 

I will assign myself here.

elguitar:
Hmmmm... Right now the implementation is using GET and POST requests, which works fine. Using AJAX would be cool and all but I don't think that it is a priority.
Does @JoosuaKoskinen have some input on this?

JoosuaKoskinen:
Not really. It sure would be nice to have. The URL can get quite ugly with the modification responds etc.

eg. wp-admin/tools.php?page=domains_page&domain=pisara.eu&paged=1&zone-updated=true&modifications%5B0%5D=Changed+the+origin+A+and+AAAA+records+to+match+the+servers+addresses

l3ku:
What about just making the GET/POST requests using `jQuery.get()` and `jQuery.post()` and dynamically modifying the DOM content based on the results and populating the data to the correct locations by CSS selectors? Of course this would require coding some JavaScript... :)

Mainerii:
Should just the zone editing use AJAX or the whole domains-page (eg. records-table on 'view' link click)? Fetching the records with AJAX should be implemented anyways so the zone editing textarea can be updated to sanitized version after each zone update.

elguitar:
Closed via 627b5068dd1727c4f1155694437bfbc51fc878ba

PR #237, opened 2019-05-08
Implement zone editing using AJAX

JoosuaKoskinen:
Implements zone editing using AJAX (as originally planned in #182) and makes it more user friendly.

Improvements:
- No need to load the page again (Can be annoying with larger domain lists).
- Don't need to start over when trying to save invalid zone.
- Gets sanitized version of zone right after saving.
- Disables editing with errors (eg. Record fetch failed) to make accidentally breaking zone harder.

### DIFF
--- a/js/domains.js
+++ b/js/domains.js
@@ -1,0 +1,133 @@
+'use strict';
+
+jQuery(document).ready(function($) {
+
+  var update_zone_button = jQuery('#update-zone-btn');
+
+  update_zone_button.click(function (event) {
+
+    event.preventDefault();
+
+    var response_div = jQuery("#zone-edit-response");
+    var spinner_div = jQuery("#zone-update-spinner");
+
+    jQuery("#zone-fetch-response").html("");
+
+    response_div.html("");
+    spinner_div.html("<img src=\"/wp-admin/images/spinner.gif\">");
+
+    jQuery.post(seravo_domains_loc.ajaxurl, {
+      'action': 'seravo_ajax_domains',
+      'section': 'update_zone',
+      'domain': $("input[name='domain']").val(),
+      'compulsory': $("textarea[name='compulsory']").val(),
+      'zonefile': $("textarea[name='zonefile']").val(),
+      'nonce': seravo_domains_loc.ajax_nonce,
+    },
+
+    function (rawData) {
+
+      if (rawData == 0) {
+        // If eg. AJAX callback not found
+        response_div.html('<p><b>No data returned for the update request.</b></p>');
+
+      } else {
+
+        var response = JSON.parse(rawData);
+
+        if (response['status'] && response['status'] === 400) {
+
+          response_div.html("<p><b>" + seravo_domains_loc.zone_update_failed + ":</b></p><p>" + response['reason'] + "</p>");
+
+        } else {
+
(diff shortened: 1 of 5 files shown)


lib/domains-dns.php
```diff
@@ -45,78 +45,45 @@ public function display_edit() {
     if ( ! isset( $this->records ) ) {
       return;
     }
-    if ( $this->records['pending_activation'] ) {
+
+    $error = isset( $this->records['error'] );
+    $domain = $error == false ? $this->records['name'] : '';
```
elguitar: This would benefit from a comment which would explain what is assigned to $domain and when

js/domains.js
```diff
@@ -0,0 +1,138 @@
...
+      'section': 'update_zone',
+      'domain': $("input[name='domain']").val(),
+      'compulsory': $("textarea[name='compulsory']").val(),
+      'zonefile': $("textarea[name='zonefile']").val(),
+      'nonce': seravo_domains_loc.ajax_nonce,
+    },
+
+    function (rawData) {
+
+      if (rawData.length == 0) {
```
elguitar: Is there a real danger of these happening?

They are okay here of course! Just wondering if there is a bug elsewhere :thinking:

js/domains.js
```diff
@@ -0,0 +1,138 @@
...
+        response_div.html('<p><b>No data returned for the update request.</b></p>');
+      } else {
+
+        var response = JSON.parse(rawData);
+
+        if (response != 0) {
```
elguitar: Same as above ^

js/domains.js
JoosuaKoskinen: This should have just been `rawData == 0` as that's what it might return. Fixed!

js/domains.js
JoosuaKoskinen: Couldn't find a way to make that statement true. It has been removed as useless.

lib/domains-dns.php
JoosuaKoskinen: I modified this part a bit because it seemed like a leftover. Also added a comment explaining all the ternary operations of the section.

lib/domains-dns.php
```diff
@@ -45,78 +45,47 @@ public function display_edit() {
...
     }
-    if ( $this->records['pending_activation'] ) {
+
+    $error = isset( $this->records['error'] );
+
+    if ( ! $error && $this->records['pending_activation'] ) {
       echo '<hr>';
       // translators: %s domain of the site
-      echo '<p style="max-width:50%;">' . wp_sprintf( __( 'Our systems have detected that <strong>%s</strong> does not point to the Seravo servers. For your protection, manual editing is disabled. Please contact the Seravo customer service if you want changes to be done to the zone in question. You can publish the site yourself when you so desire with the following button:', 'seravo'), $this->records['name'] ) . '</p>';
+      echo '<p style="max-width:50%;">' . wp_sprintf( __( 'Our systems have detected that <strong>%s</strong> does not point to the Seravo servers. For your protection, manual editing is disabled. Please contact the Seravo customer service if you want changes to be done to the zone in question. You can publish the site yourself when you so desire with the following button:', 'seravo'), $domain ) . '</p>';
```
elguitar: I believe that `$domain` is not defined.

JoosuaKoskinen: Somehow slipped in, this has been fixed now.

PR #238, opened 2019-05-16
Move rest of the pages to postboxes

cr0vy:
This PR will move shadows, domains, mails and logs pages to postboxes.

Domains page:
domains-no-items

domains-valid

Mails page:
mails-no_valid
mails-valid

Logs page:
logs-2

Shadows page:
shadows


l3ku:
Is this still WIP, as the commits are titled WIP? If not, could you transform the commits into logical entities so that they could be merged into master? Also, I think that the Postbox layout is not very suitable for the logs and domains page, as the content becomes very crowded, and e.g. editing the domain zone becomes harder:
Screen Shot 2019-05-17 at 16 06 19

This is just my opinion, but I think that those pages can use a non-postbox layout or then we need to figure out how to make the Postboxes full width somehow.

ottok:
I agree with Leo's remarks. Please fix.

cr0vy:
I dropped every "WIP" tags from the commits titles. I've also dropped the commits which move domains and logs pages to postboxes.

tlxo:
Is the postbox width limited for a reason? Is there a possibility that there might be also vertical overlap? I'd maybe use `overflow: auto;` just in case to cover both cases. 

Also, any chance those columns could have `min-width` set so that the relatively short words in the title wouldn't break?

l3ku:
Feel free to experiment, however we must take into consideration that our Postboxes also use WP styles by using common classes (which is intended, but need to take precaution).

modules/shadows.php
```diff
@@ -66,7 +66,7 @@ public static function register_shadows_scripts( $page ) {
         wp_localize_script( 'seravo_shadows', 'seravo_shadows_loc', $loc_translation );
       }
     }
-
```
ottok: You already have code styling fixes in this branch. Removing some whitespace does not need it's own commit. I suggest you squash this with the other code style fixes.

modules/mails.php
```diff
@@ -37,13 +37,26 @@ public static function register_mails_page() {
...
+      ?>
+      <form action="#" method="get" style="width: 100%; margin-bottom: 10px;">
+        <input type="hidden" name="page" value="<?php echo $_REQUEST['page']; ?>"/>
+        <?php list_domains(); ?>
+      </form>
+      <form>
+        <input type="hidden" name="page" value="<?php echo $_REQUEST['page']; ?>"/>
+        <?php
+        if ( ! empty ( $_GET['domain'] ) ) {
+          get_forwards_table();
```
ottok: This commit adds a call to this function but it is not defined anywhere. You should not have a commit like this, it is not atomic. This same commit should also include the definition of that function.

style/shadows.css
```diff
@@ -0,0 +1,5 @@
+@media only screen and (min-width: 1500px) {
+  #wpbody-content #dashboard-widgets .postbox-container {
+    width: 75%;
```
ottok: The screenshot of the shadows page shows that all table headers and columns are squeezed and unreadable. Are you sure slimming it down even more to 75% is the right thing to do?

lib/mails-page.php
```diff
@@ -159,38 +159,3 @@ function domain_reorder( $a, $b ) {
 $forwards_table = new Seravo_Mails_Forward_Table();
 // Fetch, prepare and sort our data...
 $forwards_table->prepare_items();
-
-?>
-
-<!-- Postbox wrapper -->
-<div id="dashboard-widgets" class="metabox-holder">
```
ottok: This commit removes a lot of code, but does not add it somewhere else. Are you sure this is correct?

In February you did this set of commits where WIP ended up on the master branch and the visual regression screenshot section disappeared completely:
image

You should practice diligence in these commits to get to the next level as a developer. :unicorn:

l3ku:
@cr0vy when you are ready please re-request a review so I can know when to skim through this again. :)

PR #239, opened 2019-05-17
Add new layout to cruft plugin (Closes: #215)

### LINKED ISSUE #215: Unify and beautify cruft files screen
Current Cruft files page looks like this:
image

There are no 3 types of tables here:
- one plain table with trash bin icons in front of the row
- one table with black row dividers and trash bin icons in front of the row
- one table with select boxes in front of the row

This hurts my eye. All tables should have a unified look.

Also the UI uses JavaScript alert boxes. There should be some more WordPressish in-line feedback boxes. Also none of the messages should say that an action is irrevocable. If a backup has been run, then all of the files removed in this screen can be recovered within 30 days.

The text on this page shoud also very clearly motivate the site admin to clean up unnecessary files on 3 grounds:
- Extra files take up extra disk space
- Extra data (sql dumps in particular) might leak user data and passwords and is a big liability (GDPR if not otherwise)
- Extra PHP files can serve as an entry point for hackers, so all files that are unnecessary should be removed

nasusa:
This pr closes #215.

I think that after this pr cruft plugin is more pleasant to use. Plugin has consistent layout, and you can select all cruft files, plugins and themes; and delete them.

_In cruft themes, if theme has a child theme it will be bolded and can't be deleted before child theme._ 

image

### DIFF
--- a/js/cruftplugins.js
+++ b/js/cruftplugins.js
@@ -1,19 +1,18 @@
 'use strict';
 
-
-
-
-jQuery(document).ready(function($) {
+jQuery(document).ready(function ($) {
   function seravo_ajax_delete_plugin(plugin_name, callback) {
     $.post(
-      seravo_cruftplugins_loc.ajaxurl,
-      { type: 'POST',
+      seravo_cruftplugins_loc.ajaxurl, {
+        type: 'POST',
         'action': 'seravo_remove_plugins',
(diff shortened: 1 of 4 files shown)


ottok:
UI looks nice!

Please rebase on master. Currenlty your commit is based on something that goes back way before current master, see green line in screenshot:
image

Issue #245, opened 2019-06-05
Minor fixes before 1.9.9 release


mvierma:
- [x] cruft remover: falsely suggests removal of parent theme
- [x] cruft remover: exclude node_modules
- [x] tests: remove old rspec test
- [x] database: add line break to search and replace box's to and from fields
![image](https://user-images.githubusercontent.com/26429376/58943541-9e246c80-8788-11e9-9044-6c77c5f10539.png)
- [x] reports: fix undefined index
![image](https://user-images.githubusercontent.com/42264063/58948281-6cb09e80-8792-11e9-95ea-6b9aa27d3385.png)
- [x] updates: fix undefined index
![image](https://user-images.githubusercontent.com/42264063/58948359-9bc71000-8792-11e9-9a06-1cf65d764e56.png)

ottok:
PHP version change does not seem to complete?
![image](https://user-images.githubusercontent.com/668724/59087705-1ff1d280-890e-11e9-90f4-15ac2f0714b8.png)

elguitar:
@ottok noticed that the Publish button is not working on domains_page. This is probably because it doesn't use the AJAX system yet.
This should be fixed before 1.9.9 too. 

@JoosuaKoskinen and I will sort this out :)

JoosuaKoskinen:
These issues has been fixed in the commits referencing this and 1.9.9 has been published.

PR #247, opened 2019-06-07
Minor fixes before 1.9.9

JoosuaKoskinen:
This fixes some small bugs and issues which were meant to be fixed before publishing Seravo Plugin version 1.9.9.

Changes summed up:
- Exclude node_modules and parent themes from cruft remover.
- Remove old rspec test.
- Improve search and replace form layout.
- Fix undefined index from couple of pages.

These all were reported in #245. Note that this PR doesn't include the bugs commented in the issue.


modules/reports.php
```diff
@@ -158,8 +150,7 @@ public static function seravo_disk_usage() {
 
     public static function seravo_site_info() {
       if ( ! Helpers::is_production() ) {
-        _e('This feature is available only on live production sites.', 'seravo');
-        return;
+        __('This feature is available only on live production sites.', 'seravo');
```
ottok: Should this line start with `return` so the value of __() is send back?

modules/reports.php
```diff
@@ -191,10 +182,13 @@ function print_item( $value, $description ) {
         }
       }
 
+      // Nested arrays need to be checked seperately
+      $country = ! empty($site_info['country']) ? $countries[ $site_info['country'] ] : '';
```
ottok: By the way, have you read about https://php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op ?
This could be written: `$country = $countries[ $site_info['country'] ] ?? '';`

elguitar: How is that handled in 5.6? Fatal errors because of seravo-plugin would be quite unpleasant.
Otherwise great tip! ?? simplifies things a lot.

ottok: Right, we need to support PHP 5.6 for some time still.

modules/reports.php
JoosuaKoskinen: This shouldn't be here, changed it back. Also added the same to HTTP request box.

PR #248, opened 2019-06-10
Fix the publish button in zone edit

elguitar:
This commit makes the publish button work with the new AJAX system.

This is a subtask of #245

Domains page

#### Manual testing steps?
- Start with a domain which has not been published
- Publish it through domains page.
- Test editting of zone (for regressions)

#### Screenshots
2019-06-10-104050_370x148_scrot


ottok:
@JoosuaKoskinen Please review

Issue #67, opened 2017-10-06
"Poistu varjosta" link is not working, if staging has different URL than production


runomarsista:
"Poistu varjosta" link is not working, if staging URL different than production. It should be redirected back to production site URL.

Now it's using ?wpp_shadow=clear to make redirection.

ottok:
Staging sites can have their own domains, but it is still in beta and there are known limitations, like this exit link issue and the limited availability of https certificates for arbitrary custom domains on staging sites (eg. https://dev123.example.com/ not working because the certificate for https://example.com/ does not cover it)..

nuuttitoivola:
Closed via https://github.com/Seravo/seravo-plugin/pull/251/commits/bd946bd83becc1369e9e5fe6b22331b585105185

PR #251, opened 2019-06-19
Improve shadow domain support

JoosuaKoskinen:

- Adds Seravo Plugin side support for shadows with domain (more in commit message)
- Makes the functionality of instance switcher more constant

Problem reported in issue #67

Reviewer should try accessing and exiting different shadows in different ways (url / instance-switcher / shadow banner). Site 'joosuakoskinen' in Sestak can be used for testing. It has 2 shadows, one with domain and one without.

### DIFF
--- a/js/instance-switcher.js
+++ b/js/instance-switcher.js
@@ -1,30 +1,82 @@
 document.addEventListener('DOMContentLoaded', function() {
-  // Select all links in shadow drop-down
-  var links = document.querySelectorAll('#wp-admin-bar-instance-switcher li > a');
+  // Select all links for entering shadow in admin bar dropdown
+  var shadow_links = jQuery('#wp-admin-bar-instance-switcher li.shadow-link > a');
+  // Select all links for exiting shadow (dropdown or banner)
+  var exit_links = jQuery('li.shadow-exit > a, a.shadow-exit');
 
-  var listener = function(e){
-    // Match all strings #something in the link targets.
-    // Typically #exit or #abc123 (shadow ID).
-    var instance = e.target.getAttribute('href').match(/#([a-z0-9]+)/)[1];
-    if (instance === 'exit') {
-      // If exit selected, clear cookies.
-      e.preventDefault();
+  // Listener for entering shadow
+  var shadow_listener = function(e) {
+    e.preventDefault();
+
+    var new_location = location.href;
+    var target = jQuery(this).attr('href');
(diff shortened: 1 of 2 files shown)


js/instance-switcher.js
```diff
@@ -1,30 +1,82 @@
...
-  var limit = links.length;
+    // Reload / redirect page
+    location.href = new_location;
+  }
+
+  // Add right listener for the links found
+
+  var limit = shadow_links.length;
+  for (var i = 0; i < limit; i++) {
+      shadow_links[i].addEventListener('click', shadow_listener);
```
ottok: Indentation 2 -> 4 ?

js/instance-switcher.js
```diff
@@ -1,30 +1,82 @@
...
+  }
+
+  limit = exit_links.length;
   for (var i = 0; i < limit; i++) {
-    links[i].addEventListener('click', listener);
+      exit_links[i].addEventListener('click', exit_listener);
```
ottok: Indentation 2 -> 4 ?

js/instance-switcher.js
JoosuaKoskinen: Fixed back to 2

js/instance-switcher.js
JoosuaKoskinen: Also fixed.

JoosuaKoskinen:
> The commit message is really good! Since the text does not describe only the change, but mostly how the code now permanently works, should some parts of the commit message maybe be as inline comments in the code?

The same fallback flow is commented in `instance-switcher.php`.

modules/instance-switcher.php
```diff
@@ -193,6 +216,43 @@ public static function render_shadow_admin_notice( $current_screen ) {
         echo $admin_notice_content;
       }
     }
+
+    public static function get_shadow_domain( $identifier ) {
+      $domain_list = API::get_site_data('/domains');
+      if ( is_wp_error( $domain_list ) ) {
+        die( $domain_list->get_error_message() );
```
elguitar: The error from the API should be handled gracefully. Dying at this point leads to wp-admin not drawing pages.

JoosuaKoskinen: Fixed, instead of `die()` this now returns null and the caller handles that properly.

elguitar:
Apart from failing CI tests (identation and whitespace), looks good to me!

Issue #252, opened 2019-06-20
Notice: Undefined index: HTTP_REFERER


TeemuSuoranta:
When /wp-admin/ is opened in new tab when logged in, the HTTP_REFERER is not set and should be checked before referenced.

Full notice:
```
Notice: Undefined index: HTTP_REFERER in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/modules/check-php-version.php on line 30
```

This can be simply checked by:
```
if (isset($_SERVER['HTTP_REFERER']) && strpos($_SERVER['HTTP_REFERER'], 'wp-login.php') === false) {
```

ottok:
What version are you running? Did you try `wp-seravo-plugin-update` ?

TeemuSuoranta:
It's the `1.9.9` which is the newest release. The buggy code is in https://github.com/Seravo/seravo-plugin/blob/master/modules/check-php-version.php#L30

TeemuSuoranta:
PHP version is 7.2 in case that handling `$_SERVER['HTTP_REFERER']` has somehow changed in `7.3`

JoosuaKoskinen:
This was fixed in https://github.com/Seravo/seravo-plugin/pull/253.

PR #253, opened 2019-06-20
Fix a couple of undefined notices

JoosuaKoskinen:
- Fixes HTTP_REFERER undefined notice reported in #252
- Removes usage of useless undefined variable on updates page

### DIFF
--- a/modules/check-php-version.php
+++ b/modules/check-php-version.php
@@ -27,7 +27,7 @@ if ( ! class_exists('CheckPHPVersion') ) {
 
       // Show only on main dashboard once directly after login so it
       // will not clutter too much.
-      if ( strpos($_SERVER['HTTP_REFERER'], 'wp-login.php') === false ) {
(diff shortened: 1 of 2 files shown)


ottok:
If you include "(Closes: #252)" in the commit title or message, Github will automatically close the issue when this is merged.

The bug was introduced by me in https://github.com/Seravo/seravo-plugin/commit/ed6bf84eb5ea505192b4205d2b0a3961437c0597

Fix looks correct. Thanks!

There is a similar line in
> modules/login-notification.php:        if ( apply_filters('seravo_dashboard_login', true) && strpos($_SERVER['HTTP_REFERER'], 'wp-login.php') !== false ) {

Should it maybe also be fixed?

JoosuaKoskinen:
> There is a similar line in
> 
> > modules/login-notification.php:        if ( apply_filters('seravo_dashboard_login', true) && strpos($_SERVER['HTTP_REFERER'], 'wp-login.php') !== false ) {
> 
> Should it maybe also be fixed?

No, that's fine. The existence of HTTP_REFERER is already being checked on the line before.

Issue #228, opened 2019-03-06
Fix incorrect usage of Seravo Postboxes


l3ku:
Example of places with incorrect usage:

## Mails
Incorrect call of `add_submenu_page()`, the function takes no 7th parameter and the 6th parameter should be `'Seravo\seravo_postboxes_page'`:
https://github.com/Seravo/seravo-plugin/blob/abf42101adbd0303b639caef3e3d5cd70ee7117e/modules/mails.php#L33-L35

Also the actual page content in `mails-page.php` should not define the postbox wrapper, only the content (div.inside) should be defined:
https://github.com/Seravo/seravo-plugin/blob/abf42101adbd0303b639caef3e3d5cd70ee7117e/lib/mails-page.php#L180-L191 are required

## Updates
Currently the updates module uses all postboxes as "side", however, "normal" should be primarily used. "side" is used only to force postboxes to the side, see https://codex.wordpress.org/Dashboard_Widgets_API#Advanced:_Adding_Widgets_onto_the_side for a description.

See e.g. `/modules/backups.php` for a reference of correct usage. Also, we have now added relevant docs to `/docs/README.md`. In case there is still some crucial information missing from the docs, here is a good opportunity to identify any missing information and expand the docs.

ottok:
I assume this is a symptom of what Leo reported above:
![image](https://user-images.githubusercontent.com/668724/57223232-0d823180-700e-11e9-9d34-c2292e46d054.png)

elguitar:
@cr0vy confirmed that the Updates page that @ottok mentioned is not an issue anymore. However, this #228 is still valid.

ottok:
Closed via https://github.com/Seravo/seravo-plugin/commit/f1f50b7e76b2fa3fc56861380daa3c05a0630ec8

Issue #73, opened 2017-10-25
Notify customer on wp-admin if version updates has failed


runomarsista:

simoke:
So it's propably necessary to extend the plugin to display notifications on WP admin front page (in case that timestamp of the latest successful update is prior than the latest update attempt, or null), or alternatively improve update automation to send email notifications to WP administrative users.

![updates](https://user-images.githubusercontent.com/25475688/31997999-c5799998-b996-11e7-8697-af178583e073.png)

elguitar:
b0b6fbc2e3e711dc0d86ebd9bf229f3654cf529a has some relevant code for showing information in wp-admin's dashboard. For now it shows the last login and the amount of php-errors.

l3ku:
Sending automatic emails regarding failed updates and site errors is currently in progress. I personally don't think it is good practice to flood wp-admin with notifications, so we should really consider if it is necessary to implement this notification in addition to sending emails to customers.

I made a very simple UI into the Seravo Plugin "Updates" page ([WIP](https://github.com/Seravo/seravo-plugin/tree/feature-notification-email-field)) for the customer to enter his/her specified technical contact emails:

![screen shot 2017-12-09 at 16 13 41](https://user-images.githubusercontent.com/24781811/33796425-9392f6b6-dcfc-11e7-8586-b2bd77e61fef.png)

Any opinions on the screenshot above, like is the "Updates" page the correct location for this and is the UI simple enough?

ottok:
In the future, we might have a separate Notifications page, with a fancy add/remove UI fo rmultiple e-mail addresses (or a selection from WP user list to be even fancier), and options for customers to select what notifications they want, and set a limit on how much or how frequently they want to have e-mail, log about recently sent notifications etc.

**But for now, the screenshot above is OK and enough.** Better to have something that works end-to-end with first, and make the fancy things at a later time.

annijpesonen:
The way I understand it, the most important thing would be to send a notification to customers only when there is some permanent issue on their website that makes updating impossible and requires their actions. I don't know if this kind of separation of notifications is possible, because then our system should also be able to automatically detect what the problem is. Or maybe the notification should be sent to us? Like this: "there seems to be a permanent issue on updating this website --> look into it and notify the customer". As far as I know, at the moment we only look for these problems if some customer complains about their site not being updated in a long time.

ottok:
[...]

PR #256, opened 2019-06-24
Add fixes and improvements to seravo-plugin Updates, Reports and Logs pages

mvierma:
Add more information under Site Information on Reports page per issue #318 in another repo.

Site Info now shows the current Account Manager of the site if one is set and displays a message about account manager being a part of the Seravo Enterprise plan otherwise. Site info now also shows the emails of technical contacts of the site and links to the Updates page where user can add or remove technical contacts from their site.

Site Information postbox with new content:
image

This PR also adds the notification on failed full updates of the site on the Updates page if the full update hasn't been succesful in 30 days or over, as suggested in the issue #73. The notification shows lines from the latest update.log starting with "Updates failed!" to the user and prompts them to have a site developer take a look at the issue preventing the full update. The notification also links to the update log on the Logs page.

Preview of the notification:
image

Logs page now checks which .log files were found and if any are missing, the latest timestamped log file is fetched instead. This ensures that the page always has a logfile to show for every log type.

Example of the view with update.log file missing:
image

Lastly this also fixes the incorrect add_submenu_page calls on all modules and fixes the postbox placements where applicable per issue #228


ottok:
Well done! 

Quickly skimmed through code.
- in commit 2c6039f120a41ad1e5321dff6d93f5b9ec57925e you add function `print_items()` - maybe you could extend the existing `print_item()` instead to keep the code simpler?
- in commit ccab71ca48353063c8352c0ab472270d09ad83d4 you change the previous commit - it is easier to review new code if you just squash all of the same commits to one step compared to master. Or is there something particular in the intermediate results that you want me to review instead of going directly for the end result this PR produces?
- in commit 91a9a2d4af72744883cab325acdf20611cbd3534 you add long A tags into the string that is to be translated, but there is really nothing to translate in the link syntax. Instead what link A stuff could be in the parameter and the string just keep the `see %2$s and %3$s.` as it was

ottok:
Please squash and rebase on master. Consult @l3ku on PR principles.

mvierma:
Commits should now be better organized and squashed so that each commit is it's own addition or fix.

ottok:
End result looks good!

At least the first and last commit should be squashed. A pull request should  be a suggestion on how to change the master branch for better – it should not have commits that fix earlier commits in the same pull request.

Consult @l3ku on PR principles.

l3ku:
We already discussed about how the PR should not have commits that alter something that was added in the same PR, but rather "get it right" in the first commit so that it is not necessary to alter the content in a later commit.

Issue #6, opened 2015-09-15
Collect and show login log


ottok:
All sites should collect a /data/log/login.log file. Consider adopting functionality from https://github.com/nrosvall/wp-login-log

It would also be good to have a menu item where under users can directly from WP-admin view a list of last logins. It could simply just be a login.log parser and prettifier.

ottok:
We now collect wp-login.log for all logins, and we tell users in the dashboard when they last logged in.

One last thing that would be super useful, would be to show on the profile.php page a section that lists the last 10 logins for that user, so that users themselves can view their own history, and that site admins can view the history of whichever user they want.

ottok:
See also discussion in #7

l3ku:
Before we start implementing this, we should reconsider the source for the user login data. As the `wp-login.log` file is regularily rotated, there may be a scenario that the 10 last logins are not found even though the user has most certainly logged in more than 10 times. To me, it makes no sense to start decompressing the previous gzipped logfiles in order to fetch the previous logins.

ottok:
Design outline:
- Extend page `/wp-admin/tools.php?page=security_page` to have a new postbox titled "Recent successful logins"
- Show list or table (depends on Seravo postbox width vs contents width)
- View last 10 logins - or last logins for past 10 days?
- Use latest wp-login.log + logrotated prevous files (/data/log/wp-login.log, -dateext, -dateext.gz)
- Show login time, IP address, username
- Placed in a postbox
- function parameters: username, min and max number of logins wanted, geolocation on/off (so that same function can be used in security screen)
- open logs and parse in php
- Security page is visible only to admin. Thus show all logins. In a later revision consider making some extension to profile.php so that users can see their own logins.
- Use ajax to populate postbox so users don't need to "wait" for the log processing. See reports page for example.

PR #257, opened 2019-06-26
Add postbox which shows 10 latest successful logins to security page

eerokokko:
This pull request adds postbox to security.php page and shows 10 latest successful logins from the wp-login.log and if wp-login.log doesn't have enough entries it will get the rest from the rotated log files. Ajax spinner populates the postbox until the login information is gathered.

Function has now only parameter for maximum results showed. Username seemed unnecessary as this is visible only to admin and for admin we wanted to show all the logins. Minimum parameter is also left out because opening rotated files is fast and I didn't see a real need to have a minimum to give up opening rotated files. Implementing geolocation will require installing PHP extension so that will need to be added later.

Relates to issue: https://github.com/Seravo/seravo-plugin/issues/6

#### Screenshots

image

### DIFF
--- a/lib/security-ajax.php
+++ b/lib/security-ajax.php
@@ -1,0 +1,102 @@
+<?php
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+// Returns IP, username and time of successful logins.
+// Number of results are limited by $max.
+function seravo_logins_info( $max = 10 ) {
+  // Get the latest logins from wp-loging.log
+  $login_data = file('/data/log/wp-login.log');
+
+  // If the wp-login.log has less than $max entries check older log files
+  if ( count(preg_grep('/SUCCESS/', $login_data)) < $max ) {
+    // Check the second newest log file (not gzipped yet)
+    $login_data2_filename = glob('/data/log/wp-login.log-[0-9]*[?!\.gz]');
+    // There should be only one file matching previous criterion, but
+    // count the files just in case and choose the biggest index
+    $login_data2_count = count($login_data2_filename) - 1;
+    // Merge with the first log file
+    $login_data = array_merge(file($login_data2_filename[ $login_data2_count ]), $login_data);
+
+    // Opening necessary amount of gzipped log files
+    // Find the gzip log files
+    $login_data_gz_filename = glob('/data/log/wp-login.log-[0-9]*.gz');
+    // Get the number of gzip log files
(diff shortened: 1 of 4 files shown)


ottok:
Nice! Code looks clean, and it is freshly rebased on master. Should you maybe squash all 4 commits into one big one, since they togheter add this one feature and the intermediate steps are not relevant for the final history? Also, that commit could end in "(Closes: #6)" so the issue auto-closes?

About the UI. Should the simple text list instead maybe be a HTML table so the columns always align and maybe is a bit easier to scan with the eye?

ottok:
Does @mvierma have something to comment here?

mvierma:
We talked about the HTML table formating with Eero and one thing to consider is the case of long usernames. I suggest that if a username exceeds n length it should be cut off to preserve table formating. There could be a .css tooltip on mouse hover for example to show the full username in such cases. 

Other thing to consider is adding the parameters suggested in the issue #6 even if they're not utilised yet. That way the overall functionality of the postbox would be easier to adapt elsewhere in the future, but I'm not sure if it's really important at this time as it is an easy addition to implement when needed as well.

Eero's regex approach to rotated logs seems simpler than my approach so I think we should have a look at it with Eero and see if I should use the same approach in my code as well.

style/security.css
```diff
@@ -0,0 +1,18 @@
+.login_info_table {
+  table-layout: fixed;
+  width: 100%;
+  #white-space: nowrap;
```
ottok: # does not work in CSS

lib/security-ajax.php
```diff
@@ -0,0 +1,97 @@
...
+
+    } else {
+      // Remove failed logins
+      unset($login_data[ $i ]);
+    }
+  }
+  // Re-index the array after unsetting failed logins
+  $login_data = array_values($login_data);
+  // Adding column titles
+  $column_titles = array('<table class="login_info_table"> <tr> <th class="login_info_th">IP address</th> <th class="login_info_th">User</th> <th class="login_info_th">Date</th> <th class="login_info_th">Time (UTC)</th> </tr>');
```
ottok: Translations are missing.

ottok:
Table column headers are not translatable at the moment. See example
image
and read up on https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/

ottok:
Perfect! :unicorn:

ottok:
This isn't rebased on master so merging via rebase.

PR #261, opened 2019-07-05
Add ellipsis also for IP address

eerokokko:
This will add CSS ellipsis also for IP address. Currently if postbox happens to be really narrow and the IP is near the maximum length of 15 characters the last digits of the IP are written over the username. This bugfix will shorten the IP and add it in a tooltip if the IP doesn't fit otherwise.

#### Screenshots

Currently:
Screenshot at 2019-07-04 16-33-46

With IP ellipsis:
Narrow:
Screenshot at 2019-07-05 11-03-03

Wide:
Screenshot at 2019-07-05 11-06-11

### DIFF
--- a/lib/security-ajax.php
+++ b/lib/security-ajax.php
@@ -46,6 +46,9 @@ function seravo_logins_info( $max = 10 ) {
       unset($login_data[ $i ]);
     } else if ( strpos($login_data[ $i ], 'SUCCESS' ) ) {
 
+      // Get IP. IP address is in the beginning of log line and ends to " -"
+      $ip = substr($login_data[ $i ], 0, strpos($login_data[ $i ], ' -'));
+
       // Get the username. Username in log files between first "-" and "["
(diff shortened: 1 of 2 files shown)


lib/security-ajax.php
```diff
@@ -78,7 +82,7 @@ function seravo_logins_info( $max = 10 ) {
     '<th class="login_info_th">' . __( 'IP address', 'seravo' ) . '</th>' .
     '<th class="login_info_th">' . __( 'User', 'seravo' ) . '</th>' .
     '<th class="login_info_th">' . __( 'Date', 'seravo' ) . '</th>' .
-    '<th class="login_info_th">' . __( 'Time', 'seravo' ) . ' (UTC)</th></tr>';
+    '<th class="login_info_th">' . __( 'Time UTC', 'seravo' ) . '</th></tr>';
```
elguitar: This will probably break the translation.

eerokokko: Thanks for noticing! I removed that change.

PR #263, opened 2019-07-08
Fix #262 (Update summary blank on some sites, should never be)

### LINKED ISSUE #262: Update summary blank on some sites, should never be
**Describe the bug**

I noticed on a site this UI:
image

The section under *Latest update.log:* is empty. It should always have at least one line to make sense?

Note also the untranslated texts..

mvierma:
Fixes the incorrectly displayed empty update.log contents by checking that there is content to display. Also adds missing translations to Updates page update notification and links.
Issue: #262 
2b87e5d fixes the empty display and should be checked that it now functions correctly. If update.log has no content to show but the full update has not been succesful, only the notification should now display. If the update.log has lines starting with "Updates failed!" those lines are shown.
1b761ef adds translations

### DIFF
--- a/modules/updates.php
+++ b/modules/updates.php
@@ -223,15 +223,16 @@ if ( ! class_exists('Updates') ) {
               $update_log_output = implode('<br>', $update_log_contents);
             }
 
-            echo '<p>' .
-                __(
-                  'Last succesful full site update was over a month ago. A developer should take
-                a look at the update log and fix the issue preventing the site from updating.',
-                  'seravo'
-                ) .
-                '</p><p><h3>' . __('Latest update.log:', 'seravo') . '</h3></p><p>' .
-                $update_log_output . '</p>' .
-                '<p><a href="tools.php?page=logs_page&logfile=update.log&max_num_of_rows=50">See the logs page for more info.</a></p>';
+            echo '<p>' . __(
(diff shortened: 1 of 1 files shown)


ottok:
This is not rebased on master. It looks like you did some extra work in vain, e.g updated translations which you would not have had to.

Since each commit already has titles and such, writing this down seems like a waste of time?
> 2b87e5d fixes the empty display and should be checked that it now functions correctly. If update.log has no content to show but the full update has not been succesful, only the notification should now display. If the update.log has lines starting with "Updates failed!" those lines are shown.
> 1b761ef adds translations

And the commit shasums will not be correct anyway after you have rebased the PR.

ottok:
The string `See the logs page for more info.` is just echoed and not gettexted, so it cannot be translated.

I already fixed similar things in https://github.com/Seravo/seravo-plugin/commit/6762ae7ed098ffb6ae3fc4523d0d340fbf5326b2#diff-493649143a31999adbeb8a6b8aefddcaL215 and with the example you can do it easily yourself.

mvierma:
Added the missing gettext to the string `See the logs page for more info.` and rebased to master.

ottok:
Please append to the commit message "(Closes: #262)" since that is what this is doing, right?

Issue #243, opened 2019-05-24
Clarify tests' message (tests via control panel)


Mardaukar:
The message after performing tests via control panel are misleading to the user. Even if the tests report failures, the message reads: "Testit suoritettu onnistuneesti ilman virheitä!"

Go to: WordPress control panel -> Työkalut -> Tests.
Press: "Suorita testit", under "Testit (uusi Codeception)".
After tests should read: "Testit suoritettu onnistuneesti ilman virheitä!"

![Screenshot from 2019-05-24 09-51-01](https://user-images.githubusercontent.com/50039402/58313352-06349380-7e16-11e9-85ca-f31249340ba1.png)

![Screenshot from 2019-05-24 09-49-52](https://user-images.githubusercontent.com/50039402/58313358-09c81a80-7e16-11e9-8aea-677e1dcf6758.png)

Additional context: Customer gave feedback on this.

l3ku:
The AJAX functionality in https://github.com/Seravo/seravo-plugin/blob/master/lib/tests-ajax.php should in addition to the output probably also return the command exit code, and fail if it is nonzero. Currently the mechanism checks for some string present in the output at https://github.com/Seravo/seravo-plugin/blob/2d7649bb16de2ba6b61d64d69a97ed70ba8e0316/js/tests.js#L91, which is certainly not a good design.

ottok:
Labelled this as priority since the feedback to the user is simply "Testit suoritettu onnistuneesti ilman virheitä!" by default (before they open the box with more info) so this is quite a grave functionality problem.

mvierma:
Changing the checked string from https://github.com/Seravo/seravo-plugin/blob/2d7649bb16de2ba6b61d64d69a97ed70ba8e0316/js/tests.js#L91 to https://github.com/Seravo/seravo-plugin/blob/9ade94ac6e84638590a7eee7a304b3fb987eba37/js/tests.js#L17 fixes the display so it now shows if there are any failures in the tests. 
![image](https://user-images.githubusercontent.com/26429376/60811050-25ee0400-a197-11e9-9ad4-b4981be73b93.png)

This could be used as a quick fix before a more thorough update to the program logic.

l3ku:
Can we assume that the output always contains the string "Failures"? What if "Failures" is somewhere else in the logging output of the customer's custom tests? 

[...]

PR #264, opened 2019-07-11
Fix display of wp-test results on Tests page

mvierma:
Fixes the checking of failed tests by changing the checking of "Failed examples" string to check the return value of wp-test instead. Update Tests postbox now displays failed tests correctly if the return value of wp-test is non-zero.
Issue #243 

#### Screenshots
image

### DIFF
--- a/lib/tests-ajax.php
+++ b/lib/tests-ajax.php
@@ -7,8 +7,12 @@ if ( ! defined('ABSPATH') ) {
 }
 
 function seravo_tests() {
-  exec('wp-test', $output);
-  return $output;
+  exec('wp-test', $output, $return_variable);
+  $return_arr = array(
+    'test_result' => $output,
+    'exit_code' => $return_variable,
+  );
+  return $return_arr;
 }
 
 function seravo_ajax_tests() {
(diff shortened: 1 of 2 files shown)


lib/tests-ajax.php
```diff
@@ -7,8 +7,12 @@
 }
 
 function seravo_tests() {
-  exec('wp-test', $output);
-  return $output;
+  exec('wp-test', $output, $return_variable);
+  $return_arr = array(
+    0 => $output,
```
ottok: Don't use variables named 0 and 1. Use something descriptive like 'output' and 'exit_code' maybe?

ottok:
Looks good!

Just a few very small details to fix and it's done. Note that usually the "(Closes:#243)" is in the title and there is a space between : and #.

lib/tests-ajax.php
mvierma: I removed the unnecessary variables from the array entirely, it now simply indexes both `$output` and `$return_variable` as that was my original intent with the 0 and 1 approach.

ottok:
Please consult @l3ku on good naming practices. Code is hard to read when there are checks like `data[1] === 0` or variables called "return_variable" or "return_arr". The names of things should describe their contents so humans can read `if exit code is zero` or `exit code` and `command output`.

The payload sent as JSON should also follow good practices and describe itself. If somebody is debugging this code and sees in the browser console just "0, ''" it can be hard to understand, while something like this would make much more sense
```
'exit_code' => 0,
'test_result' => ''
```

PR #266, opened 2019-07-16
Fix alloptions not updating in cache

JoosuaKoskinen:

Forces updating of 'alloptions' in cache. Some WordPress-hosts 
are running similar fixes. Issue explained in the [commit message](https://github.com/Seravo/seravo-plugin/commit/95f03d047917cb7e88e51140182a6c7752b0ef48). 

The issue has caused problems on some customers but
affects everyone.

#### Manual testing steps?

This has already been tested on some sites and probably 
doesn't need more testing. The performance impact has also been
tested with `wp-speed-test` and no clear difference was noticed.

If the issue needs to be reproduced, it should be possible like this:
1. Run `while sleep 0.1; do wp option set debug-option $(openssl rand -hex 1); done`
2. In other shell instance, update some option key a until it stops working (refuses to change):

image

Previous shouldn't be possible with the fix applied.

### DIFF
--- a/modules/fixes.php
+++ b/modules/fixes.php
@@ -69,15 +69,13 @@ if ( ! class_exists('Fixes') ) {
      */
     public static function maybe_clear_alloptions_cache( $option ) {
 
-      // error_log("added/updated/deleted option: $option");
-
       if ( wp_installing() === false ) {
         $alloptions = wp_load_alloptions(); // alloptions should be cached at this point
 
-        // If option is part of the alloptions collection then clear it.
+        // If alloptions collection has $option key, clear the collection from cache
+        // because it can't be trusted to be correct after modifications in options.
         if ( array_key_exists($option, $alloptions) ) {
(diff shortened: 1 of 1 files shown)


ottok:
Your commit message is great! You are however not adding any comments to the code at all. Most of the commit contents should maybe be as inline comment explaining why the code does what is does, or what do you think?

JoosuaKoskinen:
> Your commit message is great! You are however not adding any comments to the code at all. Most of the commit contents should maybe be as inline comment explaining why the code does what is does, or what do you think?

I updated the comments to better fit the modified fix. There's still the links to WordPress ticket and Redis issue if more information is needed.

ottok:
Did you also compare to what WP VIP is doing?
image

The inline comments should include a link to https://core.trac.wordpress.org/ticket/31245 as that issue explains very well what we try to solve (and eventually when it was solved upstream and we can remove our fix).

PR #267, opened 2019-07-18
Add Seravo plugin updater to the Updates subpage

cr0vy:
This change will add Seravo Plugin Updater.

A pictures of the change
New update available:
seravo-updater-new-version-available

Latest version:
plugin-updater-latest-version

### DIFF
--- a/lib/updates-ajax.php
+++ b/lib/updates-ajax.php
@@ -38,6 +38,30 @@ function seravo_php_check_version() {
   }
 }
 
+function seravo_plugin_version_check() {
+  $current_version = Seravo\Helpers::seravo_plugin_version();
+
+  if ( $current_version == seravo_plugin_upstream_version() ) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+function seravo_plugin_upstream_version() {
+  $upstream_version = get_transient('seravo_plugin_upstream_version');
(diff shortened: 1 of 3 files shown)


lib/updates-ajax.php
```diff
@@ -38,6 +38,30 @@ function seravo_php_check_version() {
...
+    return true;
+  } else {
+    return false;
+  }
+}
+
+function seravo_plugin_upstream_version() {
+  $upstream_version = get_transient('seravo_plugin_upstream_version');
+  if ( false === $upstream_version ) {
+    $upstream_version = exec('curl -s https://api.github.com/repos/seravo/seravo-plugin/tags | python3 -c "import sys, json; print(json.load(sys.stdin)[0][\'tarball_url\'])" | rev | cut -d / -f -1 | rev');
```
JoosuaKoskinen: Is there a reason to cut the version from 'tarball_url' instead of just using 'name'?

They seem to be the same: https://api.github.com/repos/seravo/seravo-plugin/tags

cr0vy: Nope. This bash command was from the wp-seravo-plugin-update script.

JoosuaKoskinen: Seems good now, grep is probably lighter than parsing the JSON.

PR #268, opened 2019-07-19
Redesign shadows postbox

JoosuaKoskinen:

This redesigns shadows page to be more like mails-page. The postbox width is no longer
forced to 100% to support multiple postboxes on the same page. Now that shadows work
better with custom domains, I also added 'domain' field and search-replace instructions.

These changes were supposed to come with wp-shadow-pull feature but I made this now
separately as there's some problems with it and I don't want to slow down toolbox update development. 

This was planned in wp-shadow-pull issue comments.

#### Screenshots

image

image

### DIFF
--- a/js/shadows.js
+++ b/js/shadows.js
@@ -2,21 +2,64 @@
 'use strict';
 
 jQuery(document).ready(function($) {
-  $('.shadow-reset').click(function(event) {
+  $('#shadow-selector').change(function() {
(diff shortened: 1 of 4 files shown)


elguitar:
I will review this now because this stalls development in https://github.com/Seravo/seravo-plugin/tree/feature/seravo-toolbox.

PR #269, opened 2019-07-23
Add config files checker to PHP version switcher

cr0vy:
PHP Version Switcher:
1. Check current PHP file.
2. Disable PHP Switcher if the current config file is not custom.conf, examples.conf or php.conf.
3. If php.conf exists, comment out "set $mode php" in custom.conf and exampes.conf if these exists.

Number 2 because if the developer sets "set $mode php7.3" to xyz-version.conf, the PHP Version Switcher cannot change the PHP version afterward. 

Number 3 because wp-restart-nginx could look like:
```testing nginx configuration...
nginx:` the configuration file /etc/nginx/nginx.conf syntax is ok
nginx: configuration file /etc/nginx/nginx.conf test is successful
Checking for PHP backends...
Found PHP mode definition: php7.2 in '/data/wordpress/nginx/custom.conf'
Found PHP mode definition: php7.3 in '/data/wordpress/nginx/examples.conf'
Found PHP mode definition: php7.3 in '/data/wordpress/nginx/php.conf'
WARNING: Found multiple PHP mode definitions (see above), using last one (php7.3)
PHP7.3 mode enabled
Restarting nginx...
Nginx restarted!
```

#### Manual testing steps?
1 & 2.
	Set "set $mode php7.3;" to /data/wordpress/nginx/xyz_test.php
	Reload updates page

3.
	Check that /data/wordpress/nginx does not contain any other files than: custom.conf, examples.conf and php.conf
	Check that php.conf and custom.conf or examples.conf contain uncommented "set $mode phpX.X;"
	Reload updates page
	Then check custom.conf or examples.conf. Now this file should contain "#set $mode phpX.X;"

#### Screenshots
change_php_version_2

change-php-version


ottok:
We don't require the users to use filenames custom.conf, examples.conf or php.conf. The config checker should be agnostic about the filenames. Simply automatically comment out *any* existing `set mode` lines and insert the new php.conf with the one new `set mode` line?

If we modify files that customer has in version control, a `git add nginx/*.conf; s-git-commit -m "Set new PHP version"` should be run.

ottok:
?

cr0vy:
I don't like that the config file checker could comment out `set $mode phpX.x` on client made config.conf without a permission. I'm thinking that we should add a checkbox, which ask the permission to move `set $mode phpX.x` from current php config file to php.conf.

cr0vy:
Now after latest commits
When current config file isn't custom.conf, examples.conf or php.conf:
config-file-checker

config-file-checker2

When current config file is php.conf:
config-file-checker3

Also s-git-commit command is now added.

ottok:
Did some testing now, basic functionality seems to work good!

However, when it comments out existing 'set mode' changes it ends up commenting out the entire file, not just the lines it is supposed to touch.

Also, when it makes the git commit for old files it should probably also make a git commit for the new php.conf file. That would probably be logical?


Noticed this minor related typo:
image

cr0vy:
Latest commits fix these problems.

Now if /data/wordpress contains .git folder, PHP change commit is done every time when client uses the PHP switcher to change PHP version.

PR #272, opened 2019-08-07
Sort the last 5 partial or attempted updates

cr0vy:


elguitar:
Nothing too complicated. It was clear that the sort was needed here. Good job! :) :1st_place_medal:

Issue #233, opened 2019-04-25
Seravo Plugin Page


cr0vy:
Template of the Seravo Plugin page found now from dev/introduce-control-panel branch.

**Current states:**
- Includes land page and a few example submenus
- Include dashboard widget, which use own class / code file

**Needs to brainstorming:**
- Button position in left side panel
    - After the "Settings" menu?
    - Before the "Tools" menu?
- Name
    - Control Center? 
        - I use currently this in dev/introduce-control-panel branch
    - sPanel or Seravo Panel?
        - Some resellers need the whitelabel.
    - Others?
- Style
- What content will be in dashboard widget?
    - Welcome text, which replace WordPress default "Welcome" box?
    - Seravo Blog content?

**Blockers:**
- Page which still use older postbox:
    - Mails
    - Shadows
- Using non-postbox page:
    - Domains

Dashboard Widget:
![Screenshot from 2019-04-25 13-41-19](https://user-images.githubusercontent.com/43742796/56730727-49410f80-6761-11e9-82b3-9f307eee7912.png)

Control Center:
![Screenshot from 2019-04-25 13-41-32](https://user-images.githubusercontent.com/43742796/56730786-71c90980-6761-11e9-9b98-4d534e44b494.png)

ottok:
This needs to be fixed first I believe: https://github.com/Seravo/seravo-plugin/issues/228

ottok:
Name suggestions for custom top-level menu item:
- Seravo Tools
- Seravo Toolbox
- Seravo Control Panel
- ..or just Seravo?

Menu idea:
- Seravo Toolbox
  - Site status
  - Upkeep
  - Security
  - Domains
  - Email
  - Development

cr0vy:
I made concept pictures, which shows postboxes with a new placement. At the same time I used redesign picture of the shadows postbox, which find already from another issue.

Admin menu:
![admin_menu_tools](https://user-images.githubusercontent.com/43742796/60497606-d8b4f280-9cbd-11e9-958d-349c86084976.png)    ![admin_menu](https://user-images.githubusercontent.com/43742796/60497604-d81c5c00-9cbd-11e9-98d9-0d8b34c8f8c6.png)

Site status:
![sitestatus](https://user-images.githubusercontent.com/43742796/60500602-4ca5c980-9cc3-11e9-98c4-d2a5d3274b19.png)

Upkeep:
[...]

PR #273, opened 2019-08-09
Add the first part of the Seravo Toolbox

cr0vy:
This PR will add the first part of the Seravo toolbox.
This first part contains:
- A subpages under "Tools" top level menu 
- Reorganize postboxes to under these the new subpages.

Issue #233

#### Screenshots
tools-menu

toolbox-sitestatus

toolbox-upkeep

toolbox-logs

toolbox-security

toolbox-domains

toolbox-mails

toolbox-development


ottok:
Nice!

Please review each box before and after the refactoring so that you are sure that you have not accidentally removed/lost some box in the process.

ottok:
These elements are very narrow and thus difficult to read:
image

Could they be made the width of 2 columns maybe?

This view on the other hand is far to wide:
image



Navigation before:
Screenshot from 2019-08-24 19-24-43


Navigation after:
Screenshot from 2019-08-24 19-20-08

cr0vy:
> These elements are very narrow and thus difficult to read:
> image
> 
> Could they be made the width of 2 columns maybe?

Currently postboxes do not support multi columns widget, only single column widget. If we try to set the widget for 50% width of the window in css file, it looks like multi columns view, but it resizes only column 1 width to 50% and doesn't use column 2. This also moves column 4 to the second row.


> This view on the other hand is far to wide:
> image

This looks like the cache includes older shadows.css file and sets column width to 100%. This affects other columns' width and the postboxes use column width.

ottok:
Can the order of boxes be controlled? On my test installation it is a bit odd that two backup related boxes are to the left, then there is update suff in the middle, and again one backup box to the right.
image

Yes, I know I can manually order them around but I don't think many users will do that, and instead the should be set sensible out-of-the-box. Is that possible?

cr0vy:
Yes, it's possible. Function seravo_add_postbox includes $context parameter, which takes values: 'normal', 'side', 'column3', and 'column4'. If the value is 'normal', postbox default location is column-1. If the value is 'side', default location is column-2.

Currently, if you want to reset the postbox order in the site, you need to go to wp db cli from wp-cli.

In `wp db cli`, you can clear the postbox order on every page with SQL command:
`DELETE FROM wp_usermeta WHERE meta_key LIKE 'seravo-postbox-order%';`

The next SQL command clears the postbox order on upkeep_page:
`DELETE FROM wp_usermeta WHERE meta_key = 'seravo-postbox-order_tools_page_upkeep_page';`

cr0vy:
Small update. Currently this PR waiting that we could make a wider postboxes. You will find conservation about wider postboxes from the issue #233. Also, there is two PR which will gonna effect for this pr, full-width postboxes (PR #280) and two column postboxes page (PR #282).

Also, I'm working with wider update for toolbox on dev-seravo-toolbox branch, which is going to replace this current PR changes. This change set include:
- Mails and Domains on the same page
- Tests and updates php files merged into upkeep.php
- Reports, shadows, and optimize-images php files merged into sitestatus.php
- Cruftfiles.php moved to security.php, but this not affet cruftfiles .css, .ajax and .js files.

Pictures from dev-seravo-toolbox with two column postboxes:

Site Status:
site-status


Upkeep:
upkeep


Database:
database


Backups:
backups


Security:
security


Domains:
domains


Logs:
logs

cr0vy:
This latest bunch of commits contain the next things:
    - Mails and domains page merge
    - Updates, tests and shadows files merged into upkeep files
    - Optimize-images, reports and shadows files merged into site status files

cr0vy:
The "Yhteystiedot teknisissä asioissa" link is now linked to the site status page and I removed the "test" link because the "Update Tests" postbox is now on the same page.

It's maybe a good idea to add the "Development" page, which would include the shadows and optimize-images postboxes.

l3ku:
Note that prior to making the next release, the Finnish translations should be skimmed through to ensure that there are none missing, e.g.
Screen Shot 2019-10-08 at 9 58 09
It is however out of the scope of this PR because there seem to be missing translations also from code that has not been modified here.

PR #275, opened 2019-08-16
Load assets only if user is logged in or in staging.

### LINKED ISSUE #274: instance-switcher.js and .css is included for not logged-in users
instance-switcher.js is included in the front-end of production sites, even for users that are not logged in to WordPress. Instance switcher uses jQuery, so that is also included along with instance-switcher.js.

Could you change this so instance-switcher.js would only be included for users that are logged in to WordPress? This currently has negative effect to the page speed. Cheers!

samikeijonen:
Fixes #274.

Load switcher assets only if user is logged in or in staging.

## Why are we doing this? Any context or related work?
`instance-switcher.js ` is loaded in front-end even when user is logged out. And it even has dependency of jQuery. This should not happen when user is logged out (regular site user).

### DIFF
--- a/modules/instance-switcher.php
+++ b/modules/instance-switcher.php
@@ -64,8 +64,10 @@ if ( ! class_exists('InstanceSwitcher') ) {
     * Load JavaScript and stylesheets for the switcher and the banner
     */
     public static function assets() {
-      wp_enqueue_script('seravo', plugins_url('../js/instance-switcher.js', __FILE__), array( 'jquery' ), Helpers::seravo_plugin_version(), false);
-      wp_enqueue_style('seravo', plugins_url('../style/instance-switcher.css', __FILE__), null, Helpers::seravo_plugin_version(), 'all');
+      if ( is_user_logged_in() || Helpers::is_staging() ) {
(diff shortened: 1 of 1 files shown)


modules/instance-switcher.php
```diff
@@ -64,8 +64,10 @@ public static function custom_capability() {
     * Load JavaScript and stylesheets for the switcher and the banner
     */
     public static function assets() {
-      wp_enqueue_script('seravo', plugins_url('../js/instance-switcher.js', __FILE__), array( 'jquery' ), Helpers::seravo_plugin_version(), false);
-      wp_enqueue_style('seravo', plugins_url('../style/instance-switcher.css', __FILE__), null, Helpers::seravo_plugin_version(), 'all');
+      if ( is_user_logged_in() ) {
```
JoosuaKoskinen: Exit banner in staging needs the files even when not logged in (see the issue). Maybe `Helpers::is_staging()` could be added here?

samikeijonen: Added in 032ff58.

JoosuaKoskinen:
Works fine now. Could you please squash the two commits into one? Then we're ready to merge.

samikeijonen:
@JoosuaKoskinen Done.

Issue #271, opened 2019-07-24
Add GUI for PHP compatibility check to change PHP version postbox


mvierma:
Currently the link to compatibility scan results points to an empty log if the `wp-php-compatibility-check` hasn't been run on the site.

![image](https://user-images.githubusercontent.com/26429376/61791103-b4ae8200-ae21-11e9-99d3-434b98a9f774.png)

![image](https://user-images.githubusercontent.com/26429376/61791117-c09a4400-ae21-11e9-9437-8bc23e8aa417.png)

There could be a GUI option to run the compatibility check in the postbox, so user could easily check the compatibility of their site and the latest PHP version.

GUI itself could just be a button that says "Check compatibility now" or something similar with a note to the user that the check can take a while to complete if their site has a lot of plugins and after the check is complete, a link to the logs could be shown.

PR #276, opened 2019-08-20
Add GUI for PHP compatibility check

mvierma:

Adds a button for running the wp-php-compatibility-check to the Updates page
of seravo-plugin. Once the check is done running, the number of errors found
is displayed to the user. If the check exits with a non-zero exit code, the
button is re-enabled and an error is displayed.

#271 

#### Screenshots
image

image

image

image

### DIFF
--- a/js/updates.js
+++ b/js/updates.js
@@ -133,6 +133,53 @@ jQuery(document).ready(function($) {
     $buttonsDiv.html(generateButtons(emails));
   });
 
+  jQuery('#check-php-compatibility-button').click(function() {
+    jQuery(this).fadeOut(400, function(){
+      jQuery(this).hide();
+    });
+    jQuery("#check-php-compatibility-status").fadeOut(400, function() {
+      jQuery(this).html('<img src="/wp-admin/images/spinner.gif" style="display:inline-block"> ' + seravo_updates_loc.compatibility_check_running).fadeIn(400);
+    });
+    checkPHPCompatibility();
+  });
+
+  function checkPHPCompatibility() {
+    jQuery.post(
+      seravo_updates_loc.ajaxurl, {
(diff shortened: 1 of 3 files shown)


js/updates.js
```diff
@@ -133,6 +133,47 @@ jQuery(document).ready(function($) {
...
+        if ( data['exit_code'] != 0 ) {
+          jQuery("#check-php-compatibility-button").fadeOut(400, function() {
+            jQuery(this).show();
+          });
+          jQuery("#check-php-compatibility-status").fadeOut(400, function() {
+            jQuery(this).html('<div>' + seravo_updates_loc.compatibility_run_fail + '</div>').fadeIn(400);
+          });
+        } else {
+          jQuery("#check-php-compatibility-status").fadeOut(400, function() {
+            jQuery(this).html('<div>' + seravo_updates_loc.compatibility_check_done + '</div><div style="color:red;font-weight:bold">' + data['output'] + '</div>').fadeIn(400);
```
cr0vy: Now this possible error message is the entire output by exec command. This string cannot be translated. It's should be translatable.

mvierma: Changed it so that only the number of errors is picked from the output string and used as a variable. Now the string is fully translatable.
https://github.com/Seravo/seravo-plugin/blob/f5cda38aa25a3cb1aeae345e9170bbc0e2570ea4/modules/updates.php#L98
https://github.com/Seravo/seravo-plugin/blob/f5cda38aa25a3cb1aeae345e9170bbc0e2570ea4/lib/updates-ajax.php#L66
https://github.com/Seravo/seravo-plugin/blob/f5cda38aa25a3cb1aeae345e9170bbc0e2570ea4/js/updates.js#L176

PR #280, opened 2019-08-30
Add support for non-sortable full-width postboxes

l3ku:
Add support for displaying full-width postboxes on the top of `seravo_postboxes_page` by adding the `'full_width'` postbox content parameter that can be used as the fifth parameter for `seravo_add_postbox()`.

Related to the discussions in https://github.com/Seravo/seravo-plugin/issues/233.

Locally or on a test site, try adding postboxes to pages with `full_width` context, e.g.:
```
seravo_add_postbox(
   'full-width-test',
   __('Full-width Postbox', 'seravo'),
   array( __CLASS__, 'database_adminer_postbox' ),
   'tools_page_database_page',
   'full_width'
);
```

#### Manual testing steps?
The full-width postboxes should be displayed on the top of the page before the sortable postboxes.

#### Screenshots
Firefox_Screenshot_2019-08-30T08-21-32 708Z
Firefox_Screenshot_2019-08-30T07-46-09 691Z
Firefox_Screenshot_2019-08-30T07-29-11 729Z

### DIFF
--- a/lib/seravo-postbox.php
+++ b/lib/seravo-postbox.php
@@ -203,7 +203,7 @@ if ( ! class_exists('Seravo_Postbox_Factory') ) {
       // Loop through the postboxes for this context
       if ( isset($this->postboxes[ $screen ][ $context ]) && ! empty($this->postboxes[ $screen ][ $context ]) ) {
         foreach ( $this->postboxes[ $screen ][ $context ] as $postbox_id => &$postbox_content ) {
-          $this->display_single_postbox($postbox_id, $postbox_content);
+          $this->display_single_postbox($postbox_id, $context, $postbox_content);
         }
       }
     }
@@ -224,6 +224,15 @@ if ( ! class_exists('Seravo_Postbox_Factory') ) {
 
       <!-- Postbox wrapper -->
       <div class="dashboard-widgets-wrap">
+
+        <!-- Full-width postboxes first with full_width context -->
+        <div class="metabox-holder seravo-postbox-holder full-width">
+          <div class="postbox-container">
+            <?php $this->do_postboxes($current_screen, 'full_width'); ?>
+          </div>
+        </div>
+
+        <!-- Sortable smaller postboxes -->
         <div id="dashboard-widgets" class="metabox-holder seravo-postbox-holder">
           <?php foreach ( $container_contexts as $container_context ) : ?>
             <div id="postbox-container-<?php echo $context_index; ?>" class="postbox-container">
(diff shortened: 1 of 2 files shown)


l3ku:
> Overally LGTM. But there is one thing.

What thing? I also can't see any code comments which would request changes?

lib/seravo-postbox.php
```diff
@@ -263,8 +272,15 @@ class="postbox seravo-postbox <?php echo $closed ? 'closed' : ''; ?>">
           <span class="toggle-indicator" aria-hidden="true"></span>
         </button>
 
+        <?php
+        // All other contexts but full_width are sortable
+        $handle_class = 'hndle';
+        if ( $postbox_content !== 'full_width' ) {
```
cr0vy: Most likely typo?

l3ku: Thanks for noticing, I have now fixed the variable name from `$postbox_content` to `$postbox_context`.

l3ku:
@cr0vy what is your own opinion about what would be the best features into postboxes or other tools that would help in solving the UI challenges in https://github.com/Seravo/seravo-plugin/issues/233? Would you think that this this full-width postbox a good tool for that along with support for limiting the amount of postbox columns? What about the tabs I showcased in https://github.com/Seravo/seravo-plugin/issues/233#issuecomment-526590313, maybe we could also add support for non-postbox tabs?

cr0vy:
@l3ku I like the idea of the tabs, but my opinion is that the best solution for the issue #233 is limiting amount of postbox columns. At the same time, we should consider that some pages could need full-width postboxes, too. 

Also, I tested full-width postboxes with toolbox code and full-width postboxes broke up on some pages. It looks like full-width postboxes don't like, if two or more separate php files (example: database.php and reports.php) try to add postboxes for the page.

l3ku:
> @l3ku I like the idea of the tabs, but my opinion is that the best solution for the issue #233 is limiting amount of postbox columns. At the same time, we should consider that some pages could need full-width postboxes, too.
> 
> Also, I tested full-width postboxes with toolbox code and full-width postboxes broke up on some pages. It looks like full-width postboxes don't like, if two or more separate php files (example: database.php and reports.php) try to add postboxes for the page.

Nothing should break in full-width postboxes even if different PHP files register them, as long as they are registered to the same admin screen/page. If you find a better mechanism for creating full-width postboxes along with the column width limitations, feel free to implement them in a separate development branch.

ottok:
What do you think, do we need this or should we scrap the idea (and thus implementation)?

l3ku:
If there is no need for this, we might as well drop it.

PR #282, opened 2019-09-06
Add 'two column' postboxes page layout

cr0vy:
This PR will add an optional two column page layout to postboxes. Also, in two column postboxes, if window width is more than 2300px, it will limit postboxes width to 1050px.

The issue #233

#### Manual testing steps?
Set any add_submenu_page() look like:
 ```
add_submenu_page(
    'tools.php',
    __('Example', 'seravo'),
    __('Example', 'seravo'),
    'manage_options',
    'example_page',
    'Seravo\seravo_two_column_postboxes_page'
 );
```

#### Screenshots
width 1920px:
fullhd

width 3840px:
big

### DIFF
--- a/lib/seravo-postbox.php
+++ b/lib/seravo-postbox.php
@@ -156,7 +156,7 @@ if ( ! class_exists('Seravo_Postbox_Factory') ) {
     /**
      * Filter the postboxes in such a way that they are in the user specified order.
      */
-    private function apply_user_postbox_settings() {
+    private function apply_user_postbox_settings( $column_count = 'four_column' ) {
       $screen = get_current_screen()->id;
 
       // Preload closed postboxes. If the setting is not set, WP returns an empty string, so don't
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/seravo-postbox.php [hunk 1]
```diff
@@ -211,16 +211,58 @@ private function do_postboxes( $screen, $context ) {
...
-      // These are the same postbox contexts that are used in WP core.
-      $container_contexts = array( 'normal', 'side', 'column3', 'column4' );
+    public function display_postboxes_page( $column_count = 'four_column' ) {
+      if ( $column_count === 'four_column' ) {
+        // These are the same postbox contexts that are used in WP core.
+        $container_contexts = array( 'normal', 'side', 'column3', 'column4' );
+      } else if ( $column_count === 'two_column' ) {
+        $container_contexts = array( 'normal', 'side' );
+      }
+
```
l3ku: Should an error be raised if `$column_count` is neither `four_column` or `two_column`? Or then default to `two_column`. This is not a very common scenario as only pre-defined wrapper code in `seravo-postbox.php` like `seravo_postboxes_page` calls this function, but currently there is a possibility that `$container_contexts` is left undefined.

lib/seravo-postbox.php
```diff
@@ -211,16 +211,58 @@ private function do_postboxes( $screen, $context ) {
...
+          @media only screen and (min-width: 800px) and (max-width: 1499px) {
+            #wpbody-content #dashboard-widgets .postbox-container {
+              width: 49.5%;
+            }
+
+            #wpbody-content #dashboard-widgets #postbox-container-2 {
+              width: 50.5%;
+            }
+          }
+        </style>
```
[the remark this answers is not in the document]
cr0vy: This is now in ```seravo-postbox.css```.

lib/seravo-postbox.php
cr0vy: Function parameter is "( $column_count = 'four_column' )". So four_column should be default. But I change it to check only 'two_column'.

cr0vy:
> How should we handle the following scenario: if the user has saved the postbox to locate in either `column3` or `column4` and the page is transformed to two-columns, those postboxes will no longer be displayed _**at all**_ on that page when using the two-column layout. The simplest way would of course be to modify the postbox outputting code to ensure that every postbox registered for that page gets displayed. Could you investigate whether it is possible to implement such a feature easily? If not, then we have to either completely remove the ordering feature or ensure that the ordering option gets deleted when migrating from four columns to two columns. There are also some code changes.

Latest commit contains four column context -> two_column context modify feature.

lib/seravo-postbox.php
l3ku: The idea was that if someone calls it with an invalid parameter, e.g. `display_postboxes_page('hundred_column')`, then what would happen. But as said previously, it is currently not intended usage to call that function outside `seravo-postbox.php`, so it is not a common scenario. It just helps to think functions as an interface with predictable behaviour: if I pass something that is invalid as a parameter, there should be some error.

PR #283, opened 2019-09-17
Check the exit code of the wp-backup command on search-replace ajax

cr0vy:
Add check for the third line of wp-backup output to search-replace-ajax.

Currently, search-replace-ajax check only if the fourth line of wp-backup output is "Success: Exported to 'example.sql'". But some sites wp-backup output not contains "No path given as argument. Using default path." and "Success: Exported to 'example.sql'" is in the third line of output. This causes search-replace failing.

#### Screenshots
Before change:
search-replace-fail

After change:
search-replace-success


ottok:
Could you use the exit code of the script instead? Checking for string X in line Y seems a bit brittle. In this case I broke it by changing the script and was unaware that there is a requirement for the success message to be on a certain line. Error codes are more of a convention on shell scripts as a pass/fail signal, using exit code would be idea if possible?

Otherwise we need to document in each script something like "WARNING: X uses this and expects string Y, don't change it."

Issue #284, opened 2019-09-26
PHP warning: Cannot modify header information


cr0vy:
**Describe the bug**
On some sites php-error.log get a new entry on regular intervals, which is `PHP Warning: Cannot modify header information - headers already sent in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/canonical-domain-and-https.php on line 38`. This happens only URL with wp-cron.php.

**To Reproduce**
1. Go to site CLI on SSH
2. Run `tail -fn 0 /data/log/php-error.log`
3. Run in another terminal `curl -ILs http://example.com/wp-cron.php` or `curl -ILs example.com/wp-cron.php`
4. Check php-error.log entry

**Expected behavior**
Nothing should happen in /data/log/php-error.log

**Screenshots**
http:
![cannot_modify_header_information](https://user-images.githubusercontent.com/43742796/65668536-91cf6280-e04a-11e9-972f-fe0388069536.png)

https:
![https-curl](https://user-images.githubusercontent.com/43742796/65668549-972cad00-e04a-11e9-82a2-12868995d835.png)

ottok:
Is there a `?>` in some file somewhere maybe? That will emit a newline.

The PHP files should not emit anything before WP actually starts to print out the real page.

ottok:
To recap: this is not a problem in `canonical-domain-and-https.php`. That is just the first thing to emit headers. The problem is somewhere else that emits stuff before that.

I came across a similar situation:
```
$ wp option get siteurl

https://<site>.fi-p.seravo.com
$ 
```

The correct output should be:
```
$ wp option get siteurl
https://<site>.fi-p.seravo.com
$ 
```

So there is somewhere an extra newline emitted.

I found the problem here:
```
gp-child$ head functions.php 

<?php
// Exit if accessed directly
if ( !defined( 'ABSPATH' ) ) exit;
```

So the newline was at the beginning for the theme functions.php

cr0vy:
I tested a few commit checkout to find the first commit to produce this warning.

Commit 2e4b4b1 (Bump version to 1.8.0) -> Ok
Commit 2eb11f8 (Bump version to 1.9.0) -> PHP Warning

Commit d6316fa (Remove function isPublic() as it does not seem to be used anywhere) -> Ok
Commit 64173bc (Enforce usage of https if siteurl and home has it set) -> The first commit with this PHP Warning

JoosuaKoskinen:
[...]

PR #285, opened 2019-10-01
Add 'headers_sent' check to the https redirect

cr0vy:
Issue #284

### DIFF
--- a/lib/canonical-domain-and-https.php
+++ b/lib/canonical-domain-and-https.php
@@ -29,7 +29,7 @@ if ( ! class_exists('Canonical_Domain_And_Https') ) {
 
       if ( strpos($siteurl, 'https') !== false && strpos($home, 'https') !== false ) {
         // Site uses https
-        if ( is_ssl() === false && isset($_SERVER['HTTP_HOST']) && isset($_SERVER['REQUEST_URI']) ) {
+        if ( headers_sent() === false && is_ssl() === false && isset($_SERVER['HTTP_HOST']) && isset($_SERVER['REQUEST_URI']) ) {
           // Request did not use https, force redirect
           $url = 'https://' . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'];
           // error_log("Redirect to $url");


ottok:
How did you test this? Should one check for if the constant is not true of if the constant is not set at all?

cr0vy:
DOING_CRON define should only be in the wp-cron.php file. But I agree that we should check 'DOING_CRON' constant. I added this check.

Testing:
1. Go to your site command line by ssh
2. Run `tail -fn 0 /data/log/php-error.log`
3. Run in another terminal session: `curl -IL http://example.com/wp-cron.php`
4. Check tail entry from step 2, should be empty

cr0vy:
I found a couple of `Cannot modify header information by fastcgi_finish_request` from WordPress bug reports tickets and those were on 'wontfix' resolution. This header_sent() check was a recommendation by WordPress developer.

WordPress reports:
https://core.trac.wordpress.org/ticket/46417
https://core.trac.wordpress.org/ticket/46719

Testing:
Go to your site command line by ssh
Run tail -fn 0 /data/log/php-error.log
Run in another terminal session: curl -IL http://example.com/wp-cron.php
Check tail entry from step 2, should be empty

ottok:
Added @JoosuaKoskinen as reviewer, I am a bit busy right now..

lib/canonical-domain-and-https.php
```diff
@@ -29,10 +29,10 @@ public static function load() {
 
       if ( strpos($siteurl, 'https') !== false && strpos($home, 'https') !== false ) {
         // Site uses https
-        if ( is_ssl() === false && isset($_SERVER['HTTP_HOST']) && isset($_SERVER['REQUEST_URI']) ) {
+        if ( headers_sent() === false && is_ssl() === false && isset($_SERVER['HTTP_HOST']) && isset($_SERVER['REQUEST_URI']) ) {
           // Request did not use https, force redirect
           $url = 'https://' . $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'];
-          // error_log("Redirect to $url");
+          error_log("Redirect to $url");
```
ottok: Please don't emit anything to php-error.log in production unless there is a major problem. Otherwise the log would just fill up.

ottok:
There commit message does not explain _why_ this change was made and there are no linline comments added now. Now the PR comment includes "headers_sent() check was a recommendation by WordPress developer" which seems to be the final motivation why this change was made as it is now.

Issue #292, opened 2019-10-10
Current Seravo Plugin master tip breaks wp-test: PHP Fatal error:  Uncaught Error: Class 'Seravo\Domains' not found


ottok:
While running `wp-test` with latest Seravo Plugin git master:
```
Acceptance Tests (2) ------------------------------------------------------------------------------------
⏺ Recording ⏺ step-by-step screenshots will be saved to /data/reports/tests/
Directory Format: record_5d9f917eb3919_{filename}_{testname} ----
✔ SeravoCheckWPHomeCest: Try to open home (1.64s)
✖ SeravoCheckWPLoginCest: Try to login and access wp admin (38.57s)
---------------------------------------------------------------------------------------------------------
⏺ Records saved into: file:///data/reports/tests/records.html


Time: 40.83 seconds, Memory: 12.00MB

There was 1 failure:

---------
1) SeravoCheckWPLoginCest: Try to login and access wp admin
 Test  seravo/acceptance/SeravoCheckWPLoginCest.php:tryToLoginAndAccessWPAdmin
 Step  See element "#wpadminbar"
 Fail  Browser console shows errors: 
SEVERE: https://zabbixtest.tre-test.seravo.co/wp-admin/ - Failed to load resource: the server responded with a status of 500 ()

Scenario Steps:

 14. $I->checkBrowserConsole() at seravo/acceptance/SeravoCheckWPLoginCest.php:53
 13. $I->makeScreenshot("wp-admin") at seravo/acceptance/SeravoCheckWPLoginCest.php:52
 12. $I->amOnPage("/wp-admin/") at seravo/acceptance/SeravoCheckWPLoginCest.php:51
 11. $I->getConsoleLog() at seravo/acceptance/SeravoCheckWPLoginCest.php:47
 10. $I->amOnPage("/wp-login.php?seravotest-auth-bypass=c50c8c78f97cf65de24a13256b87cbe46d02e2fbbf5a...") at seravo/acceptance/SeravoCheckWPLoginCest.php:44
 9. $I->seeElement("#wpadminbar") at seravo/acceptance/SeravoCheckWPLoginCest.php:40

Artifacts:

Png: /data/reports/tests/SeravoCheckWPLoginCest.tryToLoginAndAccessWPAdmin.fail.png
Html: /data/reports/tests/SeravoCheckWPLoginCest.tryToLoginAndAccessWPAdmin.fail.html

FAILURES!
[...]

Issue #291, opened 2019-10-10
Re-review commit d1934949c97e577a41bca53b2660917394faf6f0


ottok:
This commit removes a huge amount of code lines and adds a huge amount of code lines. Git's features to track moved content are not used in any way. We need to re-review this to track down all code that was simply removed and not added/moved anywhere.

```
commit d1934949c97e577a41bca53b2660917394faf6f0
Author: Samuli Lehtonen <samuli@seravo.fi>
Date:   Tue Sep 10 14:21:18 2019 +0300

    Reorganize optimize-images, reports, and shadows into site status

 js/optimize-images.js                         |  12 --
 js/reports.js                                 |  85 ----------
 js/shadows.js                                 |  87 ----------
 js/sitestatus.js                              | 167 ++++++++++++++++++
 lib/shadows-ajax.php                          |  39 -----
 lib/{reports-ajax.php => sitestatus-ajax.php} |  31 +++-
 modules/optimize-images.php                   | 187 ---------------------
 modules/reports.php                           | 238 --------------------------
 modules/shadows.php                           | 174 -------------------
 modules/sitestatus.php                        | 465 +++++++++++++++++++++++++++++++++++++++++++++++++++
 seravo-plugin.php                             |  32 +---
 style/optimize-images.css                     |   3 -
 style/reports.css                             |   3 -
 style/{shadows.css => sitestatus.css}         |   0
 14 files changed, 665 insertions(+), 858 deletions(-)
```

Many bugs in the latest version stem from this, for example that the optimize images javascript is now broken and we must postpone releasing 1.9.13..

Issue #290, opened 2019-10-10
Error handling in shadows is lacking


ottok:
When running latest master tip in a shadow site, I notice these places where the Seravo Plugin does not fail gracefully:
![image](https://user-images.githubusercontent.com/668724/66570067-7cdbee80-eb75-11e9-9c89-1eb0a1383f10.png)

Upkeep page:
![image](https://user-images.githubusercontent.com/668724/66570187-c3c9e400-eb75-11e9-99e0-5f72c6164fa0.png)

![image](https://user-images.githubusercontent.com/668724/66570208-cfb5a600-eb75-11e9-811b-f38d876c9eac.png)

Domains page:
![image](https://user-images.githubusercontent.com/668724/66570222-d7754a80-eb75-11e9-88ce-1bd8e8399544.png)

Most of the above applies for Vagrant as well. The plugin should not show features in the environments where they are not available. The code to check this and implement it was previously there, so this is partially a regression and/or laps when adding new features.

ottok:
Previously mails page was not visible in the menus of shadows and local development because of checks like this (remove in recent re-structuring):
![image](https://user-images.githubusercontent.com/668724/66571011-69318780-eb77-11e9-94b7-2927fcaf6b29.png)

If there is no `getenv('CONTAINER')` then there is no SWD API nor any mail admin features.

Issue #287, opened 2019-10-10
Views unaligned


ottok:
The layout is misaligned in two ways here: the margins don't align and the user experience is not aligned.

![image](https://user-images.githubusercontent.com/668724/66569416-36d25b00-eb74-11e9-886c-0d8ab998bf85.png)

We basically have two lists here. Couldn't they have the exact same layout, would that make sense?

ottok:
This is not a complete UI design, I just did a quick mockup to illustrate the need:
![image](https://user-images.githubusercontent.com/668724/66804219-92ad3300-ef2a-11e9-9ea1-b66979a02fbf.png)

JoosuaKoskinen:
@cr0vy do you still have some fix coming to this? My email postbox still seems to have 50% width. I don't want to have any conflicts with https://github.com/Seravo/seravo-plugin/issues/13.

PR #293, opened 2019-10-14
Fix bugs that came with the Seravo Toolbox PR

cr0vy:
The main changes in this PR are the fixes for bugs that came with the Seravo Toolbox.

Issues #287, #290, #291 and #292

Testing in shadow or vagrant with seravotest account and another account which is an administrator account.

#### Manual testing steps?
1. In the shadow or vagrant, check toolbox pages
2. In multisite WordPress, test with seravotest < site >/wp-admin
3. Test optimize-images from Site Status

#### Screenshots

Vagrant Multisite WordPress in wordpress.local/wp-admin with seravotest:
vagrant-multisite-seravotest

Vagrant Multisite WordPress with admin account:
vagrant_backups
vagrant_database
vagrant_security
vagrant_site_status
vagrant_upkeep

Production with admin account:
production_domains

### DIFF
--- a/js/cruftplugins.js
+++ b/js/cruftplugins.js
@@ -12,7 +12,7 @@ jQuery(document).ready(function ($) {
       },
       function (rawData) {
         var data = JSON.parse(rawData);
-        console.log(data)
+
         if (data[data.length - 1].indexOf('Success: Deleted 1 of 1 plugins.') != -1) {
           callback();
         } else {
--- a/js/sitestatus.js
+++ b/js/sitestatus.js
@@ -1,6 +1,16 @@
 // phpcs:disable PEAR.Functions.FunctionCallSignature
 'use strict';
 
+(function($) {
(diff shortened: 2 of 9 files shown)


js/cruftplugins.js
```diff
@@ -12,7 +12,7 @@ jQuery(document).ready(function ($) {
       },
       function (rawData) {
         var data = JSON.parse(rawData);
-        console.log(data)
```
ottok: Good spot! We should not output any JS console log stuff in production. Debugging is only for development.

js/sitestatus.js
```diff
@@ -1,6 +1,16 @@
 // phpcs:disable PEAR.Functions.FunctionCallSignature
 'use strict';
 
+(function($) {
```
ottok: You could have just not deleted the old file, but adding it's contents in another JS file that is loaded in the same view is OK.

PR #296, opened 2019-10-18
Mixed fixes for Seravo Plugin

cr0vy:
This PR include fixes for the next things:
  - Fix a few typos in the upkeep page code
  - Change usort_reorder to usort_reorder_mail in the domains page code
  - Fix a bug in wp-cli that causes error message on command `wp seravo updates` when Seravo Updates is disabled

Testing:
  - On the upkeep page, disable or enable Seravo updates and save settings
  - View the zones in your domain
  - Once Seravo updates is disabled, run `wp seravo updates` in the site CLI

### DIFF
--- a/lib/domains-page.php
+++ b/lib/domains-page.php
@@ -193,7 +193,7 @@ class Seravo_Domains_List_Table extends WP_List_Table {
         // Send final sort direction to usort
       return ($order === 'asc') ? $result : -$result;
     }
-    usort($data, 'usort_reorder');
+    usort($data, array( __CLASS__, 'usort_reorder' ));
 
     // Required for pagnation
     $current_page = $this->get_pagenum();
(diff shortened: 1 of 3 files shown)


lib/domains-page.php
```diff
@@ -193,7 +193,7 @@ function usort_reorder( $a, $b ) {
         // Send final sort direction to usort
       return ($order === 'asc') ? $result : -$result;
     }
-    usort($data, 'usort_reorder');
+    usort($data, array( __CLASS__, 'usort_reorder' ));
```
ottok: Why is class name used here?

lib/domains-page.php
```diff
@@ -354,7 +354,7 @@ function usort_reorder( $a, $b ) {
       // Send final sort direction to usort
       return ($order === 'asc') ? $result : -$result;
     }
-    usort($data, 'usort_reorder');
+    usort($data, 'usort_reorder_mail');
```
ottok: ..but class name is not used here?

Issue #281, opened 2019-08-30
Replace current cache check Seravo postbox with output of wp-check-http-cache


ottok:
Since we have now rolled out `wp-check-http-cache` we should use it to show the cache status post box contents.

JoosuaKoskinen:
Related to https://github.com/Seravo/seravo-plugin/issues/72

JoosuaKoskinen:
Should wp-check-http-cache replace the whole postbox? It shows way less data than the current one. Maybe it should only replace the curl section.

```
joosuakoskinen@joosuakoskinen_bd3ab9:~$ wp-check-http-cache
----------------------------------------
Seravo HTTP cache checker
----------------------------------------
Testing https://alidomain-2.fi-s.seravo.com/...
Request 1: MISS
Request 2: HIT
Request 3: HIT
----------------------------------------
SUCCESS: HTTP cache works for https://alidomain-2.fi-s.seravo.com/.
----------------------------------------
You can also test this yourself by running:
  curl -IL https://alidomain-2.fi-s.seravo.com/
```

JoosuaKoskinen:
This will be bundled with https://github.com/Seravo/seravo-plugin/issues/72.

JoosuaKoskinen:
Closing, '`wp-check-http-cache`' used instead of cURL since https://github.com/Seravo/seravo-plugin/pull/300.

Issue #72, opened 2017-10-24
Improve the cache status report


ottok:
Our current Reports page has a very small and naive cache tester section. It could be removed and replaced with a separate page, that makes at least 5 test curls on the website to determine the current HIT/MISS behavior and show a nicely formatted report about the test result. 

There should also be some listing of relevant HTTP headers and pointers to WordPress documentation so developers could more easily grasp what is the status of caching on the site and how they should change their code base.

In addition there could also be a section that shows historical cache HIT/MISS ratios based on data extracted from `/data/slog/`.

ottok:
Current cache status box:
![image](https://user-images.githubusercontent.com/668724/60666699-53be0a80-9e70-11e9-93b5-f8e67abe092f.png)

This should be split into "HTTP cache status" and "Redis cache status".

JoosuaKoskinen:
**Status update:** Will take care of this soon. The toolbox update made changes to almost everything so there's a lot of conflicts to be fixed.

JoosuaKoskinen:
Cache status postbox was improved in https://github.com/Seravo/seravo-plugin/pull/300, closing...

PR #300, opened 2019-11-25
Add cache status improvements

JoosuaKoskinen:
Improves the Redis info looks, replaces the curl with `wp-check-http-cache` and
adds parsing of `/data/slog` for historical info.

Issues #281 and #72 .

#### Screenshots
**Old:**

**New:**
new-cache-postbox

### DIFF
--- a/lib/sitestatus-ajax.php
+++ b/lib/sitestatus-ajax.php
@@ -116,32 +116,13 @@ function seravo_report_redis_info() {
 }
 
 function seravo_report_front_cache_status() {
-  exec('curl -ILk ' . get_site_url(), $output);
-  array_unshift($output, '$ curl -ILk ' . get_site_url());
+  exec('wp-check-http-cache ' . get_site_url(), $output);
+  array_unshift($output, '$ wp-check-http-cache ' . get_site_url());
 
-  if ( preg_match('/X-Proxy-Cache: ([A-Z]+)/', implode("\n", $output), $matches) ) {
-
-    switch ( $matches[1] ) {
-      case 'HIT':
-      case 'EXPIRED':
-        $result = 'Front cache is working correctly.';
-        break;
-
-      case 'MISS':
-        $result = 'Front page is not cached due to cookies or expiry headers emitted from the site.';
-        break;
-
-      default:
-        $result = 'Unable to detect front cache status.';
-        break;
-    }
-  } else {
-    $result = 'No front cache available in this WordPress instance.';
-  }
-
-  array_unshift($output, $result, '');
-
-  return $output;
+  return [
+    'success' => strpos(implode("\n", $output), "\nSUCCESS: ") == true,
+    'test_result' => $output,
+  ];
 }
 
 function seravo_reset_shadow() {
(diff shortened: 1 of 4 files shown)


lib/sitestatus-ajax.php
```diff
@@ -91,24 +91,23 @@ function seravo_report_git_status() {
...
   );
 
-  if (isset($hits) && isset($misses)) {
+  if ( isset($hits) && isset($misses) ) {
     $total = $hits + $misses;
-    if ($total > 0) {
+    if ( $total > 0 ) {
       array_push($output, 'Keyspace hit rate: ' . round(($hits / $total) * 100) . '%');
     }
   }
```
cr0vy: I think this code style fix should be in the commit 448da4f30 (the first commit in this PR), which also adds this part of the code.

lib/sitestatus-ajax.php
```diff
@@ -91,24 +91,23 @@ function seravo_report_git_status() {
...
+    } else if ( strpos($line, 'keyspace_misses') === 0 ) {
       $misses = explode(':', $line)[1];
     }
   }
 
   $output = str_replace(
-    ['expired_keys:', 'evicted_keys:', 'keyspace_hits:', 'keyspace_misses:'],
-    ['Expired keys: ', 'Evicted keys: ', 'Keyspace hits: ', 'Keyspace misses: '],
+    [ 'expired_keys:', 'evicted_keys:', 'keyspace_hits:', 'keyspace_misses:' ],
+    [ 'Expired keys: ', 'Evicted keys: ', 'Keyspace hits: ', 'Keyspace misses: ' ],
```
ottok: As @cr0vy stated, the pull request should not modify itself = the pull request should not modify the same code twice.https://docs.google.com/presentation/d/1G1sllxOIHWuCVtj-ITnJZK3epO8-ImLiBab1IPG9WUA/edit#slide=id.g5caaae5766_0_6

lib/sitestatus-ajax.php
```diff
@@ -115,6 +115,41 @@ function seravo_report_redis_info() {
   return $output;
 }
 
+function seravo_report_longterm_cache_stats() {
```
ottok: You can have this code here, but on the long term it would be better to have this info in the Goaccess report or as a command line command wp-check-http-cache-stats etc so that it can be used from the command line as well.

lib/sitestatus-ajax.php
JoosuaKoskinen: This should be fixed now.

lib/sitestatus-ajax.php
JoosuaKoskinen: This should be fixed now.

lib/sitestatus-ajax.php
JoosuaKoskinen: Sure. I wasn't sure which files should be parsed from /data/slog etc, so I think we could just leave this here for now. I can make a new feature request about the command so it can be implemented with better planning.

ottok:
@JoosuaKoskinen Please check your `git config user.email` in your repo, now all of your commits are made as `Joosua Koskinen <koskinen.joosua@gmail.com>`.

JoosuaKoskinen:
> @JoosuaKoskinen Please check your `git config user.email` in your repo, now all of your commits are made as `Joosua Koskinen <koskinen.joosua@gmail.com>`.

Not sure how that happened but it's fixed now for these commits.

Issue #308, opened 2020-01-08
Cruft remover would check content also in home directory


simoke:
Currently the Cruft remover checks known files and directories under /data/wordpress/htdocs directory, but in many cases, customers also saves data (eg. SQL dumps) under home directory, and there are plenty of such data at the moment.

It would be good improvement to check also the home directory, and perhaps from another paths.

JoosuaKoskinen:
Here's some I've seen that could be removed:
~/*backup*
~/core
~/*.bak
~/.sql
~/*palautus*
~/*before*
~/<sitename>_<yyyymmdd>

PR #311, opened 2020-01-09
Add searching cruft from home directory

JoosuaKoskinen:

Adds some new known cruft files/dirs and
also searches for cruft in home directory.

Closes: #308.

#### Manual testing steps?

Create some files matching the cruft specifications
and let cruftfiles remover on security page take care of them.

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -16,13 +16,19 @@ function add_file_information( $file ) {
 }
 
 function find_cruft_file( $name ) {
-  exec('find /data/wordpress -name ' . $name, $output);
-  return $output;
+  $user = getenv('WP_USER');
+  exec('find /data/wordpress -name ' . $name, $data_files);
+  exec('find /home/' . $user . ' -maxdepth 1 -name ' . $name, $home_files);
+  $files = array_merge($data_files, $home_files);
+  return $files;
 }
 
 function find_cruft_dir( $name ) {
-  exec('find /data/wordpress -type d -name ' . $name, $output);
-  return $output;
+  $user = getenv('WP_USER');
+  exec('find /data/wordpress -type d -name ' . $name, $data_dirs);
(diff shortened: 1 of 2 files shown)


lib/cruftfiles-ajax.php
```diff
@@ -76,11 +82,15 @@ function seravo_ajax_list_cruft_files() {
...
         '*.wpress',
+        'core',
+        '*.bak',
       );
       // List of known cruft directories
       $list_dirs = array(
         'siirto',
-        'palautus',
+        '*palautus*',
+        '*restore*',
```
ottok: Some plugins or themes have files that include 'restore' in them for valid reasons. Our cruft remover should be as safe as possible. If we recommend removing too much customers might get angry when we "trick" them into deleting too much files from their site?

This is 99.9% correct it '*restore*' is found in the top level of a home directory, e.g. /home/user/yesterday-restore-backup, but it is only 90% correct if found somewhere else, e.g. /data/wordpress/htdocs/wp-content/plugins/xyz/lib/quick-restore/. Just saying we need to be conservative here, or add more constraints..

JoosuaKoskinen: True, we could make it so some are only searched from home directory.

JoosuaKoskinen: @ottok I decided to remove the 'restore' entry and added a warning message stating that not everything is safe to delete.

modules/security.php
```diff
@@ -293,7 +293,7 @@ public static function logins_info_postbox() {
     public static function cruftfiles_postbox() {
       ?>
       <p>
-        <?php _e('Find and delete any extraneous and potentially harmful files taking up space in the file system.', 'seravo'); ?>
+        <?php _e('Find and delete any extraneous and potentially harmful files taking up space in the file system. Note that not everything is neccessarily safe to delete..', 'seravo'); ?>
```
elguitar: Typo in necessarily

JoosuaKoskinen: Fixed the typo

PR #312, opened 2020-01-09
Cache domain list for instance switcher

JoosuaKoskinen:

Currently, as instance switcher is loaded on every page
load, so is the domain list queried from the API.

This caches the list for 30 minutes as it's very rarely updated
and when it's updated, the cache is purged anyways during search-replace.


JoosuaKoskinen:
Not needed after a change in SWD API: https://github.com/Seravo/swd/pull/750

Closing and making a fix to fit this new situation.

Issue #314, opened 2020-01-20
Review alloptions fix


JoosuaKoskinen:
Seravo Plugin has a fix for alloptions not properly updating in cache. Now after 5 years, the ticket has been closed (https://core.trac.wordpress.org/ticket/31245) and a fix has been implemented in the core 5.3.1.

It's now important to find out if the old fix must be disable in Seravo Plugin for sites with WP version >= 5.3.1. There has already been a report of issues with two fixes running at the same time. Also it's possible that they slow down sites as now cache is updated twice for every update in alloptions.

JoosuaKoskinen:
I have now tested this. It seems that the fix in WP core works as well as the one in Seravo Plugin. Also tried running them both at the same time and saw a minor performance decrease. It was barely noticeable but might make a huge difference on bigger sites.

PR #317, opened 2020-01-20
Deprecate alloptions fix (Closes: #314)

JoosuaKoskinen:

A similar fix has been implemented in WP core 5.3.1 and has
been tested. Running two of these fixes at once was found
out to be more harmful.

The fix is now only ran with WordPress versions lower than 5.3.1.

Issue #314.

### DIFF
--- a/modules/fixes.php
+++ b/modules/fixes.php
@@ -45,10 +45,16 @@ if ( ! class_exists('Fixes') ) {
       /**
        * Additional hooks to option updates to ensure they get refreshed in the
        * Redis object-cache when they change.
+       *
+       * WP core has implemented a similar fix in 5.3.1,
+       * this has been depreacted since that.
        */
-      add_action('added_option', array( __CLASS__, 'maybe_clear_alloptions_cache' ));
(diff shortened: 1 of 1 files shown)


modules/fixes.php
```diff
@@ -45,10 +45,16 @@ public static function load() {
...
        * Redis object-cache when they change.
+       *
+       * WP core has implemented a similar fix in 5.3.1,
+       * this has been depreacted since that.
        */
-      add_action('added_option', array( __CLASS__, 'maybe_clear_alloptions_cache' ));
-      add_action('updated_option', array( __CLASS__, 'maybe_clear_alloptions_cache' ));
-      add_action('deleted_option', array( __CLASS__, 'maybe_clear_alloptions_cache' ));
+
+      if ( version_compare(get_bloginfo('version'), '5.3.1') === -1 ) {
```
elguitar: The readability of this line could be increased, if `operator` parameter would be used. See: https://www.php.net/manual/en/function.version-compare.php

JoosuaKoskinen: Using `operator` parameter now :)

ottok:
Looks good to me, feel free to merge once you addressed Pete's comment.

Issue #297, opened 2019-10-28
Whitelist Jetpack IP addresses from XML-RPC blocker


ottok:
Most sites don't have any valid uses for XML-RPC anymore. Most sites use WP-JSON API.

However, Jetpack does use XML-RPC still. Our XML-RPC block should whitelist the Jetpack IP addresses. See `class Security_Restrictions`.

ypcs:
Somewhat related upstream (Jetpack) issue about Automattic IP address ranges, https://github.com/Automattic/jetpack/issues/1719

JoosuaKoskinen:
Here's a list of Jetpack IPv4 ranges I found:
https://jetpack.com/ips-v4.txt
https://jetpack.com/ips-v4.json

ottok:
Right now I am not aware of anything else that should be whitelisted, so I think we could put these IPs directly in the Seravo Plugin for now instead of making any other more elaborate whitelisting list automation as discussed elsewhere.

PR #319, opened 2020-02-05
Use XML-RPC block by default and whitelist Jetpack

JoosuaKoskinen:
- XML-RPC blocking is now enabled by default
- Jetpack is whitelisted from XML-RPC blocking

Closes: #297

### DIFF
--- a/lib/security-restrictions.php
+++ b/lib/security-restrictions.php
@@ -22,19 +22,13 @@ if ( ! class_exists('Security_Restrictions') ) {
 
     public static function load() {
 
-      if ( get_option('seravo-disable-xml-rpc') ) {
-        /*
-         * When this is active any request like:
-         *   curl -d '<?xml version="1.0"?> <methodCall> <methodName>wp.getUsersBlogs</methodName> \
-         *    <params> <param> <value>username</value> </param> <param> <value>password</value> </param>\
-         *    </params> </methodCall>' https://<siteurl>/xmlrpc.php
-         * will yield 'faultCode 405' and 'XML-RPC is not available'
-         */
-        add_filter('xmlrpc_enabled', '__return_false');
+      add_action('activate_seravo-plugin/seravo-plugin.php', array( __CLASS__, 'maybe_enable_xml_rpc_blocking' ));
 
-        // Disable X-Pingback to header
-        // since when XML-RPC is disabled pingbacks will not work anyway
-        add_filter('wp_headers', array( __CLASS__, 'disable_x_pingback' ));
+      if ( get_option('seravo-disable-xml-rpc') ) {
+
+        // Block XML-RPC and X-pingback if IP not whitelisted
+        add_filter('xmlrpc_enabled', array( __CLASS__, 'maybe_block_xml_rpc' ));
+
       }
 
       if ( get_option('seravo-disable-json-user-enumeration') ) {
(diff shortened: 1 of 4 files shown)


lib/security-restrictions.php
```diff
@@ -22,7 +22,7 @@ class Security_Restrictions {
 
     public static function load() {
 
-      if ( get_option('seravo-disable-xml-rpc') ) {
+      if ( get_option('seravo-disable-xml-rpc', 'on') === 'on' ) {
```
ottok: How does this work? On new sites the option `seravo-disable-xml-rpc` is not set at all. What happens on new sites vs. existing sites with no setting set with this?

lib/security-restrictions.php
```diff
@@ -96,6 +102,29 @@ public static function disable_user_endpoints( $endpoints ) {
       return $endpoints;
     }
 
+    public function get_jetpack_whitelist() {
+      $url = 'https://jetpack.com/ips-v4.json';
```
ottok: Should we maybe store this permanently in the Seravo Plugin code base instead of fetching it many times over?

JoosuaKoskinen: The IPs are not guaranteed not to change and storing them directly to Seravo Plugin would delay the fix until next Seravo Plugin version. Also, then we would have to constantly see if the IPs have changed.

We could change the cache time from 24 hours to something like a week. The Jetpack servers are very fast and I saw no speed difference even when I disable the cache and fetched the IPs on every page load. And in reality, the fetching is probably done on heartbeat call or something.

lib/security-restrictions.php
JoosuaKoskinen: @ottok This way if the setting has not been set, XML-RPC blocking will be enabled. So this also affects existing sites.

Seravo Plugin has no way of knowing if the site is new or not. If we want existing sites to keep it disabled if the setting hasn't been set, the options are following:
- Enable XML-RPC blocking on deploy with `wp option set seravo-disable-xml-rpc on`
- Loop through all sites and run `wp option set seravo-disable-xml-rpc off` if it's not currently `on`
- Just do it this way and manually run `wp option set seravo-disable-xml-rpc off` if some sites some functionality happens breaks after Seravo Plugin update

I prefer the last one as it's easiest to do and that way not only the new sites have XML-RPC blocking enabled but also the existing sites that don't need XML-RPC for anything (more secure).

elguitar: I would say that blocking XML-RPC on almost all sites is quite hazardous. 

I think that we should set the blocking on deploy and use 'off' as a default here.

lib/security-restrictions.php
elguitar: I think that 24h transient in this situation sounds good.

If the IPs would change, clearing the transient is easier than issuing and deploying a new seravo-plugin version.

lib/security-restrictions.php
ottok: > Seravo Plugin has no way of knowing if the site is new or not.
I'll try to think of some way for it to know..

lib/security-restrictions.php
ottok: hmm.. Actually updating new Seravo Plugin versions is pretty easy I think, but yes it is a mass operation. For sites that don't have object-cache enabled (a few occasional ones in production and all sites running in WP development mode) it would be important to know when actually does it hit this endpoint and make the requests.

JoosuaKoskinen: The Jetpack IPs are now fetched only on XML-RPC requests. This means that there might be a slightly slower XML-RPC request every 24h so I don't think this is an issue anymore. But as said, I didn't see any speed difference even without transient or caching.

lib/security-restrictions.php
JoosuaKoskinen: Not completely sure if this would work with bedrock autoloader but we could do the enabling on Seravo Plugin activation hook call if it's called on the first page load after deploy.

JoosuaKoskinen: I modified it so and it seems to work good. Please review the changes.

ottok:
This should solve the "enable by default" requirement: https://github.com/Seravo/wordpress/pull/130

JoosuaKoskinen:
> This should solve the "enable by default" requirement: [Seravo/wordpress#130](https://github.com/Seravo/wordpress/pull/130)

Indeed. I removed the commit, trying to solve that, from this PR.

JoosuaKoskinen:
Added the description field as we talked:

Screenshot from 2020-02-25 09-59-14

lib/security-restrictions.php
JoosuaKoskinen: Resolved: Seravo/wordpress#130

PR #320, opened 2020-02-06
Increase minimum password requirement to strong

ottok:
This feature was originally introduced in https://github.com/Seravo/seravo-plugin/commit/5218a252d866e1229e44707985b83d940bad3090

Currently the "Update profile" button is disabled if password is weak:
image

The button becomes enabled if password is medium:
image

This change lifts the limit and only strong passwords are allowed:
image

To be discussed: is this a good idea? Is medium enough already or not?

### DIFF
--- a/js/passwords.js
+++ b/js/passwords.js
@@ -18,7 +18,9 @@
       var password = this.value;
       var strength = wp.passwordStrength.meter( password, wp.passwordStrength.userInputBlacklist(), password );
       var $submitButton = $('#wp-submit');
-      if ( strength > 2 ) {
+      // Levels: 0=very weak, 1=weak, 2=medium, 4=strong
+      // Require strong passwords
+      if ( strength > 3 ) {
         $submitButton.prop( 'disabled', false );
       } else {
         $submitButton.prop( 'disabled', true );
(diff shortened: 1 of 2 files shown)


simoke:
Does this cover all kind of users, or just administrators? I prefer strong passwords are forced to use at least for administrative accounts.

nuuttitoivola:
I personally would like to use strong passwords for all users but I know some site owners might want to have easier passwords available, if they for example have some mass registration thing going on in their site etc. 

Strong passwords for admins is a must as mentioned above. 👍

elguitar:
There is no reviewer specified. When should this be merged?

ottok:
> Does this cover all kind of users, or just administrators? I prefer strong passwords are forced to use at least for administrative accounts.

In its current form it would apply to all users.

There is no reviewer assigned, this is for now pending on a decision if we want to do this or not. Mainly we need to figure out are there any concrete drawbacks somewhere.

ottok:
I've rebased this on latest master. What do you think @elguitar, should we merge this in?

elguitar:
This is a good change. I think that this should be merged during January-February.

I wonder what's going on with Travis CI

elguitar (approved):

elguitar (changes requested):
This should be refactored so that admins have to have strong passwords and the others will remain at `strength > 2`

ottok:
Actually this change would only apply for the password reset screen. Maybe not worth it after all.

ottok:
This is actually not feasible to do, as our custom JavaScript only applies to the Reset password screen, not the normal user profile change password screen. Closing as won't implement.

PR #321, opened 2020-02-06
Show alert for pwned password

JoosuaKoskinen:

To improve security, passwords are now checked on login with `wp-check-haveibeenpwned`.

The checking is made only if it has been 90 days since last check or if the last check had failed / found pwned password. If `wp-check-haveibeenpwned` found the hash, a warning page in the image is shown.

#### Screenshots

Screenshot from 2020-02-06 15-03-53

### DIFF
--- a/modules/passwords.php
+++ b/modules/passwords.php
@@ -2,7 +2,7 @@
 /*
  * Plugin name: Passwords
  * Description: Enforce strong passwords
- * Version: 1.0
+ * Version: 1.1
  */
 
 namespace Seravo;
@@ -15,6 +15,8 @@ if ( ! defined('ABSPATH') ) {
 if ( ! class_exists('Passwords') ) {
   class Passwords {
 
+    private static $password_hash;
+
     /**
      * Load passwords features
      */
@@ -22,6 +24,10 @@ if ( ! class_exists('Passwords') ) {
 
       add_action('login_enqueue_scripts', array( __CLASS__, 'register_scripts' ));
       add_action('admin_enqueue_scripts', array( __CLASS__, 'register_scripts' ));
+
+      add_filter('wp_authenticate_user', array( __CLASS__, 'calculate_password_hash' ), 10, 3);
+      // Use 15 as priority so wp-login.log (priority 10) is filled first
+      add_filter('login_redirect', array( __CLASS__, 'search_password_database' ), 15, 3);
 
     }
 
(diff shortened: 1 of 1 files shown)


ottok:
Wow, that was fast! This is a good start. We need to think a bit of the overall user experience, what messages are we sending, what is the likely action, how do we emphasize that this check was a good thing without causing too much panic, maybe link to some help article (but also considering whitelabel) and so on. That discussion is easy to have now when there is a technical proposal visible.

JoosuaKoskinen:
I made the page more informative and less scary. Still don't know how to take white-labeling to account with this.

Screenshot from 2020-02-06 16-46-41

The links are to https://help.seravo.com/article/29-managing-user-passwords-in-wordpress and https://seravo.fi/2014/password-hygiene-every-mans-responsibility.

This feature doesn't currently check if the user is a regular user or an admin. We probably shouldn't show all this information to a regular user.

Should we...
a) Use this feature only for admins?
b) Show a different message for normal users?
c) Hide all commands, Seravo links and other sensitive?

ottok:
> Should we...
> a) Use this feature only for admins?
> b) Show a different message for normal users?
> c) Hide all commands, Seravo links and other sensitive?

Let's make this check run only for admins and not at all for any other users. That way we avoid whitelabeling issues and protect the most valuable passwords.

modules/passwords.php
```diff
@@ -49,6 +55,73 @@ public static function register_scripts( $page ) {
...
+          <?php
+            wp_admin_css('install', true);
+            wp_admin_css('ie', true);
+          ?>
+        </head>
+        <body class="wp-core-ui">
+          <p id="logo"><a href="#"><?php _e('WordPress'); ?></a></p>
+          <h1><?php _e('Password change required', 'seravo'); ?></h1>
+          <p><?php printf('Automatic password security verification has found that your password has been pwned. Please follow the instructions below!', 'seravo'); ?></p>
+          <h3>What does this mean?</h3>
```
ottok: Some of the strings are not wrapped in gettext functions and thus not translatable.

JoosuaKoskinen: Fixed this

JoosuaKoskinen:
> Let's make this check run only for admins and not at all for any other users. That way we avoid whitelabeling issues and protect the most valuable passwords.

This feature is used for admins only now.

nuuttitoivola:
This was tested on several sites and no issues were found yet.

Issue #95, opened 2018-04-19
Deprecate HTTPS_DOMAIN_ALIAS and use DEFAULT_DOMAIN instead


ottok:
Related to ongoing transition this needs to start using DEFAULT_DOMAIN instead:
```
## seravo-plugin
modules/relative-urls.php:      if ( ! defined('HTTPS_DOMAIN_ALIAS_FRONTEND_URL') && defined('WP_CONTENT_URL') && substr(WP_CONTENT_URL, 0, 1) !== '/' ) {
modules/relative-urls.php:       * When using https-domain-alias the real siteurl is stored in HTTPS_DOMAIN_ALIAS_FRONTEND_URL
modules/relative-urls.php:      if ( defined('HTTPS_DOMAIN_ALIAS_FRONTEND_URL') ) {
modules/relative-urls.php:        $content = self::relativize_content_attributes( HTTPS_DOMAIN_ALIAS_FRONTEND_URL, $content );
modules/relative-urls.php:        $url = (defined('HTTPS_DOMAIN_ALIAS_FRONTEND_URL') ? HTTPS_DOMAIN_ALIAS_FRONTEND_URL : self::$siteurl );
modules/purge-cache.php:  $nginx_purge_endpoint = 'http://' . getenv('HTTPS_DOMAIN_ALIAS') . '/purge/';
```

Issue #11, opened 2015-09-21
Deprecate relative-urls from wp-palvelu-plugin (Make sure relative links are absolute if exported out from site)


ottok:
E.g. in http://foxland.fi/feed/ the image src attributes have relative url. That will not work when in a feed reader for example. When exporting $post->contents out from the site (feed, email) relative links need to be made absolute.

Check that implementation here and in https-domain-alias is in line with each other.

onnimonni:
Feed readers are already handled in: https://github.com/Seravo/wp-palvelu-plugin/blob/master/modules/relative-urls.php#L42-L44 this was done in commit: 66d6bd271a2513978398e2cb8d5a26140df89a49

ottok:
Cool! I guess it is then just a question of getting this into production. As a pilot, can you install this for Foxland and check is it solves customers issues?

onnimonni:
This solves feed reader but I'm not sure if it solves other issues like jetpack integration.

onnimonni:
I think otherwise this is working fine but when using xmlrcp or json-api we should make all of the links absolute.

onnimonni:
Only way to be absolutely about absolute urls sure would be to only use relative urls in database and then return them with regex in this hook:
```php
<?php add_filter( 'the_content', 'filter_function_name' ) ?>
```

Otherwise we can only achieve something like 95% of possible use cases.

On the otherhand this seems like a big overkill.

ottok:
Once everybody has https we should remove relativize_content_all and use absolute URLs as way too many WordPress plugins are incompatible with relative URLs (links in MailPoet, RSS feeds etc).

ottok:
Override filter available: 
`add_filter( 'wpp_use_relative_urls', '__return_false' ); // disable relative url's feature`

The file `modules/relative-urls.php` should perhaps not be part of this plugin, but instead a part of HTTPS Domain Alias plugin: https://github.com/Seravo/wp-https-domain-alias/issues/18

ottok:
HTTP Domain Alias is now officially deprecated from new installs. The relative-urls.php should be moved to the HTTPS Domain Alias as it originally was, since it is not needed elsewhere.

anttiviljami:
I don't agree we should move to remove relative-urls just yet. Relative URLs is used on sites without https domain alias too for development purposes. 

[...]

PR #342, opened 2020-03-08
Remove relative urls features long deprecated by wp-cli search-replace

ottok:
Closes: #11
Closes: #95


thejulle:
Hi, seravo-plugin causes a PHP notice when user adds an image to an editor field inside Gutenberg. The PHP notice is related to relative-urls.php (screenshot attached).

I then noticed this PR and can confirm that the bug disappeared with this version of seravo-plugin.

Screenshot 2020-04-02 at 9 39 46

PR #343, opened 2020-03-19
Temporarily disable real-time image optimizations

ottok:
The new version of wp-optimize-images is either slow or has some bugs
in exec() spawning or performance. Disable this for some time to get
until the new tool is optimized.

### DIFF
--- a/modules/optimize-on-upload.php
+++ b/modules/optimize-on-upload.php
@@ -45,7 +45,8 @@ if ( ! class_exists('OptimizeImagesOnUpload') ) {
      *
      */
     public static function optimize_images_on_upload( $filename ) {
-      exec('wp-optimize-images ' . $filename . ' &');
+      // @TODO: Enable this again once wp-optimize-images works wlll $arrayName = array('' => , );gain
+      // exec('wp-optimize-images ' . $filename . ' &');
       return $filename;
     }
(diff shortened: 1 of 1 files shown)


heikkiorsila:
A further optimization here would be to group images together in optimize_images_on_upload(), and process the group in one batch every x seconds (giving multiple individual files as argument for wp-optimize-images). This would amortize the cost of running wp-optimize-images.

ottok:
We don't know how many times that WordPress hook/action will be called, so it is hard to group calls to it, since we don't know when the "last" run is. We are quite limited on what we can do in interacting with WordPress core on this feature.

heikkiorsila:
I don't know WP architecture, but in general this is done as follows in Python:

def action(new_item, context):
    context.item_list.append(new_item)
    if context.timer is None:
        context.timer = create_timer(5, flush, context)  # 5 secs from now

def flush(context):
    flush_items(context.item_list)
    context.item_list.clear()
    context.timer = None

Does the PHP framework support timer callbacks and shared context?

ottok:
PHP workers have no shared context with other PHP workers AFAIK.

The optimization here is to give the max dimensions on the command line directly to avoid excess execs of wp-cli by Python.

PR #348, opened 2020-05-06
Use SeravoBot user agent when requesting jetpack ips

elguitar:
For whatever reason, jetpack.com responses randomly with 429,
when the user agent is WordPress/x.x.x.

Also, include https://seravo.com as url in the user agent, so
Jetpack knows who to contact in case of problems.

### DIFF
--- a/lib/security-restrictions.php
+++ b/lib/security-restrictions.php
@@ -129,7 +129,7 @@ if ( ! class_exists('Security_Restrictions') ) {
       if ( ($data === false) || count($data) < 1 ) {
         $whitelist = array();
         // Retrieve data from API
-        $response = wp_remote_get(esc_url_raw($url));
+        $response = wp_remote_get(esc_url_raw($url), array( 'user-agent' => 'Seravo/1.0; https://seravo.com' ));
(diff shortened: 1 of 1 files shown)


lib/security-restrictions.php
```diff
@@ -129,7 +129,7 @@ public function get_jetpack_whitelist() {
       if ( ($data === false) || count($data) < 1 ) {
         $whitelist = array();
         // Retrieve data from API
-        $response = wp_remote_get(esc_url_raw($url));
+        $response = wp_remote_get(esc_url_raw($url), array( 'user-agent' => 'SeravoBot/0.0.1; https://seravo.com' ));
```
ottok, line 132: Maybe just "Seravo/1.0"? This is the final version of this feature, so 1.0. And name could be simply "Seravo" or "SeravoPlugin" (this code is no bot).

elguitar: Good point! Fixed in https://github.com/Seravo/seravo-plugin/commit/133499052d2fb32da55d350b3a3cc82cbecc75e2

PR #349, opened 2020-05-08
Domains postbox renovation

JoosuaKoskinen:

- Adds support for 100% wide postboxes
- Makes the page look better
- Makes everything use AJAX
- Cleans up the codebase
- Improves error management
- Makes the page fully translatable
- Fixes time formats
- Makes all columns sortable
- Adds sniffing of customer managed dns
- Makes the changes made to the zone shown
- Adds ability to change the primary domain

This is not ready to merge yet. Note that this is just the domains postbox and the mails postbox PR will be separate. This PR will be updated and improved later.


ottok:
Yay!

Will Pete test and review this?

elguitar:
^ Changed the title, because @JoosuaKoskinen stated that this is not ready for merge.

JoosuaKoskinen:
Renamed as this is ready for merge after @elguitar does his review. Also SWD still has some features coming related to this and I think we should wait for them before merging.

lib/domains-ajax.php
```diff
@@ -42,6 +78,31 @@ function seravo_admin_change_zone_file() {
       return seravo_respond_error_json($response->get_error_message());
       wp_die();
     }
+
+    $response_decoded = json_decode($response, true);
+    if ( isset($response_decoded['diff']) && strlen($response_decoded['diff']) > 0 ) {
+      $records = Seravo_DNS_Table::fetch_dns_records('zone', $_REQUEST['domain']);
+      if ( ! isset($records['error']) ) {
+        $zone_diff = array();
+        $diff = explode("\n", $response_decoded['diff']);
```
elguitar, line 87: This whole diff-building could use a comment describing what it does in a whole.

lib/domain-tables.php
```diff
@@ -0,0 +1,552 @@
...
+        $actions['view'] = sprintf($action_request, 'sniff', __('View', 'seravo'));
+        $actions['edit'] = sprintf($action_disabled, __('DNS not managed by Seravo.', 'seravo'), __('Edit', 'seravo'));
+      } else {
+        $actions['view'] = sprintf($action_disabled, __('This domain doesn\'t have a zone.', 'seravo'), __('View', 'seravo'));
+        $actions['edit'] = sprintf($action_disabled, __('This domain doesn\'t have a zone.', 'seravo'), __('Edit', 'seravo'));
+      }
+    }
+
+    if ( empty($item['primary']) ) {
+      $actions['primary'] = sprintf($action_request, 'primary', __('Make Primary', 'seravo'));
```
elguitar, line 79: Maybe add `(experimental)` here? That way we can merge this even before the backend logic is clear.

We should then publish a new version as soon as possible. What do you think?

lib/domains-ajax.php
JoosuaKoskinen: I added a few comments

lib/domain-tables.php
JoosuaKoskinen: `(experimental)` has been added

PR #350, opened 2020-05-12
Fix the width of the HTTP request bars

elguitar:
Found on /wp-admin/tools.php?page=site_status_page

Previously, the width of the bar was always 'auto', because the ratio was
miscalculated.

Adding the `* 100` to the calculation fixed the issue, but it caused some of
the numbers to wrap. I added a second div, which prevents the wrapping.

Before:
Screenshot_20200512_112425

After:
Screenshot_20200512_111655


JoosuaKoskinen:
LGTM!

PR #359, opened 2020-06-12
Hide specified users from user lists

sjaks:
Hide Seravo-specific WordPress users from user listings by using the `pre_get_users` hook. Users `seravo` and `seravotest` will be hidden as defined in commit b57b33b. The ability to hide certain users from the front-end was discussed in the issue https://github.com/Seravo/wordpress/issues/46.

Works for the Users page in WordPress Admin and the CLI command `wp user list` but the effect should be tested for some user listing plugins. The hidden user can see itself on the lists if they are logged in.

### DIFF
--- a/modules/hide-users.php
+++ b/modules/hide-users.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Plugin name: WP Hide Users
+ * Description: Hides prespecified and given users from a WordPress page
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+if ( ! class_exists('HideUsers') ) {
+  class HideUsers {
+
+    public static function load() {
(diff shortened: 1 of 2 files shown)


modules/hide-users.php
```diff
@@ -0,0 +1,48 @@
...
+            unset($users_to_hide[$current_user_key]);
+        }
+
+        global $wpdb;
+        $array_sql_str = implode("', '", $users_to_hide);
+
+        // Edit SQL query clause just before making a query and skip the given users
+        $user_query->query_where = str_replace(
+          'WHERE 1=1',
+          "WHERE 1=1 AND {$wpdb->users}.user_login NOT IN ('{$array_sql_str }')",
```
elguitar, line 37: There should be an inline comment, which would describe the resulting WHERE clause through an example.

elguitar: Noice

sjaks:
No problem! I squashed the two remaining commits into one.

Issue #347, opened 2020-04-21
Admin-ajax calls return always 401 Unauthorized when not logged in


timiwahalahti:
**Describe the bug**
If a user is not logged in, plugin forces admin-ajax.php to return HTTP status `401 Unauthorized` in every request. This breaks, for example, Restrict Content Pro order form which makes an admin-ajax call and checks the status code of return.

Issue is caused by `change_http_code_to_unauthorized` function called in `wp_login_failed` hook [here](https://github.com/Seravo/seravo-plugin/blob/17f270b9d6cb289c279b6f794f378829bccf41e1/modules/fixes.php#L43). Because admin-ajax.php tries to check if user is logged in, `wp_login_failed` kicks in before executing possible nonpriv callback. When nonpriv callback does return, HTTP status code is already set to 401.

**Expected behavior**
`change_http_code_to_unauthorized` should not force HTTP status for ajax call. It should bail out if `DOING_AJAX` is defined.

PR #361, opened 2020-06-12
Don't force HTTP 401 for unauthorized AJAX calls

JoosuaKoskinen:

Stops forcing HTTP 401 for unauthorized AJAX calls. Some plugins require
unauthorized use of admin-ajax and they require a HTTP 200 response.

Issue: #347


JoosuaKoskinen:
There shouldn't be any. The HTTP 401 is meant for wp-login login attempts and AJAX just happens to call the same login function.

Issue #337, opened 2020-03-06
Improve UI for postbox content


pekkakortelainen:
Security and Database pages have issues with postbox content fitting well inside their respective containers. Version in use: 1.9.16 development (installed with wp-seravo-plugin-update --dev).

To reproduce:
- adjust the width of your browser window
- adjust your browser window zoom level

Security - Trash files
![seravo-plugin-trash-files-snap](https://user-images.githubusercontent.com/61181862/76079617-acb8e200-5fad-11ea-9fcc-bbae749ef02f.png)

Database - Search and replace results
![seravo-plugin-search-replace-results-snap](https://user-images.githubusercontent.com/61181862/76079925-72037980-5fae-11ea-8364-27457dbcb8f1.png)

JoosuaKoskinen:
I assume you mean `1.9.16` as `1.9.6` is quite outdated by now :) 
Managed reproduce the problem with the instructions given :+1:

Will mark as a bug as this shouldn't happen in the first place.

pekkakortelainen:
Thanks for the heads up, edited the version number.

Moppa5:
At least I wasn't very successful at reproducing this issue, is this still relevant issue? On local vagrant and on my own site no issue. Using Ubuntu 20.04 and Firefox. Could this be somehow browser / os related bug?

PR #362, opened 2020-06-12
Fix postbox overflow (Closes: #337)

JoosuaKoskinen:

Makes the cruft files and search-replace postboxes more compact and
allows some text wrapping to be done so the content doesn't overflow
on some viewport sizes.

Issue: #337

### DIFF
--- a/js/database.js
+++ b/js/database.js
@@ -27,9 +27,16 @@ jQuery(document).ready(function($) {
           // Loops through the row column by column
           jQuery.each(row.split('\t'), function (j, col) {
             if (i === 0) {
+              // Command row
               jQuery('#search_replace_command').append('<code>' + col + '</code>');
-            } else {
-              jQuery('<td>').html(col).appendTo(tr);
+            } else if (i === 1) {
+              // Title row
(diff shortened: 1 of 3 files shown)


js/database.js
```diff
@@ -20,16 +20,23 @@ jQuery(document).ready(function($) {
           jQuery('#' + section).html('No data returned for section.');
         }
         var data = JSON.parse(rawData);
-
+        console.log(data);
```
elguitar, line 23: I guess this was forgotten during developing.

js/database.js
```diff
@@ -20,16 +20,23 @@ jQuery(document).ready(function($) {
...
         // Loops through the data array row by row
         jQuery.each(data, function (i, row) {
           var tr = jQuery('<tr>');
           // Loops through the row column by column
           jQuery.each(row.split('\t'), function (j, col) {
             if (i === 0) {
+              // Command row
```
elguitar, line 30: Good comments :+1:

js/database.js
JoosuaKoskinen: Indeed and has now been removed.

Issue #339, opened 2020-03-06
Filter trash files


pekkakortelainen:
Directory **ai1wm-backups** is listed in trash files unnecessarily when it contains no backups. By default the directory has only the following small files (16Kb in total in my case): index.php and web.config. Even if the user deletes this folder, it reappears after the next page refresh.

Plugin version in use: 1.9.16 development (installed with wp-seravo-plugin-update --dev)

List of files to delete:
![seravo-plugin-cruft-files1](https://user-images.githubusercontent.com/61181862/76086080-058f7700-5fbc-11ea-8118-0f9beb467d17.png)

After deletion:
![seravo-plugin-cruft-files2](https://user-images.githubusercontent.com/61181862/76086330-91a19e80-5fbc-11ea-8ba0-8f4740b7d5ed.png)

After page refresh:
![seravo-plugin-cruft-files3](https://user-images.githubusercontent.com/61181862/76086385-a847f580-5fbc-11ea-8bf2-69843f30f293.png)

JoosuaKoskinen:
It seems to be created at https://github.com/d0n601/All-In-One-WP-Migration-With-Import/blob/a1c5ac76879c5a3d0afb3c1a4b150188f2f21264/lib/vendor/servmask/filesystem/class-ai1wm-directory.php#L35.

I don't remember it doing so before but I guess it did. The folder is currently listed as known cruft dir: https://github.com/Seravo/seravo-plugin/blob/master/lib/cruftfiles-ajax.php#L109.

I think we should do something about that. Maybe make the cruft remover only care about the files inside, not the folder itself.

Moppa5:
> I think we should do something about that. Maybe make the cruft remover only care about the files inside, not the folder itself.

This wouldn't help the case since it seems that removing the whole folder or the files inside will result the same outcome => files are created again and listed as cruft files. It seems that we need to create a some kind of filter for the files themselves or making an issue to wp migration plugin not to create the folder when it's not really required?

PR #363, opened 2020-06-12
Exclude automatically recreated dirs (Closes: #339)

JoosuaKoskinen:

Excludes all directories containing only whitelisted
dirs or files.

This allows excluding dirs which are automatically recreated
on page load by adding all the files in that dir in whitelisted_files
array.

Already added the dirs reported in #339.

Issue: #339

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -31,6 +31,19 @@ function find_cruft_dir( $name ) {
   return $dirs;
 }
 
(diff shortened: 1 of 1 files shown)


JoosuaKoskinen:
> `only_has_whitelisted_content` is ran inside two nested foreach. Is it certain, that there will not be any performance problems?

I'll look at it but I doubt it makes much of a difference. Cruft finding doesn't necessarily need to be that fast anyways.

Issue #329, opened 2020-02-25
Log interesting settings and actions in wp-settings.log, e.g. installs and activations of plugins


ottok:
We often have questions like "When was this plugin installed?" or "When was this theme activated and by who?". Currently we can in theory recover database backups and scan their wp_options contents to see what day what plugins and themes were active, but I suggest a better approach: make Seravo Plugin log the most interesting setting/options changes in `/data/log/wp-settings.log` (or `wp-options.log`) in real-time as they happen.

All lines in this log could include:
- timestamp
- username that made the change
- setting name
- old value (if easily available)
- new value

Alternatively the 4 last fields could be one longer human readable string, like "User joe activated plugin native-lazyload".

This log would be then displayed on the log page in Seravo Plugin for site admins to view.

There are plenty of "audit" plugins in the wp.org plugin repository, they can be used to check out what techniques are used to hook into relevant settings/option changes.

The PoC of this feature should log:
- plugin installs
- plugin uninstalls
- plugin activations
- plugin inactivations

Later more features can be added.

elguitar:
@Wiljami Is there a working branch for this? It is a good practice to add a link from the commits to the issue. For example: `Related issue: #329`. That way github adds automatic references here, so progress is easy to see.

Wiljami:
feature/plugin-log branch has a working version of this feature. Plugin and theme activations, deactivations and installations are logged to /data/log/wp-settings.log. Also plugin deletions are logged there.

There is no hook for theme deletion.

elguitar:
@Wiljami What's the status here?

Wiljami:
There's now logging for following things:
- Plugin installs and uninstalls
- Plugin activation and deactivation
- Theme install
- Theme switch

There still could be some 'interesting' things to log. Such as trying to figure out a good way to log theme deletion for example. Also theme and plugin upgrades probably could be logged as well.

ottok:
Upgrades be Seravo are already logged in update.log but I guess it does not
hurt if extra logging based on WP actions also store info about upgrades so
also manual updates are logged.

ottok:
Note widely used plugin that hooks into various actions of WP:
https://wordpress.org/plugins/wp-security-audit-log/

Maybe it can be used for some inspiration of what major WP
administration actions should be logged and what hooks it uses for it.

Wiljami:
> Such as trying to figure out a good way to log theme deletion for example. Also theme and plugin upgrades probably could be logged as well.

[...]

PR #366, opened 2020-06-17
Add plugin and theme logging

Wiljami:
Log plugin and theme activations, deactivations, installations, updates and deletions in /data/log/wp-settings.log file. Tracks date and time and the user who is making the change. 

Currently theme deletions are not logged as there is no hook for it.

Related issue: #329

### DIFF
--- a/modules/wp-plugin-log.php
+++ b/modules/wp-plugin-log.php
@@ -1,0 +1,101 @@
+<?php
+/**
+ * Plugin name: WP Plugin Log
+ * Description: Logs plugin and theme activations, deactivations, updates, installations
+ * and deletions. Theme deletion is not logged as there is no hook for it.
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+if ( ! class_exists('PluginLog') ) {
+  class PluginLog {
+
+    public static function load() {
+      add_action('activate_plugin', array( __CLASS__, 'on_try_activate_plugin' ), 10, 2);
+      add_action('activated_plugin', array( __CLASS__, 'on_activate_plugin' ), 10, 2);
+      add_action('deactivate_plugin', array( __CLASS__, 'on_try_deactivate_plugin' ), 10, 2);
+      add_action('deactivated_plugin', array( __CLASS__, 'on_deactivate_plugin' ), 10, 2);
+      add_action('switch_theme', array( __CLASS__, 'on_switch_theme' ), 10, 1);
+      add_action('upgrader_process_complete', array( __CLASS__, 'on_upgrader_process_complete' ), 10, 2);
+      add_action('delete_plugin', array( __CLASS__, 'on_delete_plugin' ));
+    }
+
+    public static function on_upgrader_process_complete( $upgrader = null, $arr_data = null ) {
+      // Check to make sure there's correct variables passed to this function
+      if ( empty($upgrader) || empty($arr_data) ) {
+        return;
+      } elseif ( $arr_data['type'] !== null && $arr_data['action'] !== null ) {
+        $type = $arr_data['type'];
+        $action = $arr_data['action'];
+      } else {
+        return;
+      }
+
+      if ( $type === 'theme' ) {
+        if ( $action === 'install' ) {
+          self::write_log('installed theme ' . $upgrader->theme_info());
+        } elseif ( $action === 'update' ) {
+          self::write_log('updated theme ' . $upgrader->theme_info());
+        }
+      } elseif ( $type === 'plugin' ) {
+        if ( $action === 'install' ) {
+          self::write_log('installed plugin ' . $upgrader->plugin_info());
+        } elseif ( $action === 'update' ) {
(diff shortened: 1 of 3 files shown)


modules/wp-plugin-log.php
```diff
@@ -0,0 +1,102 @@
...
+      add_action('delete_plugin', array( __CLASS__, 'on_delete_plugin' ));
+    }
+
+    public static function on_upgrader_process_complete( $upgrader = null, $arr_data = null ) {
+      if ( empty($upgrader) || empty($arr_data) ) {
+          self::write_log('on_upgrader_process_complete empty($upgrader) || empty($arr_data))');
+        return;
+      }
+
+      if ( $arr_data['type'] === 'theme' ) {
```
elguitar, line 34: What if `$arr_data['type']` or `$arr_data['action']` does not exist?

modules/wp-plugin-log.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+      }
+    }
+
+    public static function on_delete_plugin( $plugin_name = '' ) {
+      self::write_log('deleted plugin ' . $plugin_name);
+    }
+
+    public static function on_try_activate_plugin( $plugin, $network_activation ) {
+      self::write_log('tries to activate plugin ' . $plugin);
```
elguitar, line 62: I think `is trying to activate plugin ` would be more suitable format for these try-things.

modules/wp-plugin-log.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public static function on_activate_plugin( $plugin, $network_activation ) {
+      self::write_log('activated plugin ' . $plugin);
+    }
+
+    public static function on_try_deactivate_plugin( $plugin, $network_activation ) {
+      self::write_log('tries to deactivate plugin ' . $plugin);
```
elguitar, line 70: See above

modules/wp-plugin-log.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public static function write_log( $message ) {
+      $log_directory = dirname(ini_get('error_log'));
+
+      if ( empty($log_directory) ) {
+        // If there is no log directory, just log one directory above WordPress
+        // and hope that directory is writeable but not accessible from the web
+        $log_directory = '..';
+      }
+      $time_local = gmdate('j/M/Y:H:i:s O');
```
elguitar, line 89: Why this time format? Could the same be used as in `php-error.log`?
It's like this: `[16-Jun-2020 23:00:31 UTC] somethings error stuffs`

modules/wp-plugin-log.php
```diff
@@ -0,0 +1,102 @@
...
+          return;
+        }
+        if ( $arr_data['action'] === 'update' ) {
+          self::write_log('updated plugin ' . $upgrader->plugin_info());
+          return;
+        }
+      }
+    }
+
+    public static function on_delete_plugin( $plugin_name = '' ) {
```
elguitar, line 57: In the other functions, the parameter is called `$plugin`. Is this different in some way? If not, it would be cool, if they would be consistent.

elguitar:
Hello again!

I noticed, that in #329 it is specified that this log would be shown on "the log page of seravo-plugin".

Would you implement that too? I believe, it won't be a large task.

Please don't hesitate to ask, if you have any questions :)

Wiljami:
As it is seravo-plugin already displays the log in the log page as it parses through the folder.

I did however add wp-settings.log file to the array of log files in logs.php, so that it operates the same way as the rest of the logs

modules/wp-plugin-log.php
Wiljami: PHP_CodeSniffer throws error on date() and suggests to use gmdate() instead.

elguitar:
Oh great!

Please request a re-review when the comments are resolved :)

modules/wp-plugin-log.php
Wiljami: Changing to match the rest of the code.

modules/wp-plugin-log.php
Wiljami: I agree. Fixing it to match it.

modules/wp-plugin-log.php
Wiljami: I agree. Fixing it to match it.

modules/wp-plugin-log.php
Wiljami: Created checks for this occasion.

modules/wp-plugin-log.php
```diff
@@ -0,0 +1,102 @@
...
+        if ( $arr_data['action'] === 'install' ) {
+          self::write_log('installed plugin ' . $upgrader->plugin_info());
+          return;
+        }
+        if ( $arr_data['action'] === 'update' ) {
+          self::write_log('updated plugin ' . $upgrader->plugin_info());
+          return;
+        }
+      }
+    }
```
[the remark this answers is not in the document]
Wiljami: Refactored the code and made the if structure more appropriate.

The last major if block doesn't require else as the else would simply lead to do the same return anyway.

Added those two as variables.

PR #371, opened 2020-06-23
Include more evident cruft files in the cruft remover

sjaks:
Extend the cruft remover to detect clear backup and temp files as well as some well-known unnecessary files.

The cruft remover already had some common patterns for backup and temp files and this extends the list of them. OSX metadata file DS_Store should also be marked as cruft together with misplaced .htaccess and db.php files.

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -97,6 +97,9 @@ function seravo_ajax_list_cruft_files() {
         '*.wpress',
         'core',
         '*.bak',
+        '*deleteme*',
+        '*.deactivate',
+        '.DS_Store',
       );
       // List of known cruft directories
       $list_dirs = array(
@@ -105,6 +108,7 @@ function seravo_ajax_list_cruft_files() {
         'before*',
         'vanha',
         '*-old',
+        '*-OLD',
         '*-copy',
         '*-2',
         '*.bak',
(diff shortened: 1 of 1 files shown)


lib/cruftfiles-ajax.php
```diff
@@ -97,6 +97,10 @@ function seravo_ajax_list_cruft_files() {
         '*.wpress',
         'core',
         '*.bak',
+        '*.off',
```
elguitar, line 100: There no need to match `everything.off`. `wp-content/object-cache.php.off` is the one that should be matched exactly.
It is present on multiple sites, but it actually does nothing.

lib/cruftfiles-ajax.php
```diff
@@ -114,14 +119,21 @@ function seravo_ajax_list_cruft_files() {
         '*.orig',
         '-backup',
         '*.backup',
+        '*_bk_*',
```
elguitar, line 122: `*_bk_*` is too broad and it has a high risk of leading to false positives.

elguitar: ^ Therefore, I would suggest that it should not be tried to match

lib/cruftfiles-ajax.php
sjaks: Removed that entry and put `wp-content/object-cache.php.off` in the list of known files.

lib/cruftfiles-ajax.php
sjaks: Removed this :+1:

Issue #13, opened 2015-09-25
Allow customers to edit their email aliases


ottok:
Once #12 is completed and the secure Gandi API tool is implemented, extend it to alow customers to edit their email aliases.

ottok:
Status: The API has been implemented. Next step is to implement it on the Seravo Plugin dashboard side.

elguitar:
@JoosuaKoskinen What's the status here? Is there a WIP branch somewhere?

JoosuaKoskinen:
There's a PR now https://github.com/Seravo/seravo-plugin/pull/372

PR #372, opened 2020-06-23
Add mail forwards table

JoosuaKoskinen:

Adds mail forwards table to mails postbox on domains page.

The actions available are:
- Fetch forwards for a domain
- Create a new forward for a domain
- Delete a source and its destinations
- Modify the destinations of a source
- Modify the source

Closes: #13 

These actions must be tested with a domain under Seravo's management.

#### Screenshots
After fetch:
mails-postbox-1
Editing:
mails-postbox-2

### DIFF
--- a/js/domains.js
+++ b/js/domains.js
@@ -36,7 +36,7 @@ jQuery(document).ready(function($) {
 
     init: function () {
 
-      $(".sort-spinner, .action-spinner").remove();
+      $("#domains-table-spinner .sort-spinner, #domains-table-spinner .action-spinner").remove();
 
       $('#domains-table-wrapper .manage-column.sortable a, #domains-table-wrapper .manage-column.sorted a').click(function (e) {
         e.preventDefault();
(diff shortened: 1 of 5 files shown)


elguitar:
CI fails with PHP<=7.2. Weird :raised_eyebrow: 
```
PHP Parse error:  syntax error, unexpected ')' in ./lib/domain-tables.php on line 367
```

Is there some 7.4 specific syntax or something?

I will proceed with the review regardless of this :)

JoosuaKoskinen:
> Is there some 7.4 specific syntax or something?
There indeed was

JoosuaKoskinen:
> When modifying, If I press `return`, it doesn't change to a new line. Why is that? However, it does work and the end result is parsed correctly.

I can't reproduce this. All the empty lines are removed when submitting the form. Maybe it submits the form somehow when you press `return`.

> When I save the forwards for a source, I think that it should show me all the forwards instead of staying in the empty "create" state.

That's because a lot of different sources are often needed. This way saving them is a lot faster. Do you think that should be changed?

elguitar:
I PM'd you a video clip about what I ment with the return thing.

>That's because a lot of different sources are often needed. This way saving them is a lot faster. Do you think that should be changed?

Your reasoning seems valid. Let's leave it be and wait for feedback.

JoosuaKoskinen:
The return thing only happened on Firefox. Fixed now!

JoosuaKoskinen:
Found a couple of bugs, will change back to WIP

JoosuaKoskinen:
All good now, can you review again @elguitar?

Issue #124, opened 2018-06-08
Relabel "Technical contacts" to "Contacts" and require at least one email to be present


l3ku:
As these contact emails may actually be used for other than technical purposes, we should relabel the field and move it away from the "Updates" page to a completely new context, e.g. a page for site plan details and contact information. Also, we should make it required in the UI for at least one contact email to be present.

elguitar:
There are three subtasks to this issue:
1. Relabel Technical contacts to Contacts
1. Implement form validation that there is more than zero contact
1. Move the contact field away from Updates box

@Moppa5 You can focus on the first two. The third one should be done as a part of https://github.com/Seravo/seravo-plugin/issues/233 and/or https://github.com/Seravo/seravo-plugin/issues/98

PR #373, opened 2020-06-24
Add empty check for contact emails (Closes: #124)

Moppa5:
Translations needs to be done

This PR adds a few bug fixes and enhancements for the technical contact email handling on Upkeep page. There must be at least one valid email address and user is notified by this if they're trying to remove all emails. Emails are now on also required to be unique so user can't add/spam multiple same emails. Email input value text is reset after adding a new email. 

Issue #124 and #374

### DIFF
--- a/modules/upkeep.php
+++ b/modules/upkeep.php
@@ -106,6 +106,7 @@ if ( ! class_exists('Upkeep') ) {
           'running_tests' => __('Running rspec tests...', 'seravo'),
           'ajaxurl'    => admin_url('admin-ajax.php'),
           'ajax_nonce' => wp_create_nonce('seravo_upkeep'),
+          'email_fail' => __('There must be at least one contact email', 'seravo'),
         );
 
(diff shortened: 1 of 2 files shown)


elguitar:
@Moppa5 has more stuff coming in. We might as well review them at the same time.

Moppa5:
I edited the PR since there's issue #374 added too.

modules/upkeep.php
```diff
@@ -569,9 +571,11 @@ public static function seravo_admin_toggle_seravo_updates() {
       if ( isset($_POST['technical_contacts']) ) {
         $validated_addresses = array();
 
+        // There must be at least one contact email
         if ( ! empty($_POST['technical_contacts']) ) {
 
-          $contact_addresses = explode(',', $_POST['technical_contacts']);
+          // Only unique emails are valid
+          $contact_addresses = explode(',', array_unique($_POST['technical_contacts']));
```
elguitar, line 578: This seems to be backwards. The $_POST['technical_contacts'] is a string, which should be exploded first

PR #375, opened 2020-06-24
Prevent reading huge log files

JoosuaKoskinen:

Prevents reading of log files over 256MB.

Multiple issues about wp-admin not working with huge log files.


JoosuaKoskinen:
> Is there an issue, which this should close?
There is, will take care of them

Issue #288, opened 2019-10-10
Shadows box not user friendly


ottok:
The new shadows box is not very user friendly. When opened by default it looks like this:
![image](https://user-images.githubusercontent.com/668724/66569530-700acb00-eb74-11e9-8c43-75619289fe6b.png)

- untranslated text
- no list of shadows visible
- empty placeholders with no value visible even though a shadow does exits

When opened it looks also unfriendly:
![image](https://user-images.githubusercontent.com/668724/66569589-8ca70300-eb74-11e9-8819-192f6998f911.png)

Why do we have a drop-down here, would not a list be much better?

When opened it looks very complex:
![image](https://user-images.githubusercontent.com/668724/66569638-a5171d80-eb74-11e9-9327-30dfe92d94b9.png)

Compare this to the old shadows list, which was clear and there was a simple button to reset the shadow (the nr 1 feature users actually want to find and use):
https://seravo.com/docs/images/shadows-list.png

I opened this as a feature request, since this needs some designer to do the UI work.

JoosuaKoskinen:
Related to https://github.com/Seravo/seravo-plugin/issues/277. I agree, it's clearly not the best design. The narrow postbox makes it difficult to fit a lot of information.

Note that the new design should be ready for `wp-shadow-pull` in addition of reset.

ottok:
@nuuttitoivola @elguitar @JoosuaKoskinen This issue needs an status update perhaps? Also this seems to be a duplicate of #277 to a large part.

Issue #277, opened 2019-08-24
Improve UI for shadows box


ottok:
The old shadows list was quite clear. If you had multiple shadows, all would be immediately visible in the list.
![Screenshot from 2019-08-24 18-56-14](https://user-images.githubusercontent.com/668724/63640509-841c6b00-c6a1-11e9-9a8c-9f7a0e0db53b.png)

The new shadows box feels quite cluttered. One needs to find the shadows from a drop down. The text alignment does not look good, and with the new search-replace integrated there is more features yes, but the look is a bit messy.
![Screenshot from 2019-08-24 18-57-55](https://user-images.githubusercontent.com/668724/63640523-b5953680-c6a1-11e9-8cf9-16fdcee67ce6.png)

The functionality itself is good, but some rearrangement and UI design would be needed here to make the user experience smooth.

JoosuaKoskinen:
It seems that there's some CSS problems in the screenshot you posted. But I agree, it could look better. Due to postbox working only with boxes with 25% of width, there's not much room for design. The old 100% table was bit of a hack which can't be used with the new Seravo-toolbox design. That's why the page now uses mails-page style dropdown. New ideas for shadows-page are desired.

Which site is that screenshot from and can you still reproduce the problems? 

![shadows](https://user-images.githubusercontent.com/42264063/63706009-16a24300-c837-11e9-92d5-c06fa9b769f9.png)

ottok:
I wrote now a help page about this since there was not docs on it so far: https://help.seravo.fi/article/358-varjosivusto

Still I think the UI is ugly. Also from a novice user point of view the translation is clumsy and some parts (such as the shadow comment) are not translated at all (and unfortunately cannot directly be, since it is a string from the info field of the shadow and not Seravo Plugin.

ottok:
@nuuttitoivola @elguitar @JoosuaKoskinen This issue needs an status update perhaps?

elguitar:
Closed via https://github.com/Seravo/seravo-plugin/commit/959083f9d47e63e4b732c5886eb8ad16f32368dd

PR #377, opened 2020-06-26
Redesign shadows box (Closes: #288 #277)

nellileinonen:
### What are the main changes in this PR?
The items in shadows box are rearranged and some of the instruction texts are updated.
The shadows box is now arranged in a way that the new feature wp-shadow-pull can be easily added.

Translations are not done.

Shadows box needed redesign to be more user friendly. 
Issues: #288 #277 

#### Manual testing steps?
1. Create a shadow (if you don't have one already)
2. Make changes to production site
3. Go to Tools --> Site Status --> Shadows
4. Select a shadow from drop-down
5. Read shadow information
6. Click "Resetting the shadow from production"
7. Test resetting the shadow from production

#### Screenshots
Screenshot from 2020-06-26 15-51-22
Screenshot from 2020-06-26 15-53-14
Screenshot from 2020-06-26 16-01-32

### DIFF
--- a/modules/sitestatus.php
+++ b/modules/sitestatus.php
@@ -300,8 +300,7 @@ if ( ! class_exists('Site_Status') ) {
       ?>
       <div class="seravo-section">
         <div style="padding: 0px 15px">
-          <p><?php _e('Allow easy access to site shadows. Resetting a shadow copies the state of the production site to the shadow. All files under /data/wordpress/ will be replaced and the production database imported. For more information, visit our  <a href="https://seravo.com/docs/deployment/shadows/" target="_BLANK">Developer documentation</a>.', 'seravo'); ?></p>
-          <hr>
+          <p><?php _e('Manage the site shadows. For more information, visit our <a href="https://seravo.com/docs/deployment/shadows/" target="_BLANK">Developer documentation</a>.', 'seravo'); ?></p>
         </div>
         <div style="padding: 5px 15px 0 15px">
           <?php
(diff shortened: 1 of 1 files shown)


modules/sitestatus.php
```diff
@@ -311,55 +310,52 @@ public static function seravo_shadows_postbox() {
...
+                    $shadow_list[$shadow]['domain'] = $domain['domain'];
+                  }
+                }
+              }
+              ?>
+            </select>
+            <div id="shadow-table">
               <?php
-              function add_shadow( $identifier, $ssh, $created, $domain, $hidden = true ) {
+              function add_shadow( $name, $identifier, $ssh, $created, $domain, $hidden = true ) {
```
elguitar, line 332: I would advice on keeping the `$identifier` as the first parameter, since it is the one parameter that can be truly used to identify a shadow from another.

Also, `$name` is counterintuitive, since the function is now called: `add_shadow($shadow_data['info'], $shadow_data['name'], ...)`. Maybe it should be "description" in both the parameter name and in the UI? :thinking:

modules/sitestatus.php
```diff
@@ -311,55 +310,52 @@ public static function seravo_shadows_postbox() {
...
+                  <tr><th><?php _e('Domain', 'seravo'); ?></th><td><?php echo (empty($domain) ? '-' : $domain); ?></td></tr>
+                  <tr><th><?php _e('Creation Date', 'seravo'); ?></th><td><?php echo $created; ?></td></tr>
+                  </tbody>
+                </table>
                 <?php
                 if ( ! empty($identifier) ) {
                   ?>
-                  <tr class="data-actions"><th><?php _e('Actions', 'seravo'); ?></th><td><a class="action-link closed" href=""><?php _e('Move Data', 'seravo'); ?><span></span></a></td></tr>
+                  <div data-shadow="<?php echo $identifier; ?>" class="<?php echo ($hidden ? 'shadow-hidden ' : ''); ?>shadow-row" style="padding: 10px 0px">
+                    <a class="data-actions action-link closed" href=""><?php _e('Resetting the shadow from production', 'seravo'); ?><span></span></a>
```
elguitar, line 347: I am unsure of this phrasing. I think that this kind of link should indicate some action. For example: `Reset the shadow from production`. However, since the link does not reset the shadow, that is not good here.

The link opens up the shadow resetting options. Could the text be: `Click to open shadow resetting options` or `Shadow resetting options` or something? What do you think?

modules/sitestatus.php
```diff
@@ -311,55 +310,52 @@ public static function seravo_shadows_postbox() {
...
                 }
                 ?>
-                </tbody>
                 <?php
               }
-              // One empty entry for 'select shadow' in dropdown
-              add_shadow('', '', '', ' ', false);
               foreach ( $shadow_list as $shadow_data ) {
-                add_shadow($shadow_data['name'], $shadow_data['ssh'], $shadow_data['created'], $shadow_data['domain']);
+                add_shadow($shadow_data['info'], $shadow_data['name'], $shadow_data['ssh'], $shadow_data['created'], $shadow_data['domain']);
```
elguitar, line 355: If the parameters are changed above, remember to change this too.

ottok:
> What do you think, could it be transformed into a "radioboxable" list? How much work it would require?

Most sites have only one shadow, and bigger sites only 1-3. Using a drop-down here does not make much sense, it only "hides" information behind an element where users need to discover it and make a selection.

The view should be opimized for the most common action, that is resetting the shadow. From a users point of view they should be able to just open this view, click on the shadow name and press 'reset to fresh copy of production'. The button to reset should not be hidden behind yet another selection, but visible as directly as possible.

However the search-and-replace thing is something customers rarely use, and it complicates the usual use case, so that is something what could be hidden under a "Advanced settings" click or something.

nellileinonen:
I created a new pull request with table design: #385

Issue #380, opened 2020-06-30
Database Size pie chart should be bigger


sjaks:
The Database Size pie chart graphic under the Database-page is really small and thus, hard to read. When there are lots of sectors, the chart and sector tooltips are quite unnecessary.

![Screenshot from 2020-06-30 16-19-02](https://user-images.githubusercontent.com/44066308/86130855-7881a800-baed-11ea-8e01-6fb4638911b5.png)

The chart should be enlarged to be something like below.
 
![Screenshot from 2020-06-30 16-20-48](https://user-images.githubusercontent.com/44066308/86130997-b7176280-baed-11ea-9693-90fdbdb5662b.png)

sjaks:
It seems that the chart resized in an odd way and it happens that in 1080p resolution, the chart is really small. Zooming the page a bit in either direction makes the chart larger.

This relates to the upstream issue https://github.com/chartjs/Chart.js/issues/5049, which is a dulplicate of the open issue https://github.com/chartjs/Chart.js/issues/3761.

Changing the size of the chart canvas element doesn't have any effect on the chart nor does disabling responsive sizing of the chart in `reports-chart.js` fix anything.

sjaks:
This will become outdated after [/feature/enhanced-disk-usage-view](https://github.com/Seravo/seravo-plugin/tree/feature/enhanced-disk-usage-view) is merged. Closing because the previous features won't be fixed.

sjaks:
This relates to https://github.com/Seravo/seravo-plugin/issues/397. The current implementation won't be altered, so closing this.

PR #383, opened 2020-07-02
Make Database pie chart a bit easier to read

sjaks:
Make the Database Size chart found under the Database page a bit larger especially on fullHD monitors since before it was really small.

This is a temporary measure to make the sizing even a bit better until a better solution is found.

Sizing not made perfect since there is issues in the upstream of chart.js. Relates: #380

Before:
Screenshot from 2020-07-02 10-38-41

After:
Screenshot from 2020-07-02 10-37-06

### DIFF
--- a/js/reports-chart.js
+++ b/js/reports-chart.js
@@ -11,6 +11,8 @@ function generateChart(JSONdata) {
   });
   // Generate chart
   var ctx = document.getElementById('pie_chart');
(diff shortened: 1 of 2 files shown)


elguitar:
Is this PR still valid or should it be closed?

sjaks:
Won't fix. Closing.

PR #385, opened 2020-07-03
Redesign shadows box to table (Closes: #288 #277)

nellileinonen:
The items in shadows box are rearranged to a table and some of the instruction texts are updated.

Translations are not done.

Shadows box needed redesign to be more user friendly.
Issues: #288 #277

#### Manual testing steps?

1. Create a shadow (if you don't have one already)
2. Make changes to production site
3. Go to Tools --> Site Status --> Shadows
4. Select a shadow from table
5. Read shadow information
6. Click "Reset"
7. See that the shadow was reset from production

#### Screenshots
Screenshot from 2020-07-03 13-19-28
Screenshot from 2020-07-03 13-20-37
Screenshot from 2020-07-03 13-21-06
Screenshot from 2020-07-03 13-21-55
Screenshot from 2020-07-03 13-22-53

### DIFF
--- a/modules/sitestatus.php
+++ b/modules/sitestatus.php
@@ -300,8 +300,8 @@ if ( ! class_exists('Site_Status') ) {
       ?>
       <div class="seravo-section">
         <div style="padding: 0px 15px">
-          <p><?php _e('Allow easy access to site shadows. Resetting a shadow copies the state of the production site to the shadow. All files under /data/wordpress/ will be replaced and the production database imported. For more information, visit our  <a href="https://seravo.com/docs/deployment/shadows/" target="_BLANK">Developer documentation</a>.', 'seravo'); ?></p>
-          <hr>
+          <p><?php _e('Manage the site shadows.', 'seravo'); ?></p>
+          <p><?php _e('<strong>Warning: </strong>Resetting a shadow copies the state of the production site to the shadow. All files under /data/wordpress/ will be replaced and the production database imported. For more information, visit our  <a href="https://seravo.com/docs/deployment/shadows/" target="_BLANK">Developer documentation</a>.', 'seravo'); ?></p>
         </div>
         <div style="padding: 5px 15px 0 15px">
           <?php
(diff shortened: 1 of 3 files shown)


modules/sitestatus.php
```diff
@@ -314,50 +314,56 @@ public static function seravo_shadows_postbox() {
...
+                    $primary_domain = $domain['domain'];
+                  }
                 }
-                ?>
+              ?>
+                <!-- Two rows per shadow: by default, one visibe and another hidden -->
+                <tbody id="<?php echo ($shadow_data['name']); ?>">
+                  <tr class="view" >
+                    <td class="open-folded"><?php echo $shadow_data['name']; ?></td>
+                    <td class="open-folded"><button class="button reset" id="shadow-reset-status-new"><?php _e('Reset', 'seravo'); ?></button></td>
```
elguitar, line 332: Is there a reason for the `-new` postfix in the id?

nellileinonen: Actually no, there isn't. I'll change that

nellileinonen: Changed

elguitar:
I tested this. Seems to work out nicely!

However, I think that pressing `Reset` should not trigger the folding action.
Peek 2020-07-07 17-16

elguitar:
Please rebase on master to start the Travis CI.

nellileinonen:
> I tested this. Seems to work out nicely!
> 
> However, I think that pressing `Reset` should not trigger the folding action.
> Peek 2020-07-07 17-16

I agree, the 'Reset' should not trigger the folding. However, I added that feature because if the shadow has a domain, it is crucial that the user sees the instructions to execute search-replace after resetting. Now this information is only in the folded section. The best solution would be to show an alert to the user after pressing 'Reset'. The alert should have the instructions for search-replace if it is needed. Should I add this alert feature to this pull request or should I make a new issue about this usability problem?

elguitar:
Maybe it should be ensured, that after pressing the "Reset" button, the
fold is left unfolded (open). If it is already open, do nothing. If it is
folded, open it. What do you think?

ke 8. heinäk. 2020 klo 16.13 Nelli Leinonen (notifications@github.com)
kirjoitti:

nellileinonen:
> Maybe it should be ensured, that after pressing the "Reset" button, the fold is left unfolded (open). If it is already open, do nothing. If it is folded, open it. What do you think?

I agree. It's now fixed to work like this.

All in all, I think this is just a quick fix and third best solution to a bigger problem that the search-replace needs to be done manually by user after resetting the shadow. The best solution would be that the search-replace was done automatically on reset if needed and the user would not need to do it by themselves manually at all. The second best solution would be to show a separate success/failure message after the "Reset"-button press in which there would be the instructions to do the search-replace

nellileinonen:
Now the usability is better as there is a separate success/failure message after reset.

elguitar:
Looks good!

Although, I had some problems with my personal site:
kuva

elguitar:
^ Weird, it seems that all of that is shown, if the Notice is visible.

nellileinonen:
> Looks good!
> 
> Although, I had some problems with my personal site:
> kuva

Fixed. Thanks for noticing the bug!

nellileinonen:
> Very good!
> One thing that came to my mind:
> When the user resets a shadow and there is a custom domain for the shadow, a message is shown:
> 
> ```
> Success!
> 
> This shadow uses a custom domain. For the shadow to be accessible after reset, please run search-replace in the shadow with the values below:
> 
> From: ://ptrmts.fi
> To: ://seravo.example
> 
> You can run search-replace either on "Tools --> Database --> Search-Replace Tool" or with wp-cli. Instructions can be found from documentation.
> ```
> 
> There is no indication that the search-replace should be done in the shadow environment. If the user follows the quide without understanding, it could be possible that the production could be compromised.

The instruction text says: "...please run search-replace in the shadow with the values below..."
But clearly that "in the shadow" part needs to be emphasized more. I will write the instructions better

Issue #386, opened 2020-07-08
Buttons to external resources have misaligned icons


sjaks:
Like below, all this kind of buttons have misaligned icons. The bottom most one is a mockup of a fixed one.

![Screenshot from 2020-07-08 14-04-13](https://user-images.githubusercontent.com/44066308/86911562-33302c80-c124-11ea-8a49-f1397f4f58dd.png)

PR #387, opened 2020-07-08
Fix dashicon align inside buttons (Closes: #386)

sjaks:
Set the vertical align of dashicon icons to middle so that they are inline with button text.

Now all buttons with icons are like the one at the bottom. Other ones are examples of the previous look.
Screenshot from 2020-07-08 14-04-13

### DIFF
--- a/js/sitestatus.js
+++ b/js/sitestatus.js
@@ -52,7 +52,7 @@ jQuery(document).ready(function($) {
                 month.date +
                 '.html" target="_blank" class="button hideMobile">' +
                 seravo_site_status_loc.view_report +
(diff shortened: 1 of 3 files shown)


elguitar:
Please rebase on master to start the Travis CI.

elguitar:
Looks good now!
kuva

When screen width is small, the icon still looks a bit wonky.
kuva

sjaks:
@elguitar okay, that was because the icon took its anchor from the button text. Had to adjust the line height instead. Works well now but is not so elegant.

elguitar:
LGTM! Please merge :)

PR #391, opened 2020-07-13
Enhance disk space postbox

sjaks:
Use apexcharts.js for disk space postbox, implement a chart for total used disk space and redesign the per-directory chart.

Per-directory chart is now a bar chart that shows the size of certain directories listed in `sitestatus-ajax.php`.
The following logic is also followed:

- any directory under `/data` is shown
- all directories with size less than 1MB is not shown
- size of Seravo backups is not counted in the total used space nor is the directory listed in the bar chart

The donut chart for total disk use provides a simple view on how much of available space is used and the bar chart
provides a quick way to see what parts of the site take a lot of space.

#### Main view
Screenshot from 2020-07-13 11-39-42

#### When disk space is low
Screenshot from 2020-07-13 11-37-48

### DIFF
--- a/js/reports-chart.js
+++ b/js/reports-chart.js
@@ -80,3 +80,112 @@ function generateChart(JSONdata) {
     }
   });
 }
+
+function generateDiskDonut(gauge, maximum) {
+  var available = Math.round(maximum - gauge);
+  var colors = ['#ef7c1a', '#47aedc'];
+
+  if (available < 0) {
+    var available = 0;
+    var colors = ['#cc0000'];
+    jQuery('#disk_use_notification').css("display", "block");
+  }
+
+  // Generate donut chart
+  var options = {
+    series: [Math.round(gauge), available],
+    colors: colors,
+    chart: {
+      type: 'donut',
+      height: 120,
+    },
+    labels: ['Used', 'Available'],
+    responsive: [{
+      breakpoint: 480
+    }],
+    legend: {
+      show: false
+    },
+    dataLabels: {
+      enabled: false,
+    }
+};
+
+  var chart = new ApexCharts(document.querySelector("#donut_single"), options);
+  chart.render();
+}
+
+function generateDiskBars(JSONdata) {
+  var data = [];
+  var labels = [];
+  var human_vals = [];
+
+  var axis = {
+    categories: labels,
+    axisBorder: {
+      show: false
+    },
+    axisTicks: {
+      show: false
+    },
+    labels: {
+      show: false
+    }
+  };
+
+  Object.keys(JSONdata).forEach(function( folder ) {
+    data.push(JSONdata[folder].percentage);
+    labels.push(folder);
+    human_vals.push(JSONdata[folder].human);
+  });
+
+  var options = {
+    series: [{
+      data: data
+    }],
+    colors: ['#44A1CB'],
+    chart: {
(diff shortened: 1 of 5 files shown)


js/reports-chart.js
```diff
@@ -80,3 +80,112 @@ function generateChart(JSONdata) {
...
+    var colors = ['#cc0000'];
+    jQuery('#disk_use_notification').css("display", "block");
+  }
+
+  // Generate donut chart
+  var options = {
+    series: [Math.round(gauge), available],
+    colors: colors,
+    chart: {
+      type: 'donut',
```
elguitar, line 99: Why a donut chart instead of a pie?

js/sitestatus.js
```diff
@@ -74,8 +74,17 @@ jQuery(document).ready(function($) {
 
       if (section === 'folders_chart') {
         var allData = JSON.parse(rawData);
-        jQuery('#total_disk_usage').text(allData.data.human);
-        generateChart(allData.dataFolders);
+        var used_disk = (allData.data.size - allData.dataFolders['/data/backups/'].size) / 1000000;
+        var max_disk = $("#maximum_disk_space").html() * 1000;
+
+        allData.dataFolders = Object.fromEntries(
```
elguitar, line 80: This block seems confusing. An inline comment should be added to make it more "skimmable".

modules/sitestatus.php
```diff
@@ -220,18 +220,49 @@ public static function seravo_cache_status() {
...
+        <div id="donut_single" style="width: 30%; float: right"></div>
+        <div class="disk_usage_desc">
+          <?php _e('Disk space in your plan: ', 'seravo'); ?>
+          <span id="maximum_disk_space"><?php echo $max_disk; ?></span>GB
+          <br>
+          <?php _e('Space in use: ', 'seravo'); ?>
+          <span id="total_disk_usage"></span>
+          <br>
+          <span id="disk_use_notification" style="display: none">
+            <?php _e('Disk space low!  ', 'seravo'); ?>
```
elguitar, line 240: Is this double space intentional?

modules/sitestatus.php
```diff
@@ -220,18 +220,49 @@ public static function seravo_cache_status() {
...
+            <a href='https://help.seravo.com/article/280-seravo-plugin-site-status#diskusage' target='_BLANK'>
+              <?php _e('Read more.  ', 'seravo'); ?>
```
elguitar, line 242: Is this double space intentional?

js/reports-chart.js
sjaks: The donut chart was chosen because it seems clearer.

> Pie charts actually offer three separate options. You compare the ratio of the angles, the ratio of the areas of each wedge or slice, or last you can use the length of the outer circumference. The doughnut chart only offers the latter --

[Source](https://charting-ahead.corsairs.network/doughnuts-vs-pie-a-visual-primer-629b846aac67)

modules/sitestatus.php
sjaks: Fixed

modules/sitestatus.php
sjaks: Fixed

js/sitestatus.js
sjaks: Added comments

PR #392, opened 2020-07-13
GUI for wp-backup-list-changes-since

Moppa5:
Adding GUI for wp-backup-list-changes-since command and thus providing customers an easy way to see file / folder modifications between backups. We've as well been discussing about bringing the backup restoring feature so this could potentially be one step towards that. Users can track down what have changed between backups and possibly use this to choose which backup they should restore. 

Translations are yet to be done. 

We've been discussing and planning this feature during the sprint. 

#### Screenshots
Here's the basic GUI
Screenshot

After the run results / affected rows are shown in the toggle box. OBS, the rows affected shows now 5 rows since I modified HTML so this is no real bug. 
Screenshot(1)

### DIFF
--- a/js/upkeep.js
+++ b/js/upkeep.js
@@ -112,20 +112,13 @@ jQuery(document).ready(function($) {
   });
 
   jQuery(window).load(function(){
-    jQuery('.ba-slider').each(function(){
-      var cur = jQuery(this);
-      // Adjust the slider
-      var width = cur.width() + 'px';
-      cur.find('.ba-resize img').css('width', width);
-      // Bind dragging events
-      drags(cur.find('.ba-handle'), cur.find('.ba-resize'), cur);
-    });
-  
-    function seravo_load_change_report() {
+
+    function seravo_load_change_report(given_date) {
       jQuery.post(
         seravo_upkeep_loc.ajaxurl,
         { 'action': 'seravo_ajax_upkeep',
           'section': 'seravo_changes',
+          'date': given_date,
           'nonce': seravo_upkeep_loc.ajax_nonce },
         function(rawData) {
           if ( rawData.length == 0 ) {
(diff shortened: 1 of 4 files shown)


js/upkeep.js
```diff
@@ -111,6 +111,90 @@ jQuery(document).ready(function($) {
...
 
+  jQuery(window).load(function(){
+    jQuery('.ba-slider').each(function(){
+      var cur = jQuery(this);
+      // Adjust the slider
+      var width = cur.width() + 'px';
+      cur.find('.ba-resize img').css('width', width);
+      // Bind dragging events
+      drags(cur.find('.ba-handle'), cur.find('.ba-resize'), cur);
+    });
```
elguitar, line 122: I believe this has been pasted here by accident. Should be removed.

js/upkeep.js
```diff
@@ -111,6 +111,90 @@ jQuery(document).ready(function($) {
...
+            jQuery('#seravo_changes').html(seravo_upkeep_loc.no_data);
+          }
+          jQuery('#seravo_changes_status').fadeOut('slow', function() {
+            var data = JSON.parse(rawData);
+            var data_joined = data['output'].join("\n");
+
+            if ( data['rowCount'] === 0 ) {
+              // No rows affected
+              jQuery(this).html(seravo_upkeep_loc.no_affected_rows).fadeIn('slow');
+              jQuery('.seravo-test-result-wrapper').css('border-left', 'solid 0.5em #038103');
```
elguitar: The `changes_since` box should not use CSS classes, which are clearly made for the `wp-test` box.
Please rename.

js/upkeep.js
```diff
@@ -111,6 +111,90 @@ jQuery(document).ready(function($) {
...
+            } else {
+
+              jQuery(this).html(data['rowCount'] + " " + seravo_upkeep_loc.affected_rows).fadeIn('slow');
+              jQuery('.seravo-test-result-wrapper').css('border-left', 'solid 0.5em #e74c3c');
```
elguitar: The `changes_since` box should not use CSS classes, which are clearly made for the `wp-test` box.
Please rename.

js/upkeep.js
```diff
@@ -111,6 +111,90 @@ jQuery(document).ready(function($) {
...
+        }
+      ).fail(function() {
+        jQuery('#seravo_changes_status').html(seravo_upkeep_loc.run_fail);
+        jQuery('#run-changes-since').prop('disabled', false);
+      });
+    }
+  
+    jQuery('#run-changes-since').click(function() {
+      jQuery('#seravo_changes').html('');
+      jQuery('.seravo-test-result-wrapper').css('border-left', 'solid 0.5em #e8ba1b');
```
elguitar: The `changes_since` box should not use CSS classes, which are clearly made for the `wp-test` box.
Please rename.

PR #393, opened 2020-07-13
Add GUI for speed test

Wiljami:

Under Tools -> Site Status there is now block for Speed Test to run tests to different urls within the wordpress site. The test runs multiple curl commands from within the server to see the speed of generating the html from php. The tests are run with both with and without cache enabled.

The chart uses ApexCharts library, which is the same library as the one used in #391. 

There is button in the wordpress admin bar to test the speed of current page using the above mentioned GUI.

Before the speed test was only available from the CLI. With this the feature is more easily available to more people.

#### Screenshots
The result chart of a successful speed test:
pr 1

Button for the speed test in the admin bar:
pr 2

### DIFF
--- a/js/sitestatus.js
+++ b/js/sitestatus.js
@@ -62,6 +62,185 @@ jQuery(document).ready(function($) {
     );
   }
 
+  // Speed test data for the chart
+  let speedData = [];
+  // Speed test data for the chart
+  let speedDataCached = [];
+  // Visual chart of the speed test
+  let speedChart;
+  // Number of curl commands the speed test uses for both cached and non-cached
+  const speedNumberOfTests = 10;
+
+  $('#run-speed-test').click(run_speed_test);
+
+  function run_speed_test() {
+    $("#speed_test_url").attr("disabled", true);
+    $("#run-speed-test").prop('disabled', true);
+
+    // If an instance of the chart already exists, wipe it out
+    if (speedChart instanceof ApexCharts) {
+      speedChart.destroy();
+      speedData = [];
+      speedDataCached = [];
+    }
+
+    var options = {
+      series: [{
+              name: seravo_site_status_loc.latency,
+              data: speedData,
+          },
+          {
+              name: seravo_site_status_loc.cached_latency,
+              data: speedDataCached,
+          }
+      ],
+      chart: {
+          type: 'line',
+          height: 350,
+          zoom: {
+              enabled: false
+          }
+      },
+      plotOptions: {
+          bar: {
+              horizontal: false,
+              endingShape: 'flat',
+          },
+      },
+      dataLabels: {
+          enabled: false
+      },
+      stroke: {
+          show: true,
+          width: 2,
+          curve: 'smooth'
+      },
+      yaxis: {
+          title: {
+              text: 'ms'
+          },
+          min: 0
+      },
+      xaxis: {
+          min: 1,
+          max: speedNumberOfTests,
+          tickAmount: speedNumberOfTests,
+          tickPlacement: 'on'
+      },
+      fill: {
+          opacity: 1
+      },
+      tooltip: {
+          y: {
+              formatter: function(val) {
+                  return val + " ms"
+              }
+          }
+      },
+  };
+
+    speedChart = new ApexCharts(document.querySelector("#speed-test-results"), options);
+    speedChart.render();
+
+    let data = {
+      action: 'seravo_speed_test',
+      nonce: seravo_site_status_loc.ajax_nonce,
+      cached: false,
+    }
+
+    let loc = $("#speed_test_url").val();
+    if (loc.length != 0) {
+      data.location = loc;
+    }
+
+    // Number of tests done
+    let numberOfTest = 0;
+
+    // Sum of latencies for calculating average
+    let latencySum = 0;
+    // Sum of cached latencies for calculating average
+    let latencySumCached = 0;
+    // Result of a single speed test curl
+    let testResult = 0;
+    // Result of a single cached speed test curl
+    let cachedTestResult = 0;
+    speed_test();
+    // speed_test runs up to speedNumberOfTests of ajax calls for both cached
+    // and non-cached curl speed tests
+    function speed_test() {
+      $.post(seravo_site_status_loc.ajaxurl, data,
+        function(response) {
+          if (response.success) {
+            if (! data.cached) {
+              testResult = Math.round(response.data[4] * 1000);
+              latencySum += testResult;
+              speedData.push(testResult);
+            } else {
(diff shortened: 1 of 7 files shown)


js/sitestatus.js
```diff
@@ -62,6 +62,174 @@ jQuery(document).ready(function($) {
...
+        opacity: 1
+    },
+    tooltip: {
+        y: {
+          formatter: function (val) {
+            return val + " ms"
+          }
+          }
+    },
+    };
```
elguitar, line 135: The indentation with this structure is wonky. You could possibly use something like https://beautifier.io/ to easily reformat it.

js/sitestatus.js
```diff
@@ -62,6 +62,174 @@ jQuery(document).ready(function($) {
...
+
+    let numberOfTest = 0;
+
+    let latencySum = 0;
+    let latencySumCached = 0;
+    let testResult = 0;
+    let cachedTestResult = 0;
+    speed_test();
+
+    function speed_test() {
```
elguitar, line 159: Please include an inline comment which describes briefly what `speed_test()` does.
It also should mention, that it modifies the state of the variables above (`latencySum` etc.)

modules/sitestatus.php
```diff
@@ -129,9 +139,11 @@ public static function enqueue_site_status_scripts( $page ) {
       wp_register_script('chart-js', 'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.2/Chart.min.js', null, Helpers::seravo_plugin_version(), true);
       wp_register_script('seravo_site_status', plugin_dir_url(__DIR__) . '/js/sitestatus.js', '', Helpers::seravo_plugin_version());
       wp_register_style('seravo_site_status', plugin_dir_url(__DIR__) . '/style/sitestatus.css', '', Helpers::seravo_plugin_version());
+      wp_register_script('apex-charts', 'https://cdn.jsdelivr.net/npm/apexcharts', null, Helpers::seravo_plugin_version(), true);
```
elguitar, line 142: Note to @Wiljami and myself, these will collide with https://github.com/Seravo/seravo-plugin/pull/391/commits/063a8c98a3519a18f9ff8bbb007df6229d5841c4

modules/sitestatus.php
```diff
@@ -495,6 +507,39 @@ public static function optimize_images_postbox() {
       submit_button(__('Save', 'seravo'), 'primary', 'btnSubmit');
       echo '</form>';
     }
+
+    public static function speed_test() {
+      $target_location = isset($_GET['speed_test_target']) ? $_GET['speed_test_target'] : '';
+      echo('Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.<br>');
```
elguitar, line 513: Please use [_e()](https://developer.wordpress.org/reference/functions/_e/) to make this string translatable.

modules/sitestatus.php
```diff
@@ -495,6 +507,39 @@ public static function optimize_images_postbox() {
...
+      echo('<label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" id="speed_test_url" value="' . $target_location . '"><br>');
+      echo('<button type="button" class="button-primary" id="run-speed-test">Run Speed Test</button>');
```
elguitar, line 515: `Run Speed Test` should be made translatable. This can be achieved with [__()](https://developer.wordpress.org/reference/functions/__/) or [_e()](https://developer.wordpress.org/reference/functions/_e/)

lib/sitestatus-ajax.php
```diff
@@ -219,3 +219,48 @@ function seravo_ajax_site_status() {
   wp_die();
 
 }
+
+function seravo_speed_test() {
+  check_ajax_referer('seravo_site_status', 'nonce');
+  // Take location for the speed test from the ajax call. If there is not one, use WP home
+  $url = isset($_POST['location']) ? get_home_url() . '/' . trim($_POST['location']) : get_home_url();
```
elguitar, line 226: User gets to input the location here. Could the user inject something nasty here?
https://xkcd.com/327/

js/sitestatus.js
Wiljami: Fixed the extra error messages.

I ran it though the https://beautifier.io/ and unfortunately it didn't change it much. It's slightly better now at least.

modules/sitestatus.php
Wiljami: Changed to _e or __ for translations

modules/sitestatus.php
Wiljami: Changed to _e or __ for translations

lib/sitestatus-ajax.php
Wiljami: Using filter_var with FILTER_VALIDATE_URL to validate the url once the user input has been added to it.

Considering the scope, as a php's curl command, is this enough?

modules/sitestatus.php
Wiljami: With #391 merged with master. I'll start sorting it out.

lib/sitestatus-ajax.php
elguitar: I think it's enough.

modules/sitestatus.php
```diff
@@ -495,6 +511,39 @@ public static function optimize_images_postbox() {
...
+
+    public static function speed_toolbar_button( $admin_bar ) {
+      $target_location = ltrim($_SERVER['REQUEST_URI'], '/');
+
+      if ( substr($target_location, 0, 9) === 'wp-admin/' ) {
+        $admin_bar->add_menu(
+          array(
+            'id'    => 'seravo-speed-test',
+            'title' => '<span title="' .
+              sprintf(__('Test the speed of the current page. Note: This does not function within wp-admin folder!', 'seravo')) . '" class="ab-label">' . __('Speed Test', 'seravo') . '</span>',
```
elguitar, line 532: `wp-admin folder` is too technical. `in wp-admin` should be enough :)

Also, should there be an icon for this button? Lightning of course :zap:

Wiljami: Changed the admin folder.

Turns out adding an icon is becoming more of an headache than I anticipated. Still working on it.

Also WordPress Dashicons doesn't have lightning. We'll have to settle with a speedometer icon.

elguitar: Speedometer is fine too :)

Wiljami: As it turns out the way css, wordpress and the admin bar works out with each other, adding the icon turned out to be way bigger job than I expected. In any case it's done, but it lead to quite a few changes in the code, such as the admin bar part of the speed test being moved to its own file.

PR #394, opened 2020-07-13
Log important WordPress user changes

Wiljami:
Log following changes in users to /data/wp-user.log file:
 - User is created by another user
 - User registers
 - User's email is changed
 - User's password is changed
 - User is deleted
 - User's role is changed
 - User resets password

Related #329

### DIFF
--- a/modules/wp-user-log.php
+++ b/modules/wp-user-log.php
@@ -1,0 +1,140 @@
+<?php
+/**
+ * Plugin name: WP Plugin Log
+ * Description: Logs plugin and theme activations, deactivations, updates, installations
+ * and deletions. Theme deletion is not logged as there is no hook for it.
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+if ( ! class_exists('UserLog') ) {
+  class UserLog {
+
+    public static function load() {
+      add_action('edit_user_created_user', array( __CLASS__, 'on_edit_user_created_user' ), 10, 1);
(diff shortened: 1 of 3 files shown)


modules/wp-user-log.php
```diff
@@ -0,0 +1,140 @@
...
+      } else {
+        self::write_log_user('changed role of user (ID:' . $user_id . ') to ' . $role);
+      }
+    }
+
+    // Add current user info to the log entry
+    public static function write_log_user( $message ) {
+      $current_user = wp_get_current_user();
+      $user_id = $current_user->ID;
+      // eg ID is 0 when the change is made using WP CLI
```
elguitar, line 120: Nice! This could have been an easy thing to forget.

modules/wp-user-log.php
```diff
@@ -0,0 +1,140 @@
...
+        return;
+      }
+
+      $user = get_userdata($user_id);
+      // Check if the user content is reassinged to another user
+      if ( $redirect_user_id === null ) {
+        self::write_log_user('attempts to delete user ' . $user->user_login . ' (ID:' . $user_id . ')');
+      } else {
+        $redirect_user = get_userdata($redirect_user_id);
+        self::write_log_user('attempts to delete user ' . $user->user_login . ' (ID:' . $user_id . ') and reassigned content to user ' . $redirect_user->username . '(ID:' . $redirect_user_id . ')');
```
elguitar, line 60: Does not seem to output the redirect_user's userame:
`13/Jul/2020:13:24:23 +0000 User petri (ID:10) attempts to delete user testipetri (ID:14) and reassigned content to user (ID:10)`

Also if it did, there would not be a space after it.

Wiljami: Fixed!

PR #399, opened 2020-07-15
Adjust speed test button in adminbar

### LINKED ISSUE #398: Speed test panel button is confusing in WP-Admin
In WP-admin, clicking the speed test top panel button does nothing. This is confusing.

There's a tooltip, which gives a hint about what is happening, but it can easily go unnoticed.

There are three ways that this can be handled:
1. Remove the button, if the user is in wp-admin
1. Clearly react to the button press and inform the user, that the speed test should be run when not in wp-admin
1. Run the speedtest for home anyway

What do you think @Wiljami ?

Wiljami:
Speed test button is changed to lead to the site status page where the speed test is. Added an explanation text to the speed test block to say that the test can not be ran within wp-admin.

Closes #398

### DIFF
--- a/modules/sitestatus.php
+++ b/modules/sitestatus.php
@@ -566,13 +566,12 @@ if ( ! class_exists('Site_Status') ) {
 
     public static function speed_test() {
       $target_location = isset($_GET['speed_test_target']) ? $_GET['speed_test_target'] : '';
-      _e('Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.', 'seravo');
-      echo('<br><label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" id="speed_test_url" value="' . $target_location . '"><br>');
-      echo('<button type="button" class="button-primary" id="run-speed-test">' . __('Run Speed Test', 'seravo') . '</button>');
+      echo ('<p>' . __('Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.', 'seravo') . '</p>');
+      echo('<br><label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" placeholder="Front Page by Default" id="speed_test_url" value="' . $target_location . '"><br>');
+      echo('<button type="button" class="button-primary" id="run-speed-test">' . __('Run Test', 'seravo') . '</button>');
       echo('<div id="speed-test-results"></div>');
       echo('<div id="speed-test-error"></div>');
     }
-
 
   }
 
--- a/modules/speed-test.php
+++ b/modules/speed-test.php
@@ -22,22 +22,32 @@ if ( ! class_exists('Speed_Test') ) {
     // Add speed test button to WP Admin Bar
     public static function speed_test_button( $admin_bar ) {
       $target_location = ltrim($_SERVER['REQUEST_URI'], '/');
+      $url = get_home_url() . '/wp-admin/tools.php?page=site_status_page';
 
       if ( substr($target_location, 0, 9) === 'wp-admin/' ) {
         $admin_bar->add_menu(
           array(
-            'id'    => 'speed-test',
-            'title' => '<span class="ab-icon seravo-speed-test-icon"></span><span class="ab-label seravo-speed-test-text" title="' .
-              sprintf(__('Test the speed of the current page. Note: This does not function within wp-admin!', 'seravo')) . '">' . __('Speed Test', 'seravo') . '</span>',
+            'id'    => 'speed-test-blocked',
+            'title' => '<span class="ab-icon seravo-speed-test-icon blocked"></span><span class="ab-label seravo-speed-test-text blocked">' .
+              __('Speed Test', 'seravo') . '</span>',
+          )
+        );
+
+        $admin_bar->add_menu(
+          array(
+            'parent'  => 'speed-test-blocked',
+            'id'      => 'speed-test-menu',
+            'title'   => 'Speedtest cannot be run for wp-admin pages',
           )
         );
       } else {
-        $url = get_home_url() . '/wp-admin/tools.php?page=site_status_page&speed_test_target=' . $target_location;
+        $url .= '&speed_test_target=' . $target_location;
         $admin_bar->add_menu(
           array(
             'id'    => 'speed-test',
-            'title' => '<span class="ab-icon seravo-speed-test-icon"></span><span class="ab-label seravo-speed-test-text" title="' .
-            sprintf(__('Test the speed of the current page.', 'seravo')) . '">' . __('Speed Test', 'seravo') . '</span>',
+            'title' => '<span class="ab-icon seravo-speed-test-icon title="derp"></span><span class="ab-label seravo-speed-test-text" title="' .
+              __('Test the speed of the current page', 'seravo') . '">' .
+              __('Speed Test', 'seravo') . '</span>',
             'href'  => $url,
           )
         );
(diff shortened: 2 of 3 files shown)


elguitar:
Nice feedback from @Moppa5! :+1:

modules/speed-test.php
```diff
@@ -22,17 +22,18 @@ public static function load() {
...
     public static function speed_test_button( $admin_bar ) {
       $target_location = ltrim($_SERVER['REQUEST_URI'], '/');
+      $url = get_home_url() . '/wp-admin/tools.php?page=site_status_page';
 
       if ( substr($target_location, 0, 9) === 'wp-admin/' ) {
         $admin_bar->add_menu(
           array(
             'id'    => 'speed-test',
-            'title' => '<span class="ab-icon seravo-speed-test-icon"></span><span class="ab-label seravo-speed-test-text" title="' .
+            'title' => '<span class="ab-icon seravo-speed-test-icon blocked"></span><span class="ab-label seravo-speed-test-text blocked" title="' .
```
elguitar, line 31: The blocked icon does not provide the user any reason why it is blocked. I think that the block icon is not needed and it clutters user experience.

Instead, the button should somehow react to the button press. For example:
1. The button is pressed in wp-admin
1. The button flashes red and returns to normal. Tooltip opens up immediately with an explanation why speed test is not run.

elguitar: Or just not show the button in wp-admin :D

WP itself removes the `Customize` button when not on a page, so I don't see how this would be bad practice.

Wiljami: > Or just not show the button in wp-admin :D
> 
> WP itself removes the `Customize` button when not on a page, so I don't see how this would be bad practice.

I am starting to like this option a lot. I think we should go with this one!

Wiljami:
I changed the tooltip inside the dropdown menu. Now it's more snappy with its function compared to before.

modules/speed-test.php
```diff
@@ -22,25 +22,42 @@ public static function load() {
...
       $target_location = ltrim($_SERVER['REQUEST_URI'], '/');
+      $url = get_home_url() . '/wp-admin/tools.php?page=site_status_page';
 
       if ( substr($target_location, 0, 9) === 'wp-admin/' ) {
         $admin_bar->add_menu(
           array(
-            'id'    => 'speed-test',
-            'title' => '<span class="ab-icon seravo-speed-test-icon"></span><span class="ab-label seravo-speed-test-text" title="' .
-              sprintf(__('Test the speed of the current page. Note: This does not function within wp-admin!', 'seravo')) . '">' . __('Speed Test', 'seravo') . '</span>',
+            'id'    => 'speed-test-blocked',
```
elguitar, line 30: Are these block-icons necessary? They don't fit the overall appearance very well.

modules/sitestatus.php
```diff
@@ -566,14 +566,14 @@ public static function optimize_images_postbox() {
 
     public static function speed_test() {
       $target_location = isset($_GET['speed_test_target']) ? $_GET['speed_test_target'] : '';
-      _e('Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.', 'seravo');
-      echo('<br><label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" id="speed_test_url" value="' . $target_location . '"><br>');
-      echo('<button type="button" class="button-primary" id="run-speed-test">' . __('Run Speed Test', 'seravo') . '</button>');
+      _e('<p>Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.</p>', 'seravo');
+      _e('This test can not be run within wp-admin.', 'seravo');
```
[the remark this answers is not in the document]
Wiljami: Agreed

modules/speed-test.php
```diff
@@ -22,25 +22,42 @@ public static function load() {
...
+            'title' => '<span class="ab-icon seravo-speed-test-icon blocked"></span><span class="ab-label seravo-speed-test-text blocked">' .
+              __('Speed Test', 'seravo') . '</span>',
+          )
+        );
+
+        $admin_bar->add_menu(
+          array(
+            'parent'  => 'speed-test-blocked',
+            'id'      => 'speed-test-menu',
+            'title'   => 'This does not function within wp-admin',
```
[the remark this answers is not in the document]
Wiljami: Yes, changed to that version

modules/speed-test.php
Wiljami: You're right that mouse cursor icon is bit extreme for it. Changed the cursor to the default pointer. It is still different from the hand icon you get when hovering over links.

modules/speed-test.php
```diff
@@ -22,25 +22,42 @@ public static function load() {
...
           )
         );
+
+        $admin_bar->add_menu(
+          array(
+            'parent'  => 'speed-test',
+            'id'      => 'speed-test-menu',
+            'title'   => 'Test the speed of the current page',
+          )
+        );
```
elguitar, line 60: I would actually rather keep this in a tooltip. It's easier for the eyes, if this doesn't show up every time I want to click speed test.

asdf

modules/sitestatus.php
```diff
@@ -566,14 +566,13 @@ public static function optimize_images_postbox() {
 
     public static function speed_test() {
       $target_location = isset($_GET['speed_test_target']) ? $_GET['speed_test_target'] : '';
-      _e('Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.', 'seravo');
-      echo('<br><label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" id="speed_test_url" value="' . $target_location . '"><br>');
-      echo('<button type="button" class="button-primary" id="run-speed-test">' . __('Run Speed Test', 'seravo') . '</button>');
+      _e('<p>Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.</p>', 'seravo');
+      echo('<br><label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" placeholder="Front Page by Default" id="speed_test_url" value="' . $target_location . '"><br>');
+        echo('<button type="button" class="button-primary" id="run-speed-test">' . __('Run Test', 'seravo') . '</button>');
```
elguitar, line 571: Unneeded indentation

modules/sitestatus.php
```diff
@@ -566,14 +566,13 @@ public static function optimize_images_postbox() {
 
     public static function speed_test() {
       $target_location = isset($_GET['speed_test_target']) ? $_GET['speed_test_target'] : '';
-      _e('Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.', 'seravo');
-      echo('<br><label for="speed_test_url" class="speed_test_form" for="sr-from"> ' . get_home_url() . '/</label> <input class="speed_test_input" type="text" id="speed_test_url" value="' . $target_location . '"><br>');
-      echo('<button type="button" class="button-primary" id="run-speed-test">' . __('Run Speed Test', 'seravo') . '</button>');
+      _e('<p>Speed test measures the time how long it takes for PHP to produce the HTML output for the WordPress page.</p>', 'seravo');
```
elguitar, line 569: It is a good practice to keep HTML tags outside of gettext functions (`__()`, `_e()` etc.)

modules/speed-test.php
Wiljami: Ok!

modules/sitestatus.php
Wiljami: Funny how phpcs didn't pick that up

modules/sitestatus.php
Wiljami: Ok

PR #400, opened 2020-07-15
Do not limit the number of shown large dirs and fix styling

sjaks:
Before only 5 of the largest directories were included in the chart. Remove this limit because there is no reason to have it in place and users cannot see all the culprits for low disk space. Since now the number of directories can be really big, there is now a wrapper that makes the list scrollable if it becomes too large. The dir names can be really long, so they are now aligned so far to the left as possible. Because of the scroll bar, dir paths didn't always fit the screen very well and were cut off. Apex Charts tooltips have been enabled and configured now.

**Just a few dirs**
Screenshot from 2020-07-15 11-34-24

**Tons of dirs**
Screenshot from 2020-07-15 11-33-04

Test dirs were created with
```
$ wget https://speed.hetzner.de/100MB.bin
$ for i in {1..50}; do mkdir d$i ; done
$ for i in {1..50}; do cp 100MB.bin d$i/tiedosto ; done
```

### DIFF
--- a/js/reports-chart.js
+++ b/js/reports-chart.js
@@ -146,7 +146,7 @@ function generateDiskBars(JSONdata) {
     colors: ['#44A1CB'],
     chart: {
       type: 'bar',
-      height: 420,
+      height: labels.length * 40,
       toolbar: {
         show: false,
       }
@@ -154,6 +154,9 @@ function generateDiskBars(JSONdata) {
     plotOptions: {
       bar: {
         horizontal: true,
+        dataLabels: {
+          position: 'bottom',
+        },
       }
     },
     dataLabels: {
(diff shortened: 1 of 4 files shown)


Moppa5:
Nice! It's also more readable now since there's more padding between the directory sections.

sjaks:
> I noticed that the tooltips are inconsistent. Is this intentional?
That is the default behavior of Apex Charts. I will check if there is a way to spawn them in a consistent location.

sjaks:
@elguitar I created a custom tooltip element and overrode the apexcharts tooltip with it.

I took this opportunity to also make the chart a bit clearer:

- There is now some padding on the right so the bars do not touch the scrollbar
Screenshot from 2020-07-16 13-29-14

- There is a line under the chart
Screenshot from 2020-07-16 13-30-52

Issue #397, opened 2020-07-14
Database chart should use Apex Charts and be clearer


sjaks:
The charts introduced in https://github.com/Seravo/seravo-plugin/pull/393, https://github.com/Seravo/seravo-plugin/pull/392, https://github.com/Seravo/seravo-plugin/pull/391 and https://github.com/Seravo/seravo-plugin/pull/396 use Apex Charts as their chart library. There is now one chart on the Database page that still uses Chart.js.

The Database chart acts in an odd way because of an upstream bug, see issue https://github.com/Seravo/seravo-plugin/issues/380.

A pie chart is also a bad choice for such a chart, so maybe the type should also be changed. The Disk space chart was transformed into a bar chart and it seems quite clear. Maybe a bar chart could be used here.

Bar chart example:
![Screenshot from 2020-07-14 12-06-27](https://user-images.githubusercontent.com/44066308/87407179-7d088f00-c5ca-11ea-9f33-cc532a01374c.png)

Currently, the chart looks like this and is hard to read:
![](https://user-images.githubusercontent.com/44066308/86130997-b7176280-baed-11ea-9693-90fdbdb5662b.png)

Moppa5:
The proposed chart is likely to be at least more clearer when looking the database size as individual tables. Yet still I think that some kind of pie chart could be there as it's nice for visualizing too. Nevertheless, for consistency this proposed chart is much better :)

sjaks:
The bar chart can potentionally grow so large, it could mess up the Databse page. If it is used, the chart should be put into a wrapper of fixed height and set the overflow of the chart to be scrollable.

@Moppa5 I agree that it is difficult to see the relativity of the sizes here. What takes the highest percentage and so on. I am tyring to think if it's important to the user.

I think the important features of the chart are
- ability to see the largest tables
- ability to see the size of a specific table when debugging something

The bar chart example above already provides these features. Do you have ideas what a bar chart could provide?

(Or a donut chart: https://charting-ahead.corsairs.network/doughnuts-vs-pie-a-visual-primer-629b846aac67)

sjaks:
I was experimenting with a scroll bar. The scroll bar is essential if there are hundreds of tables. However, the default site has just a few and an installation with Woocommerce, also only a handful, so the height shouldn't ne fixed but a `max-height` should be set. The value could be automatically derived from the window height but this could require JS. A fixed max-height is enough, in my opinion.

[...]

PR #401, opened 2020-07-16
Use a bar chart as database chart (Closes: #397)

sjaks:
Change database chart library to Apex Charts.
Use a bar chart instead of a pie chart.

If there are many plugins, for instance, the number of database tables can be really big,
so there is now a wrapper that makes the list scrollable if the number becomes too large.

Database table names are also aligned as far left as Apex Charts allows because table names can be long and be cut off.
Because of possible long names and the scroll bar introduced in this commit, there are now Apex Charts tooltips that have
the same information as the bar themselves.

The bar chart is clean and I don't think there is need for other types of visual elements here.

The chart provides an easy way to see which tables are the largest and shows the exact size of all of them.

Screenshot from 2020-07-16 10-31-45

### DIFF
--- a/js/database.js
+++ b/js/database.js
@@ -45,7 +45,7 @@ jQuery(document).ready(function($) {
         jQuery('#sr-button').prop('disabled', false);
       }
     ).fail(function () {
-      jQuery('#' + section + '_loading').html('Failed to load. Please try again.');
+      jQuery('.' + section + '_loading').html('Failed to load. Please try again.');
     });
   }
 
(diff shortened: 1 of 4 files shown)


sjaks:
Made the changes also now present in https://github.com/Seravo/seravo-plugin/pull/400:

- Custom tooltip, that isn't wonky, implemented.

- There is now some padding on the right so the bars do not touch the scrollbar (screenshot from other chart but relevant)
Screenshot from 2020-07-16 13-29-14

PR #406, opened 2020-07-17
Fix and improve disk chart

sjaks:
The total disk space donut chart only had number values on the tooltips. Since they are always in MB, add the MB sign after the values.

Fixed the logic of calculating the used disk space and showing certain directories. Now directories are shown with the following logic:
- certain directories are always fetched
- directories with more than 100MB are always fetched

But there are directories the sizes of which are not counted in total usage:
- /data/backups
- /data/log
- /data/slog

Also, content of /data/backups are never shown even if their sizes exceed the 100M threshold.

This means the total disk usage is the size of /data minus the three dirs listed above.

### DIFF
--- a/js/sitestatus.js
+++ b/js/sitestatus.js
@@ -258,20 +258,15 @@ jQuery(document).ready(function($) {
 
       if (section === 'folders_chart') {
         var allData = JSON.parse(rawData);
-        // Calculate the used disk space; deduct the size of backups from the total
-        // space used. allData.data.size is in bytes, so divide into MB.
-        var used_disk = (allData.data.size - allData.dataFolders['/data/backups/'].size) / 1000000;
+        // Calculate the used disk space;
+        // allData.data.size is in bytes, so divide into MB
+        var used_disk = (allData.data.size) / 1000000;
 
         // Read plan's maximum disk space and transform into bytes
         var max_disk = $("#maximum_disk_space").html() * 1000;
 
-        allData.dataFolders = Object.fromEntries(
-          Object.entries(allData.dataFolders)
-            .filter(([key]) => ! key.startsWith('/data/backups/'))
-        )
-
         jQuery('#total_disk_usage').text(Math.round(used_disk) + 'MB');
-        generateDiskDonut(used_disk, max_disk);
+        generateDiskDonut(used_disk, max_disk, "MB");
         generateDiskBars(allData.dataFolders);
       } else if (section === 'front_cache_status') {
         var data = JSON.parse(rawData);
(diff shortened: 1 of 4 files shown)


js/sitestatus.js
```diff
@@ -260,18 +260,13 @@ jQuery(document).ready(function($) {
         var allData = JSON.parse(rawData);
         // Calculate the used disk space; deduct the size of backups from the total
```
elguitar, line 261: Is this comment still valid, since the dataFolders['/data/backups/'] is not substracted anymore?

lib/sitestatus-ajax.php
```diff
@@ -45,11 +45,30 @@ function seravo_ajax_report_http_requests() {
...
+    '--exclude=/data/slog',
+  );
+  // Directories not shown in the front-end even if their size
+  // exceed $dir_threshold
+  $hidden_dirs = implode(
+    '\|',
+    array(
+      '/data/backups',
+    )
+  );
```
elguitar, line 63: How should the end result here look? An example could be added to a comment.

lib/sitestatus-ajax.php
```diff
@@ -58,15 +77,13 @@ function seravo_report_folders() {
...
     du -sb /data/wordpress/htdocs/wordpress/wp-includes/ &&
     du -sb /data/wordpress/htdocs/wordpress/wp-admin/ &&
-    du -sb /data/backups/ &&
     du -sb /data/redis/ &&
-    du -sb /data/log/ &&
     du -sb /data/reports/ &&
-    du -sb /data/slog/ &&
     du -sb /data/db/
-    ) | sort -hr',
+    ) | grep -v "' . $hidden_dirs . '" | sort -hr',
```
elguitar, line 83: Frankly, I don't understand this. Should be documented somehow.

js/sitestatus.js
sjaks: Nice catch, removed!

lib/sitestatus-ajax.php
sjaks: Added a more explanatory comment and an output example

lib/sitestatus-ajax.php
sjaks: Added a comment that, with an example of `$hidden_dirs`, should be more understandable. The grep command here removes the lines with the items in the hidden_dirs string.

Feel free to give tips if there is something more that should be added!

elguitar: Ah! Now I got it! The quotes here can be really misleading. I thought that `git -v "' . 10000000 . '"` is run here.

sjaks:
Changes made. @elguitar do you think there is anything left to do?

PR #413, opened 2020-07-21
Change the PHP update instruction urls

### LINKED ISSUE #388: Replace or hide WordPress PHP nag
**Description**
WordPress nags about old PHP versions on the WP-admin Dashboard. The link to the WordPress' doc about PHP versions is not helpful at all for Seravo customers.

**To Reproduce**
1. Change PHP version to 7.0
1. Go to wp-admin front page

**Expected behavior**
Seravo should show a similar nag instead of the WordPress default one. It should have Seravo specific information on how to change the PHP version.

**Screenshots**
phpnag

Wiljami:
The PHP update instructions now point to the host's update instruction pages instead of default ones.

Change PHP update admin notice to take in account user locale for the instruction url.

Closes #388


Wiljami:
Had to update the translations as the url in admin notice was changed to a variable.

modules/check-php-version.php
```diff
@@ -43,8 +43,20 @@ public static function _seravo_check_php_version() {
 
     }
 
-    public static function _seravo_show_php_recommendation( $recommended_version ) {
+    public static function _seravo_update_php_url() {
+      return self::_seravo_find_update_url();
+    }
 
+    public static function _seravo_find_update_url() {
```
elguitar, line 50: To me, `find` refers to searching for a value from a data structure. What do you think, could there be a more suitable name for this function?

modules/check-php-version.php
```diff
@@ -53,9 +65,10 @@ public static function _seravo_show_php_recommendation( $recommended_version ) {
       // The line below is very long, but PHPCS standards requires translation
       // strings to be one one line
       printf(
-        // translators: %1$s: current php version, %2$s: recommended php version
-        __('PHP %s is available but not used on this site. Developers might want to <a href="tools.php?page=upkeep_page">upgrade the latest PHP version</a> for faster performance and new features. Read more about <a target="_blank" href="https://help.seravo.com/article/41-set-your-site-to-use-newest-php-version">PHP version upgrades</a>.', 'seravo'),
-        $recommended_version
+        // translators: %1$s: current php version
```
elguitar, line 68: This seems backwards. Shouldn't the comment about %2$s be here instead of the old version? :thinking:

modules/check-php-version.php
Wiljami: Renaming it

modules/check-php-version.php
```diff
@@ -43,8 +43,20 @@ public static function _seravo_check_php_version() {
 
     }
 
-    public static function _seravo_show_php_recommendation( $recommended_version ) {
+    public static function _seravo_update_php_url() {
+      return self::_seravo_find_update_url();
+    }
```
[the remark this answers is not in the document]
Wiljami: I changed the _seravo_show_php_recommendation function to use _seravo_find_update_url as well to localize it as well.

Now I however realize that I can just change the _seravo_update_php_url to just have the check and have the _seravo_show_php_recommendation function use it as well.

languages/seravo.pot
```diff
@@ -299,9 +299,9 @@ msgstr ""
 msgid "The HTTPS protocol is not currently active in the <a href=\"%s/wp-admin/options-general.php\">site settings</a>. Please <a href=\"https://help.seravo.com/article/24-how-do-i-enable-the-https-on-our-website\" target=\"_BLANK\">use HTTPS</a>."
 msgstr ""
 
-#. translators: %1$s: current php version, %2$s: recommended php version
-#: modules/check-php-version.php:57
-msgid "PHP %s is available but not used on this site. Developers might want to <a href=\"tools.php?page=upkeep_page\">upgrade the latest PHP version</a> for faster performance and new features. Read more about <a target=\"_blank\" href=\"https://help.seravo.com/article/41-set-your-site-to-use-newest-php-version\">PHP version upgrades</a>."
+#. translators: %1$s: current php version
```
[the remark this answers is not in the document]
Wiljami: Right! Updating the comment

modules/check-php-version.php
```diff
@@ -53,9 +61,10 @@ public static function _seravo_show_php_recommendation( $recommended_version ) {
       // The line below is very long, but PHPCS standards requires translation
       // strings to be one one line
       printf(
-        // translators: %1$s: current php version, %2$s: recommended php version
-        __('PHP %s is available but not used on this site. Developers might want to <a href="tools.php?page=upkeep_page">upgrade the latest PHP version</a> for faster performance and new features. Read more about <a target="_blank" href="https://help.seravo.com/article/41-set-your-site-to-use-newest-php-version">PHP version upgrades</a>.', 'seravo'),
-        $recommended_version
+        // translators: %1$s: recommended php version
```
elguitar, line 64: This should include the explanation for `%2$s` too.

modules/check-php-version.php
```diff
@@ -43,8 +43,16 @@ public static function _seravo_check_php_version() {
...
     }
 
-    public static function _seravo_show_php_recommendation( $recommended_version ) {
+    public static function _seravo_update_php_url() {
+      if ( get_user_locale() === 'fi' ) {
+        return 'https://help.seravo.fi/article/234-uusimpaan-php-versioon-siirtyminen';
+      } else {
+        return 'https://help.seravo.com/article/41-set-your-site-to-use-newest-php-version';
+      }
+    }
```
elguitar, line 52: Why localize using `if/else` structure? Could this be achieved with `__()`?

Wiljami: It did not occur to me to use the localization for this. I'll make this change.

PR #414, opened 2020-07-21
Remove special characters on upload

### LINKED ISSUE #71: Automatically rename uploads with special characters
Every now and then we come across sites that suffer from problems because they have ä and ö letters in their URLs.

To fix it at least for media uploads, we could finalize the [automatic renamer](https://github.com/Seravo/wp-rename-uploads) which would automatically transform all _ä_ and _ö_ into _a_ and _o_.

This could be enabled by default or as an opt-in feature of the Seravo Plugin.

Wiljami:
Replace characters like ä with a and ö with o when a file with such characters is uploaded.

By default this is off.

Add a toggle to opt-in for this feature in 'Site status'.
I think this should actually be under 'Site upkeep', but since 'Optimize images' is under 'Site status' I think this feature is better to be where ever it is.

Closes #71


Wiljami:
> Now, it goes the other way. If I change the setting on the file sanitizer, I get the "Settings saved" notification over the image optimizer.
> 
> Also, there is no confirmation of the change in the file sanitizer.

Right. The problem is with the way WP handles the settings fields. I am trying to use two different fields for these settings and they're not cooperating at all like I'd want them to. I can make the errors appear where they should, but then I can't make the confirmations work properly.

I think solution is to either code the toggle, button and notification separately for the file name sanitization or combine the two settings blocks to one settings block. eg. "File upload settings" or something like that.

elguitar:
I believe that [settings_errors()](https://developer.wordpress.org/reference/functions/settings_errors/) could be used with the `string $setting` parameter to limit the output to one setting per call.

Wiljami:
Yes. However the 'Settings saved' confirmation also makes use of it and giving it an id is not trivial.

Wiljami:
As it is now in master when you add a valid value to width and invalid value to height (or other way around) the block will print out an error message and no message of setting saved. However the valid value is saved anyway in this case.

Wiljami:
I'll change it to work as following so it's consistent and the settings messages work with it and the special characters sanitizing:
image optimizer issue
I think if it's functionality is to be changed, that's a separate issue.

ottok:
Should we maybe have type=number on the input element directly and let the
browser do the validation?

https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation

Wiljami:
> Should we maybe have type=number on the input element directly and let the browser do the validation? 

Yes, for sure. I did validation server side in #415 and I'll sort this there.

modules/sitestatus.php
```diff
@@ -501,17 +510,21 @@ public static function check_default_settings() {
...
+      if ( get_option('seravo-enable-sanitize-uploads') === false ) {
+        update_option('seravo-enable-sanitize-uploads', 'off');
+      }
+
     }
 
     public static function seravo_image_max_width_field() {
       $image_max_width = get_option('seravo-image-max-resolution-width');
-      echo '<input type="text" class="' . self::get_input_field_attributes()[0] . '" name="seravo-image-max-resolution-width"' . self::get_input_field_attributes()[1] . '
+      echo '<input type="number " class="' . self::get_input_field_attributes()[0] . '" name="seravo-image-max-resolution-width"' . self::get_input_field_attributes()[1] . '
```
elguitar, line 521: Unneeded space after 'number'

PR #417, opened 2020-07-22
Fix optimize images input validation

### LINKED ISSUE #415: Optimize images does not validate width and height properly
You can add text behind the width and height options in 'Optimze images' and it saves the values.

When you refresh the page, the values with letters behind them are still there.

optimize images

Wiljami:
It was possible to put values such as '100asd' as width and height.
Now the values are checked to make sure they're proper numbers.

Seems php has interesting ways of comparing values. 1000 < '2000asdf' returns true even when the other is a string and with extra non-number characters!

Closes #415

### DIFF
--- a/modules/sitestatus.php
+++ b/modules/sitestatus.php
@@ -526,31 +526,35 @@ if ( ! class_exists('Site_Status') ) {
     }
 
     public static function sanitize_image_width( $width ) {
-      if ( $width < self::$min_width && $width !== null && get_option('seravo-enable-optimize-images') === 'on' ) {
-        add_settings_error(
-          'seravo-image-max-resolution-width',
-          'invalid-width',
-          sprintf(
-            // translators: %s numeric value for the minimum image width
-            __('The minimum width for image optimisation is %1$s px. Setting suggested width of %2$s px.', 'seravo'),
-            self::$min_width,
-            self::$max_width_default
-          )
-        );
-        return self::$max_width_default;
+      if ( get_option('seravo-enable-optimize-images') === 'on' && $width !== null ) {
+        if ( ! is_numeric($width) || $width < self::$min_width ) {
+          add_settings_error(
+            'seravo-image-max-resolution-width',
+            'invalid-width',
+            sprintf(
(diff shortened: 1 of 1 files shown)


modules/sitestatus.php
```diff
@@ -526,31 +526,35 @@ public static function optimize_images_settings_description() {
...
-          'invalid-width',
-          sprintf(
-            // translators: %s numeric value for the minimum image width
-            __('The minimum width for image optimisation is %1$s px. Setting suggested width of %2$s px.', 'seravo'),
-            self::$min_width,
-            self::$max_width_default
-          )
-        );
-        return self::$max_width_default;
+      if ( get_option('seravo-enable-optimize-images') === 'on' && $width !== null ) {
```
Wiljami, line 529: I cut this if line to two to make it clearer what's going on.

modules/sitestatus.php
```diff
@@ -526,31 +526,35 @@ public static function optimize_images_settings_description() {
...
     public static function sanitize_image_height( $height ) {
-      if ( $height < self::$min_height && $height !== null && get_option('seravo-enable-optimize-images') === 'on' ) {
-        add_settings_error(
-          'seravo-image-max-resolution-height',
-          'invalid-height',
-          // translators: %s numeric value for the minimum image height
-          sprintf(__('The minimum height for image optimisation is %1$s px. Setting suggested height of %2$s px.', 'seravo'), self::$min_height, self::$max_height_default)
-        );
-        return self::$max_height_default;
+      if ( get_option('seravo-enable-optimize-images') === 'on' && $height !== null ) {
```
Wiljami, line 548: I cut this if line to two to make it clearer what's going on.

modules/sitestatus.php
```diff
@@ -262,7 +262,7 @@ public static function seravo_disk_usage() {
         <?php
           $api_response = API::get_site_data();
           if ( is_wp_error($api_response) ) {
-            die($api_response->get_error_message());
+          die($api_response->get_error_message());
```
elguitar, line 265: This indentation is wrong

PR #421, opened 2020-07-28
Fix admin panel freeze with too large php-error log

Wiljami:
Abnormally long lines in the log file cause the admin panel to freeze. 

Now if too many chunks are generated while reading the log file, give up reading the log file to prevent the admin panel from freezing and return false like with too large files.

This does not fix the underlying issue that there is such log files though, and is a band-aid to stop the admin panel from freezing.

### DIFF
--- a/modules/login-notification.php
+++ b/modules/login-notification.php
@@ -92,6 +92,20 @@ if ( ! class_exists('Login_Notifications') ) {
       // Read and reverse the php error logfile
       $output = Logs::read_log_lines_backwards('/data/log/php-error.log', -1, self::$max_rows);
       if ( ! $output ) {
+        // If output is false rather than an empty array then something went wrong with reading the file
+        if ( ! is_array($output) ) {
+          // Display admin dashboard error message
+          add_action(
+            'wp_dashboard_setup',
+            function() {
+              wp_add_dashboard_widget(
+                'seravo-error-widget',
+                __('Error log read failure', 'seravo'),
+                array( __CLASS__, 'display_admin_log_read_notification' )
+              );
+            }
+          );
+        }
         return 0;
       }
       $output_reversed = array_reverse($output);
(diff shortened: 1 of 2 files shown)


Wiljami:
Since it is related to this topic I believe adding the feedback on log size errors to the users is within the this PR. The second commit adds feedback to the dashboard and the logs portion.

Wiljami:
TODO: Translations once code is OK

modules/logs.php
```diff
@@ -353,6 +363,9 @@ public static function read_log_lines_backwards( $filepath, $offset = -1, $lines
       // the newline in the end accouts for an extra line
       $lines--;
 
+      // Set max amount chunks to be read according to to the lines wanted from the log.
```
elguitar, line 366: typo

modules/login-notification.php
```diff
@@ -192,6 +206,16 @@ public static function display_admin_errors_notification() {
       );
       echo '<div>' . $msg . '</div>';
     }
+    /**
+    * Display the amount of php-error.log lines that have appeared this week.
+    */
```
elguitar, line 211: This comment does not correspond to the code beneath

modules/logs.php
```diff
@@ -210,6 +210,11 @@ public function render_log_view( $logfile, $regex = null, $max_num_of_rows ) {
       </div>
       <?php elseif ( ! $result ) : ?>
         <p><?php _e('Log empty', 'seravo'); ?></p>
+        <?php
+        // result -1 is the signal that something went wrong with reading the log
+        elseif ( $result === -1 ) :
+        ?>
+        <p><?php _e('Log is abnormally large and can not be displayed.', 'seravo'); ?></p>
```
elguitar, line 217: Should this be indentated?

modules/logs.php
```diff
@@ -210,6 +210,11 @@ public function render_log_view( $logfile, $regex = null, $max_num_of_rows ) {
       </div>
       <?php elseif ( ! $result ) : ?>
         <p><?php _e('Log empty', 'seravo'); ?></p>
+        <?php
+        // result -1 is the signal that something went wrong with reading the log
+          elseif ( $result === -1 ) :
```
elguitar, line 215: This should not be indentated

Issue #315, opened 2020-01-20
Add handling of 504 AJAX-responses


JoosuaKoskinen:
Some AJAX-calls are quite likely to take over the timeout limit. Seravo Plugin should detect when so happens and instruct the user how to do the same with SSH.

Users might think (and have thought) there's something broken with their site or Seravo Plugin when the spinner keeps running infinitely.

One of calls that may do this is the wp-php-compatibility-check in php version changer.

JoosuaKoskinen:
Example: https://github.com/Seravo/seravo-plugin/issues/187

Wiljami:
Adding a .fail to the jquery posts ought to solve this issue. I created such to the upkeep page regarding to compatibility check in #423.

What other ajax calls are there in seravo-plugin that would also require this?

elguitar:
@Wiljami is there still something to do in this issue or can we close this?

Wiljami:
I think it's solved the best it could be. It informs the user if the response takes too long.

elguitar:
Closing in favor of #465

PR #423, opened 2020-07-29
Add error feedback on php-compatibility timeout

Wiljami:
Add feedback to the user if the php-compatibility ajax check takes too long.

Related #315

TODO: Translations once the code is ok.

### DIFF
--- a/js/upkeep.js
+++ b/js/upkeep.js
@@ -290,6 +290,17 @@ jQuery(document).ready(function($) {
           });
         }
       }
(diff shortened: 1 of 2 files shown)


js/upkeep.js
```diff
@@ -290,6 +290,18 @@ jQuery(document).ready(function($) {
           });
         }
       }
+      ).fail(function(msg) {
+        console.log(msg.status);
```
sjaks, line 294: No console.logs in production

Issue #187, opened 2018-10-29
Resetting a shadow causes HTTP 504 if the operation takes long


l3ku:
While successful, running `wp-shadow-reset` via SSH on our site takes somewhat long:
```
wp-palvelu@wp-palvelu:~$ time wp-shadow-reset wp-palvelu_xxxxx
Warning: the database and all files in /data/wordpress in shadow wp-palvelu_xxxxx will be deleted and replaced with a clone from production. Are you sure you want to proceed? (y/n): y
Exporting production database...
Moving database from production to wp-palvelu_xxxxx...
Moving files from production to wp-palvelu_xxxxx...
Importing production database to wp-palvelu_xxxxx...
Success: the shadow wp-palvelu_xxxxx has been updated with production.

real	1m10.517s
user	0m0.908s
sys	0m0.400s

(Shadow ID has been removed)
```
Presumably the large size of the database or some other phenomenon causes the AJAX request that is made when resetting the exact same shadow via WP admin to timeout with HTTP 504, taken from chrome dev console response/request headers:

```
Request URL: https://wp-palvelu.fi/wp-admin/admin-ajax.php
Request Method: POST
Status Code: 504 
...
server: nginx
status: 504
type: POST
action: seravo_reset_shadow
resetshadow: wp-palvelu_xxxxx
...
```

There is also no indication to the user that the operation failed without opening the developer console, the button just remains disabled.

ottok:
The process should output something once a minute to not time out.

Wiljami:
In 5c8d318 an error box is added so that user can see when the timeout happens.

Unfortunately I don't think there is any easy way to create output to avoid this timeout. The php times out after one minute in the ajax call when it's running its exec command. There is no output from the exec command itself until it has completed, so I don't know how to make the php extend the timeout period during it.

elguitar:
Closing in favor of #465

PR #424, opened 2020-07-29
Add timeout error message to shadow reset

Wiljami:
The user is now informed when the ajax call times out.

When the ajax returns 504 timeout message the user now gets an error message about it.

Unfortunately there is no easy way to know if the shadow is actually resetted or not.

Related #315, #187

TODO: translations once code is ok

### DIFF
--- a/languages/seravo-fi.po
+++ b/languages/seravo-fi.po
@@ -741,6 +741,16 @@ msgstr ""
 msgid "Change Password"
 msgstr "Vaihda Salasana"
 
+#: modules/plugin-checker.php:40
+#, fuzzy
+#| msgid "Unnecessary plugins"
+msgid "Bad plugins"
+msgstr "Tarpeettomat lisäosat"
+
+#: modules/plugin-checker.php:49
+msgid "Bad plugins found: "
+msgstr ""
+
 #. translators: %s cache refresh interval
 #: modules/purge-cache.php:38
 msgid ""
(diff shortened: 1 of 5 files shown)


Wiljami:
Admittedly it is misleading. Should the message then say that the reset might or might not have occurred? As far as I know there is no way to check when the shadow has been reset last time.

elguitar:
Maybe the text could say that the reset is running in the background and that the user should check the shadow in a while. It should also say that if the user runs into trouble, she should find help from the docs and provide a link there.

Wiljami:
Changed the text and also changed the background color of the box to yellow to indicate that this is not outright error.

Screenshot from 2020-08-03 14-20-44

modules/sitestatus.php
```diff
@@ -410,6 +410,7 @@ public static function seravo_shadows_postbox() {
             </div>
           </div>
           <div class="alert" id="alert-failure"><button class="closebtn">&times;</button><p><?php _e('Failure!', 'seravo'); ?></p></div>
+          <div class="alert" id="alert-timeout"><button class="closebtn">&times;</button><p><?php _e('The shadow reset is still running on the background. You should check the status of the shadow after few minutes. If there are problems with the shadow, see the documentation from the link above.', 'seravo'); ?></p></div>
```
elguitar, line 413: after **a** few

languages/seravo-fi.po
```diff
@@ -741,6 +741,16 @@ msgstr ""
...
+#: modules/plugin-checker.php:40
+#, fuzzy
+#| msgid "Unnecessary plugins"
+msgid "Bad plugins"
+msgstr "Tarpeettomat lisäosat"
+
+#: modules/plugin-checker.php:49
+msgid "Bad plugins found: "
+msgstr ""
+
```
elguitar, line 753: Untranslated

elguitar: But this has nothing to do with this PR :D Resolving

Issue #356, opened 2020-06-10
Add SVG support


ottok:
The WordPress core does not support uploading SVG images to the media library: it is disabled by default due to security concerns.

SVG is however a great file format for the type of graphics that can use it (computer graphics, logos, drawings, other vector – many PNG images could be SVG). Nowadays even I[nternet Explorer support SVG](https://caniuse.com/#feat=mdn-svg_elements_svg), so there is no need to postpone adopting SVG images.

Some people install https://wordpress.org/plugins/svg-support/ to get SVG support. Basically it is just 10-20 lines of code that is needed, we could consider adding that to the Seravo Plugin and encourage everybody to use SVG.

In addition to having SVG support via Seravo Plugin, we could automatically sanitize them via [Scour](https://github.com/scour-project/scour/) in future if/when [Scour adds those SVG sanitization features](https://github.com/scour-project/scour/issues/254).

nellileinonen:
Because of the security concerns, should we only allow the users with administrator role to be able to upload SVG images?

The SVG sanitizer (https://github.com/darylldoyle/svg-sanitizer) that @ottok mentions in the Scour's issue has GPL-2.0 license. Because Seravo plugin has GPL-3.0 it can't be directly used.

elguitar:
For now, the safe way to go is to allow SVGs for admins only.

Once a good sanitizing process is implemented, it is easy to broaden the ability to upload SVGs for everyone.

We can go around the licensing problem by running `scour` on the command line since for example `exec('scour something');` is not linking, or by implementing the sanitizer outside of seravo-plugin.

Please proceed with the finalizing :) We will see if the sanitization function is implemented in scour and act accordingly.

ottok:
The wp-image-optimizer could be extended to run scour. The image optimizer
is anyways already run on all image uploads (but currently not for SVG, as
it is not part of the native WP image processing pipeline).

Note however that scour is not a sanitizer, just an optimizer.

Since this plugin is ours (and license is dictated by us) and the
theoretical question is the GPLv2+GPLv3 compatibility, I don't think there
are any legal risks here, using the library is also fine.

nellileinonen:
Administrators can now upload SVG images to media library. SVG is not sanitized. Using library with incompatible licence doesn't feel like the right thing to do.

ottok:
[...]

PR #425, opened 2020-08-03
Add support for SVG images

nellileinonen:
After this change, administrators can upload SVG images to media library. SVG is not sanitized.

Related issue: #356

#### Manual testing steps?
1. Login as an administrator.
2. Try to add SVG image to media library.
4. Log in as a non-admin user.
5. Try to add SVG image to media library.

### DIFF
--- a/modules/svg-support.php
+++ b/modules/svg-support.php
@@ -1,0 +1,29 @@
+<?php
+/*
+ * Support SVG images
+ */
+
+namespace Seravo;
+
+// Deny direct access to this file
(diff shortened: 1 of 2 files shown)


nellileinonen:
> Looks good!
> 
> I tested this and found out that SVG previews are not visible in wp-admin.
> image
> 
> @nellileinonen what do you think, should the previews be implemented before merging or can we live with it?

I can live with that for now. But if you're not in a hurry getting this to the new seravo plugin version I can also find out how to fix the thumbnail images :)

PR #428, opened 2020-08-03
Add translations for error messages

Wiljami:
Add Finnish translations to texts added in #421 and #423


elguitar:
Conflict in seravo.pot

Wiljami:
Fixed the conflict

PR #433, opened 2020-08-07
Fix speed test label overlap

### LINKED ISSUE #430: Speed test gui average labels overlap on similar values
If average latencies are same or very close to each other the labels added to the chart overlap:

image

Reported by @pekkakortelainen

Wiljami:
Speed test labels used to go over each other when the latency averages were too close to each other.

Now the lower value's label is underneath the line.

Worst case scenario where the labels would have been right over each other looks like following now:
image

Closes #430

### DIFF
--- a/js/sitestatus.js
+++ b/js/sitestatus.js
@@ -86,63 +86,63 @@ jQuery(document).ready(function($) {
 
     var options = {
       series: [{
-              name: seravo_site_status_loc.latency,
-              data: speedData,
-          },
-          {
-              name: seravo_site_status_loc.cached_latency,
-              data: speedDataCached,
-          }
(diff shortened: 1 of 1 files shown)


js/sitestatus.js
```diff
@@ -132,9 +132,9 @@ jQuery(document).ready(function($) {
       },
       tooltip: {
           y: {
-              formatter: function(val) {
-                  return val + " ms"
-              }
+          formatter: function(val) {
+              return val + " ms"
+            }
```
elguitar, line 137: Indentation seems to have gone wrong

elguitar:
I don't have time to test this in production. Could you @nellileinonen test and review?

nellileinonen:
Works and looks good!

PR #434, opened 2020-08-07
Sanitize SVG and allow all users upload SVG images

nellileinonen:
SVG images are sanitized and all users are allowed to upload SVG images.

Related issue: #356 

#### Manual testing steps?
1. Get or create an SVG image and add some malicious content to it.
2. Log in to your WP admin dashboard and upload the image to media library.
3. SSH to your site, find the downloaded SVG from uploads, check the file contents and see that the malicious content has been removed.
(4. If the sanitization wasn't successful, see an error message displayed after trying to upload image.)


elguitar:
Including 127,378 lines of code to seravo-plugin could induce performance issues. I will run some benchmarks.

ottok:
Wow. What a library. Does it need to load all the time? What if it was a
separate svg-sanitize.php that did runs only when asked to? E.g. `php
svg-sanitize.php /path/to/image` via cli?

elguitar:
Actually, most of its lines come from tests, which includes many svgs.

I ran a test, where I refreshed the wp-admin page and wrote down the results (master vs feature/svg-support, n=20).

The average was slightly higher on the feature/svg-support one, but [t-test](https://en.wikipedia.org/wiki/Student%27s_t-test) showed, that from this test, a difference in averages cannot be concluded.

elguitar:
If I recall correctly, @nellileinonen faced major issues when including the library manually.

`vendor/compose` has just the autoloader. It is used for the convenience of including the SVG Sanitizer. [The example above uses this method](https://github.com/Seravo/woo-pakettikauppa/tree/master/vendor) and [the author of SVG Sanitizer himself uses the autoload method in his own plugin](https://plugins.svn.wordpress.org/safe-svg/trunk/lib/vendor/).

ottok (approved):
If you think this is the best version, then feel free to proceed with it.

elguitar (approved):

PR #440, opened 2020-08-13
WIP Add hook for Restrict Content Pro login form

ypcs:
Customer notified us about wp-test testsuite not working on their site,
which uses Restrict Content Pro for providing login form. It seems that
this plugin completely overrides WordPress login processing, and doesn't
trigger `login_init` action at all.

Based on WordPress developer documentation
<https://developer.wordpress.org/reference/hooks/login_init/> this hook
*should* be fired when login form is initialized, but it seems that RCP
has done it's own decisions to skip the hook.

So, let's do bugfix for RCP and hook on their custom hook. On RCP
version 3.3.10 this is defined in `includes/login-functions.php`,
function `rcp_process_login_form()`, and has this inline comment

    /**
     * Third party plugins can use this action to add additional error checks and messages.
     *
     * @param array $_POST Posted data.
     */
    do_action( 'rcp_login_form_errors', $_POST );

So it's not meant for our purpose, and that function will also fail, as
it doesn't have correct credentials to call WordPress core function
`wp_signon()`. However we should be fine, as after this call we should
be able to access `/wp-admin/` anyways, as long as that path hasn't been
modified.


ypcs:
Seems that hookin to RCP function described in initial version isn't working, but if we hook to `init` before RCP does, that would do the job. However, this would also mean that *on every pageload* we run couple lines of the bypass check code, so if we'd like to continue with this path, we would need to do some benchmarking about effects of this.

ypcs:
Note, PR still contains some extra comments that should be removed before merging. However, this now needs at least benchmarking, and optionally comments from sites using this.

PR #446, opened 2020-09-29
Extend admin email checking

sjaks:
Extend the check for bad admin emails by listing unwanted emails
in an array. Include noreply@, no-reply@ and vagrant@.

This is a needed feature since having a no-reply or vagrant email address can make it impossible to receive WP emails.

## Testing

Changing WP admin email to something like `wp option set admin_email noreply@seravo.com` should yield:
image

### DIFF
--- a/modules/check-default-email.php
+++ b/modules/check-default-email.php
@@ -1,8 +1,9 @@
 <?php
 /**
  * Plugin name: Seravo Check Default Email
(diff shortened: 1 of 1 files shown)


ottok (approved):
I did not test running the code, but looks good to me!

elguitar (approved):
Worked great! Feel free to merge

PR #451, opened 2020-10-19
Fix PHP<7.0 compatibility issue with SVG support module

elguitar:
PHP below version 7.0 does not support the `levels` parameter of `dirname`.

This fix uses a compatible way to traverse up the directories.

### DIFF
--- a/modules/svg-support.php
+++ b/modules/svg-support.php
@@ -5,7 +5,7 @@
 
 namespace Seravo;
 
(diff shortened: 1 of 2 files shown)


ottok (approved):
LGTM!

Did not test, but indeed https://www.php.net/dirname says levels was added in 7.0.0. So this fix is needed for 1 or 2 PHP 5.6 sites :)

ypcs (approved):
LGTM!

PR #453, opened 2020-10-19
Add Strip Metadata option to Optimize Images

sjaks:
Implement a checkbox so that user can enable metadata stripping
when uploaded images are optimized.

Image optimization is run only for the thumbnails since this hooks to the `image_make_intermediate_size` filter.
The reasoning is that the original images would get backed up. This can be desired for metadata too but for some metadata types the original image probably should be stripped as well. Does @ottok have opinion on this? Should this be extended more extensively?


ottok:
This looks fine. Also the project template @Seravo/WordPress needs to be updated to set metadata stripping 'on' by default on new sites.

sjaks:
Related: https://github.com/Seravo/wordpress/pull/151

ottok (approved):
LGTM!

modules/sitestatus.php
```diff
@@ -126,6 +127,13 @@ public static function register_optimize_image_settings() {
         'optimize_images_settings',
         'seravo-optimize-images-settings'
       );
+      add_settings_field(
+        'seravo-strip-image-metadata-field',
+        __('Strip Image Metadata', 'seravo'),
```
elguitar, line 132: This is not translated yet. Do you need help with the translations?

elguitar (commented):
Looks good. There's one untranslated string.

modules/sitestatus.php
sjaks: There were two untranslated strings -> translated.

sjaks: There seems to be some abnormal changes in the translation files - at least to my eye.

elguitar: It seems to revert changes in https://github.com/Seravo/seravo-plugin/commit/617019cb49920ca6be0fffad2e20f7df67a46cdf#diff-d216ba5c5cb4f488966fc057f833dd86133eaee6e3eed47a19db4ce8f73677b5L708-L709

sjaks:
Added translations
image

elguitar (approved):
LGTM!

PR #454, opened 2020-10-22
Fix wp-user and wp-plugin logging

Wiljami:
Make it clear when plugin and theme changes were made from WP-CLI.

Log username of the user whose role is being changed.

Exclude seravotest user from the wp-user.log


elguitar (approved):
LGTM!

Issue #441, opened 2020-08-14
Missing php-error.log interpret as abnormally large


Moppa5:
**Describe the bug**
After cleaning up my own site I stumbled upon this little bug of abnormally large php-error.log when the actual case is that the file is missing. 

**To Reproduce**
1. Go to /data/log
2. Remove php-error.log file
3. Go to wp admin dashboard

**Expected behavior**
There should be some kind of check for not showing "abnormally large" message if the file doesn't exist. Maybe even prompt user on dashboard with something like "php-error.log is missing or deleted". It could also create empty php-error.log if it's missing. 

**Screenshots**
![Screenshot](https://user-images.githubusercontent.com/16706187/90245408-4927c000-de3b-11ea-85ac-23d9fcb5e24f.png)

Wiljami:
Right now the feedback for an error-log that can't be read is that. So really what needs to be done is to give relevant feedback to the user depending on the type of error.

Wiljami:
Closed by #455

PR #455, opened 2020-10-23
Add log read status as a return value

### LINKED ISSUE #442: Improve feedback on php-error.log issues
Right now there is no proper feedback for when the php-error.log has issues. There should be different messages for different situations. For example:

- missing php-error.log, like in #441
- abnormally large php-error.log file (The errors can still be displayed here!)
- abnormal php-error.logs that can cause freezes, such as in #416

Wiljami:
Reading logs now returns array with the result of read rows and status of the log file read in case the log file is large, missing or broken.

Add message to the user when the log is unreadable and message when the log file is large, but still readable.

logs.php now returns an array which contains the status of the log and the lines it has read. 

Right now the large log file threshold is at 256MB. This is from when it was a workaround fix to try and fix the log reading problems and it really didn't work. Should this threshold actually be lower since 256MB log file is already massive and even few megabyte log file is a good indicator that there is something seriously wrong. 

Closes #442 #441


elguitar (approved):
LGTM! Feel free to merge

elguitar:
No reason to get hung up on whether the limit should be 256 or something else. We can change it later if needed :)

PR #457, opened 2020-10-30
Fix mail forwards language incompatibility

elguitar:
The mail forwards were not rendered correctly, because "Forwards" was
hard-coded and it happened to be a translatable string.


ottok (approved):
LGTM!

PR #461, opened 2020-12-16
Include quarantine directories in cruft remover

sjaks:
Makes cruft remover detect directories with name quarantine-*
inside `/data` path. Quarantine directories can often grow
large and they are mostly cruft that can be removed by the user.


ottok:
Not needed as discussed today.

Issue #459, opened 2020-12-07
Non-unique element IDs found


ottok:
Noticed these on the Site Status page:
![image](https://user-images.githubusercontent.com/668724/101392174-144de600-38ce-11eb-8d64-d5759430e72f.png)

Elementary and easy to fix.

Moppa5:
It seems that the "_wpnonce" related inputs are not generated in Seravo Plugin or at least I wasn't able to find any references to any "_wpnonce".

Moppa5:
Closed via 851f68b. There were no references on seravo plugin codebase regarding any "_wpnonce" so it's generated somewhere else.

ottok:
The lines in question are:
> <input type='hidden' name='option_page' value='seravo-optimize-images-settings-group' /><input type="hidden" name="action" value="update" /><input type="hidden" id="_wpnonce" name="_wpnonce" value="390fe6f8d4" />

and

> <input type='hidden' name='option_page' value='seravo-sanitize-uploads-settings-group' /><input type="hidden" name="action" value="update" /><input type="hidden" id="_wpnonce" name="_wpnonce" value="1b5eaf8291" />

Source seems to be:

```
wp-includes/functions.php

function wp_nonce_field( $action = -1, $name = '_wpnonce', $referer = true, $echo = true ) {
	$name        = esc_attr( $name );
	$nonce_field = '<input type="hidden" id="' . $name . '" name="' . $name . '" value="' . wp_create_nonce( $action ) . '" />';
```

Are we using the function correctly? Or does WP core have a bug setting an ID to this element?

Moppa5:
@ottok, There is no usage of wp_nonce_field function on site status page. However, all the references that I found are
- list-table.php => only one instance and it doesn't use $name parameter
- seravo-postbox.php => two instances for postbox order and closed state, both use $name parameter
- upkeep.php => one instance named "seravo-updates-nonce", doesn't use $name parameter

I'm unsure whether any of these could cause the duplicate ID issue. To me it seems that the function has been used appropriately.

ottok:
> There is no usage of wp_nonce_field function on site status page

Yes there is an indirect usage of that function.

> To me it seems that the function has been used appropriately.

OK, then we'll just ignore the error emitted in the dev console if
your conclusion is this and there is nothing actionable to do.

PR #469, opened 2021-02-01
Add individual id and name for the submit buttons (Related: #459)

Moppa5:
Give individual ids and names for the submit buttons. Previously they were both just "btnSubmit".

Related to #459


elguitar (approved):
Looks good! There is no mention of `btnSubmit` anywhere in the codebase, so this should be very safe. Feel free to merge :)

Issue #471, opened 2021-02-01
Set new default max image dimensions 2560px


ottok:
Our image optimization has by default as the max image dimensions setting 4000x4000 pixels in https://github.com/Seravo/seravo-plugin/blob/8a0c06f8c29d189bfde87f713286845a63ad4615/modules/sitestatus.php#L13-L15

However, WordPress 5.3 started to enforce 2560px in core, see https://make.wordpress.org/core/2019/10/09/introducing-handling-of-big-images-in-wordpress-5-3/ (although it does still leave the big pics on the filesystem using the original name, and saves a scaled copy of `x.jpg` and as `x-scaled.jpg, and `x.jpg cannot be accessed from WordPress while only `x-scaled.jpg` is available to add to a post/page.)

I propose the Seravo Plugin would on new installations default to 2560px instead of 4000px since WordPress core would default to that anyway, so it must be a good default.

PR #476, opened 2021-02-09
Set new default max image dimensions (Closes: #471)

Moppa5:
Set the new image optimization max dimension to 2560px. This height and width value have been enforced since WordPress 5.3 version as it's stated on the issue #471


elguitar (approved):
LGTM!

Issue #466, opened 2021-01-22
Ensure all functions are properly namespaced


ottok:
I noticed that not all files define `namespace Seravo;`. Doing so would be good practice so that any functions defined in the plugin would not accidentally conflict with a function name from some other plugin.

Alternatively, if using namespace everywhere is not feasible, the functions should be prefixed with `seravo_`.

See for example `lib/*-ajax.php`.

elguitar:
The commit above by ottok can be used as a starting point.

Moppa5:
@elguitar, I will start looking this up now. I think checking and implementing the namespace related stuff is feasible However, as discussed, the class implementation is a whole another case and should be implemented separately.

Moppa5:
As far as I've found out now, a good plan is to check all the files under the lib/ and use seravo_ prefix at least on the ajax files since we have been using all along seravo_ajax_{insert_name_here} type naming. Some class files under the lib/ can be converted to use Seravo namespace. With this plan we can get seravo plugin codebase a little bit more consistent.

elguitar:
With https://github.com/Seravo/seravo-plugin/commit/5410ba9551a4eec865bde6fd18d4fc866c059c8b, all the functions are either prefixed or namespaced. However, namespacing is still considered a better practice than prefixing so we should unify the codebase by using namespacing everywhere. Breakage may occur, so testing is essential (as always :D)

Reopening :)

Note: AFAIK there is no need for class conversion when adding a namespace to a file.

PR #477, opened 2021-02-09
Ensure seravo prefix or namespace usage on modules (Closes: #466)

Moppa5:
Ensure that all the modules under the lib/ or modules/ use proper naming conventions like using Seravo namespace or seravo_ prefix on function names. Since we have been using seravo_ prefix on function names on ajax side anyway, I didn't make any major namespace related modifications. 

Some notices, I didn't force Seravo_Domains_List_table class (lib/domain-tables.php) to use Seravo namespace since the instance of this class is only used on domains-ajax.php where seravo_ prefix is already used on handling. The namespace usage would have also caused quite a lot refactoring on lib/ and modules/ regarding the domain view. JS modules have been left out since the issue doesn't really consider them and I think they are not so important when it comes to this namespace / prefix thing. 

See issue #466

### DIFF
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -4,7 +4,7 @@ if ( ! defined('ABSPATH') ) {
   die('Access denied!');
 }
 
-function add_file_information( $file ) {
+function seravo_add_file_information( $file ) {
   exec('du ' . $file . ' -h --time', $output);
   $size = explode("\t", $output[0]);
 
@@ -15,7 +15,7 @@ function add_file_information( $file ) {
 
 }
 
-function find_cruft_file( $name ) {
+function seravo_find_cruft_file( $name ) {
   $user = getenv('WP_USER');
   exec('find /data/wordpress -name ' . $name, $data_files);
   exec('find /home/' . $user . ' -maxdepth 1 -name ' . $name, $home_files);
@@ -23,7 +23,7 @@ function find_cruft_file( $name ) {
   return $files;
 }
 
-function find_cruft_dir( $name ) {
+function seravo_find_cruft_dir( $name ) {
   $user = getenv('WP_USER');
   exec('find /data/wordpress -type d -name ' . $name, $data_dirs);
   exec('find /home/' . $user . ' -maxdepth 1 -type d -name ' . $name, $home_dirs);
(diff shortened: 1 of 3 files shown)


lib/cruftfiles-ajax.php
```diff
@@ -59,23 +59,23 @@ function find_cruft_core() {
...
 }
 
-function list_known_cruft_dir( $name ) {
+function seravo_list_known_cruft_dir( $name ) {
   exec('ls -d ' . $name, $output);
   return $output;
 }
 
-function rmdir_recursive( $dir, $recursive ) {
+function seravo_rmdir_recursive( $dir, $recursive ) {
```
elguitar, line 72: I think that this `rmdir_recursive` is actually called from `Security::rmdir_recursive` (line 404). What do you think, should it be changed to `self::rmdir_recursive`, so that it really is recursive? Seems like a bug to me.
```
petri@t470s:~/seravo/seravo-plugin> ag rmdir_recursive
lib/cruftfiles-ajax.php
72:function seravo_rmdir_recursive( $dir, $recursive ) {
78:      seravo_rmdir_recursive("$dir/$file", 1);
262:            $unlink_result = seravo_rmdir_recursive($file, 0);

modules/security.php
382:                $unlink_result = self::rmdir_recursive($file, 0);
398:    public static function rmdir_recursive( $dir, $recursive ) {
404:          rmdir_recursive("$dir/$file", 1);
```

elguitar (commented):
LGTM! Could you test, if the cruftfile removal is really buggy or not?

lib/cruftfiles-ajax.php
Moppa5: @elguitar, Interesting indeed. They seem to be exactly the same function both on security.php and cruftfiles-ajax.php with some refactoring of course since the function names are different. However, I think that the security.php one is excessive since the ajax remove does the job already. What do you think about that?

Moppa5: @elguitar, I changed that line on cruftfiles-ajax.php to `Seravo\Security::rmdir_recursive("$dir/$file", 1);` and it still seems to work normally with some basic folders and files. However, we can't use `self::` outside of the class, am I right?

elguitar: Okay cool. If it works, it is fine by me to merge this :)

elguitar (approved):
LGTM please feel free to merge

Issue #472, opened 2021-02-01
Put defaults on sites that don't have any values set manually


ottok:
Currently on all new sites the [installer.php in Seravo/WordPress](https://github.com/Seravo/wordpress/blob/master/htdocs/wp-content/install.php#L610-L619) sets the image optimizer and other features on by default for new sites.

The settings are WordPress options, stored in the database. However, if a new site is migrated, along comes its database and these settings disappear.

I propose that Seravo Plugin would sniff if these settings have any value set, and it they don't exists at all, it would set a default value. Such a behavior would make the default apply to both new sites, migrated sites as well as old sites.

Since the defaults are pretty sane and have served us well, having them apply to old sites as well seems pretty safe and sensible at this point.

elguitar:
The options that need to be set are found in https://github.com/Seravo/wordpress/blob/master/htdocs/wp-content/install.php#L610-L619

Docs for handling options in WP: https://developer.wordpress.org/reference/functions/get_option/

You can add a new module and include it in `seravo-plugin.php`.

ottok:
Follow-up PR filed at https://github.com/Seravo/wordpress/pull/177

PR #480, opened 2021-02-17
Create check-default-options module

frimro:

The module sets checks for default options, and if they are not set, it sets the missing options on.

Closes: #472

#### Manual testing steps?

1. Removing said options from a development site, and generating them back using the module
2. Setting an option to be not used, and test that the module does not overwrite it


elguitar (approved):
Very nice! Feel free to rebase & merge

PR #482, opened 2021-02-22
Fix margins

AatuYH:
Fixes some of the margins in the plugin so elements are no longer right next to each other. Also fixes some of the small issues with the css with different screen sizes.

Some of the margins didn't work properly with different screen sizes and caused some of the elements to appear misaligned. After the fixes the elements should look okay on most modern screen sizes.

**Before**

**After**


elguitar (approved):
Nice! Feel free to merge :)

Issue #483, opened 2021-02-22
Not aligned input elements in Search-Replace Tool


elguitar:
In the English version, the input elements are misaligned, it would be a lot nicer, if they would start from the same point. This does not happen in the Finnish version, because "Mistä:" and "Mihin:" are quite the same width, when rendered by the browser.

To play it safe, the solution should be CSS only. This way the logic behind the tool should not break. If some structural changes (like adding new elements or js) are needed, moro thorough testing will be required.

![image](https://user-images.githubusercontent.com/16290563/108713166-84105c00-7520-11eb-9305-7b96fe85733a.png)

PR #484, opened 2021-02-23
Fix input element misalignment (Closes: #483)

AatuYH:
The English language version of the plugin had some issues with the alignment of the input elements. The problem was the words "From" and "To" being different length. Giving the labels a width fixed the issue.

**Screenshots before and after**
before
after


elguitar (approved):
Brilliant! Works like a charm :) Feel free to rebase and merge

PR #485, opened 2021-02-23
Allow caching of HTTP requests to certain domains

ypcs:
This allows developers easily enforce caching of HTTP requests to
certain domains. This makes it easy to eg. force caching of slow API
queries etc.

Example, to cache POST and GET requests to api.wpml.org, one could write

    add_filter('seravo_cache_http_post_api_wpml_org', '__return_true', 10, 1);
    add_filter('seravo_cache_http_get_api_wpml_org', '__return_true', 10, 1);

in eg. functions.php.

Note: This doesn't currently check that eg. ports match.

Add new filters to enable cache enforcing.

This allows easily caching certain third-party API requests that return data very slowly, and then cause horrible slowdowns. As we have no control of those third-party servers there is very little we can do to make them always perform good > second best option is to reduce number of requests against them.

This should not cause any regressions. This will not by default do anything, developer must explicitly add some filters to enable caching for specific domain.

### DIFF
--- a/modules/thirdparty-fixes.php
+++ b/modules/thirdparty-fixes.php
@@ -35,6 +35,56 @@ if ( ! class_exists('ThirdpartyFixes') ) {
 
       // Maybe log SQL queries
       add_action('shutdown', array( $this, 'log_queries' ));
+
+      // Maybe cache HTTP requests
+      add_filter('pre_http_request', array( $this, 'http_maybe_use_cached' ), 10, 3);
+      add_filter('http_response', array( $this, 'http_maybe_cache' ), 10, 3);
+    }
+
+    /**
+     * Maybe used cached results for HTTP request
+     *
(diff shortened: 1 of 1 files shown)


elguitar (approved):
Tested with: `wp eval-file cachetest.php`, where 

cachetest.php:
```
<?php

add_filter('seravo_cache_http_get_google_com', '__return_true', 10, 1);
echo 1;
wp_remote_get('https://google.com');
echo 2;
wp_remote_get('https://google.com');
echo 3;
wp_remote_get('https://google.com');
```

On first run, it outputted 1 quickly, 2 slowly and 3 quickly again.

On the second run, everything was quick.

Works as intended! :) :1st_place_medal: 

Do you have a plan about how to document this for the customers?

ypcs:
I don't have plans to do it, but of course we could and should do it.

Issue #487, opened 2021-02-23
Executing wp-test causes two blank lines on php error log


Moppa5:
**Describe the bug**
For some reason, when executing wp-test for instance on upkeep.php page "Update tests", two blank php error lines are caused every time the wp-test command is run. Only time stamp is available. See the following
```
[23-Feb-2021 11:01:12 UTC] 23/Feb/2021:10:47:27 +0000
[23-Feb-2021 11:01:12 UTC] 2021-02-23T10:47:27+00:00
```
These same blank lines are also generated when running `wp-test` on command prompt.

**To Reproduce**
1. Go to 'upkeep.php' and to section 'Update tests'
2. Click on 'Run Tests' button
3. Check your php-error.log
4. See the two blank warning lines

**Expected behavior**
PHP error log shouldn't be filled with empty lines/warning, or even better running command `wp-test` shouldn't cause these issues in the first place.

ypcs:
https://github.com/Seravo/seravo-plugin/blob/3a1ec730266c5a3a67d038535b40b58d70e8464c/modules/login-notification.php#L182-L183

PR #488, opened 2021-02-23
Fix excessive error logging (Closes: #487)

Moppa5:
There were some excessive `error_log()` usage on login-notification.php module that caused empty lines to php-error.log with only timestamp on them. This removes those excessive errors.

See issue #487


elguitar (approved):
They seem to be added in 8b6e0dc4. They have no functional purpose and seem like they have been forgotten to the code by accident.

The removal is therefore justified. Please feel free to merge :)

PR #490, opened 2021-03-01
Ensure seravo namespace usage on all modules (Closes: #466)

Moppa5:
Ensure that all modules under folders lib/ and modules/ are properly namespaced. Note, please also test carefully the modules etc. in case of any errors. I've tested all the pages comprehensively and all features and functionality seem to work alright / normally as before. 

See #466

### DIFF
--- a/lib/backups-ajax.php
+++ b/lib/backups-ajax.php
@@ -1,5 +1,7 @@
 <?php
 
+namespace Seravo;
+
 // Deny direct access to this file
 if ( ! defined('ABSPATH') ) {
   die('Access denied!');
--- a/lib/cruftfiles-ajax.php
+++ b/lib/cruftfiles-ajax.php
@@ -1,4 +1,7 @@
 <?php
+
+namespace Seravo;
+
 // Deny direct access to this file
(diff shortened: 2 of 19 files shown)


elguitar (approved):
Looks good!

It seems that `Seravo\` was added to functions (`Seravo\usort_reorder_domains`) and removed from classes (`API::get_site_data`). Could you add the logic behind this to the commit message?

Then feel free to merge :)

Moppa5:
@elguitar, I think the adding you mentioned is quite self-evident. Lines like `usort($data, 'Seravo\usort_reorder');` require `Seravo\` namespace if they are passed as a string argument. `Seravo\` was removed from for instance `API::get_site_data` since the class was converted to use that namespace, and there would have been no `Seravo\Seravo\API::get_site_data`. 

What do you think? Surely we can add more information, but as I see it's self-evident on this case.

elguitar (approved):
Oh! That was my bad, sorry about that. Of course they need the explicit namespace when they are passed as strings. :sweat_smile: 

Feel free to merge :)

Issue #486, opened 2021-02-23
PHP error log gets filled with undefined array key


Moppa5:
**Describe the bug**
I noticed that going into upkeep.php page causes some excessive PHP warning about undefined array key. The warning exactly is the following one:
`[23-Feb-2021 10:50:47 UTC] PHP Warning:  Undefined array key "report" in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/seravo-plugin.php on line 142`

Even the exact line is provided here. I think it would be best to either check if the `$_GET["report"]` exists or even better find out why in some instances it doesn't exist and why it's nagging now. 

Also, I found out that going into domains.php page causes the following warning:
```
[23-Feb-2021 11:07:36 UTC] PHP Warning:  Undefined array key "hook_suffix" in /data/wordpress/htdocs/wordpress/wp-admin/includes/class-wp-screen.php on line 223
[23-Feb-2021 11:07:36 UTC] PHP Warning:  Undefined array key "hook_suffix" in /data/wordpress/htdocs/wordpress/wp-admin/includes/class-wp-screen.php on line 223
```

**To Reproduce**
1. Keep track of your php-error.log or alternatively remove it
2. Go to upkeep.php and domains.php site
3. Check the php-error.log 
4. See the warnings described

**Expected behavior**
There shouldn't be any excessive errors about missing array keys. Array keys should be defined or otherwise handled / checked that they exist.

Moppa5:
Closing this as the fundamental `$_GET['report']` related array key errors are now fixed, and the discussion can be moved and worked separately on #478.

Issue #478, opened 2021-02-12
PHP Fatal error and warnings on Seravo Plugin pages with PHP8


pekkakortelainen:
Browsing through Tools pages provided by Seravo Plugin results in a fatal error on Domains page and warnings on other pages when running PHP8. 

**Steps to reproduce:**

1. Set PHP mode to 8.0 in `/data/wordpress/nginx/php.conf`
2. Restart NGINX: `wp-restart-nginx`
3. Use `wp-watch-php` or `tail -f /data/log/php-error.log` to monitor PHP error log
4. Browse through pages provided by Seravo Plugin in wp-admin and monitor the PHP error log simultaneously

**Found errors and warnings:**

Domains page
```
PHP Fatal error:  Uncaught TypeError: date_format(): Argument #1 ($object) must be of type DateTimeInterface, bool given in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/domain-tables.php:98
Stack trace:
#0 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/domain-tables.php(98): date_format(false, 'j.n.Y H:i')
#1 [internal function]: Seravo_Domains_List_Table->column_expires(Array)
#2 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/list-table.php(1357): call_user_func(Array, Array)
#3 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/domain-tables.php(274): Seravo\WP_List_Table->single_row_columns(Array)
#4 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/list-table.php(1289): Seravo_Domains_List_Table->single_row(Array)
#5 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/list-table.php(1274): Seravo\WP_List_Table->display_rows()
#6 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/domain-tables.php(294): Seravo\WP_List_Table->display_rows_or_placeholder()
#7 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/domains-ajax.php(23): Seravo_Domains_List_Table->display()
#8 /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/lib/domains-ajax.php(313): seravo_get_domains_table()
[...]

PR #491, opened 2021-03-01
Check array key at get (Related: #486, #478)

Moppa5:
Quick fix for `undefined array key "report"` warnings issued when going to upkeep page. 

See #486 and #478


elguitar (approved):
Looks good! Feel free to merge :)

Issue #492, opened 2021-03-01
Plugin causes huge amounts of errors when working locally, due to incorrect value passed to fopen


SeenNotHurd:
**Describe the bug**
When working locally (at least I haven't seen this in production) errors are appearing such as the following:
`Warning: fwrite() expects parameter 1 to be resource, bool given in *******/mu-plugins/seravo-plugin/modules/thirdparty-fixes.php on line 69`

This has only started to appear I believe after 1.9.25 or .26.

**To Reproduce**
1. Setup a project locally
2. Open the project in your browser
3. See error

**Expected behavior**
If there is something relying on production settings or data, check for an environment or similar var.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser agnostic

**Additional context**
-

Moppa5:
I think that the issue is here
```
$handle = fopen($logfile, 'a');
fwrite($handle, '### ' . date(\DateTime::ISO8601) . ' sid:' . $_SERVER['HTTP_X_SERAVO_REQUEST_ID'] . ' total:' .        $wpdb->num_queries . chr(10));
        foreach ( $wpdb->queries as $q ) {
            fwrite($handle, "SQL: $q[0]" . chr(10));
            fwrite($handle, "Time: $q[1] s" . chr(10));
            fwrite($handle, "Calling functions: $q[2]" . chr(10));
            fwrite($handle, "Query begin: $q[3]" . chr(10));
            fwrite($handle, 'Custom data: ' . print_r($q[4], true) . chr(10) . '--' . chr(10));
        }
fwrite($handle, '### EOF' . chr(10));
```
fopen() returns false since it fails to open the given logfile thus all the fwrite() errors. Could be quite easily fixed by checking the $handle status before attempting to write. 

It also seems that your seravo-plugin on local is running either 1.9.25 or .26 since currently there are no fwrites anywhere near on line 69. You might want to update your seravo-plugin.

Moppa5:
@SeenNotHurd, There is now a fix for the fwrite() errors on Seravo Plugin master. Could you pull it or run `wp-seravo-plugin-update --dev` and report whether the issue is now fixed?

JoosuaKoskinen:
@Moppa5 Is there still something to do or can this be closed?

Moppa5:
I think we can close the issue now as there's the current fix nor for `$handle` variable. I haven't been able to reproduce this issue.

PR #493, opened 2021-03-02
Check that logfile can be opened before writing (Related: #492)

Moppa5:
Check that the file pointer is truly returned when opening the logfile. This fixes situations where fopen returns false and fwrite attempts multiple times to write on to those false values. 

See #492


elguitar (approved):
LGTM! :)

Issue #494, opened 2021-03-02
Changes status feature not working on upkeep page


Moppa5:
**Describe the bug**
On upkeep page there's this changes status feature for running command `wp-backup-list-changes-since`. However, for some reason the feature is not working any more as clicking "Run" doesn't work or load anything.

**To Reproduce**
1. Go to upkeep page
2. Find changes status feature
3. Click "Run" button to execute
4. See that nothing happens

**Expected behavior**
Pressing the "Run" button should execute the command and output some results even if there are no changes at all.

PR #495, opened 2021-03-02
Remove changes status feature from load function (Closes: #494)

Moppa5:
Remove the changes status ajax and JS side functions from under the window load function. This should fix the changes status feature not working. On the other hand, the JS window load function is excessive and unnecessary on this instance as well. 

See #494


elguitar (approved):
Good find! Feel free to merge

PR #496, opened 2021-03-04
Add 'apply_filters' to $site_info

spktklr:
Add 'apply_filters' to $site_info.

Disable seravo_updates via PHP.

upkeep.php row 133.

#### Manual testing steps?
Not required.

#### Screenshots
Not applicable.

### DIFF
--- a/modules/upkeep.php
+++ b/modules/upkeep.php
@@ -130,7 +130,7 @@ if ( ! class_exists('Upkeep') ) {
 
     public static function seravo_admin_get_site_info() {
       $site_info = API::get_site_data();
-      return $site_info;
+      return apply_filters( 'seravo_site_info', $site_info );
     }
 
     public static function seravo_updates_postbox() {


spktklr:
Perhaps the pull request was a bit hasty and it is not actually possible to disable automatic updates in code, but the setting has to be submitted to Seravo's server?

Is there another way to do this in code?

Moppa5:
@spktklr, Is there a specific reason for you to disable this in code? There is this checkbox on upkeep page if you wish to opt out from Seravo automatic updates (see the screenshot). 
 
Screenshot(2)

spktklr:
@Moppa5 thanks for looking into this. Writing the browser automation to disable this on all sites seems more laborous and non-future-proof than just having adding it at scale in version control.

Moppa5:
@spktklr, Sorry for not getting back to you sooner. As mentioned and noticed earlier this fix is unfortunately not going to work. However, you have good idea here and we will be looking this on our internal implementation as having an effect requires modifications there. Meanwhile we can open an issue about this case here on Seravo Plugin and inform there about the development.

PR #497, opened 2021-03-12
Fix PHP related fatal error and array key warnings (Related: #478)

Moppa5:
Fix fatal error with timestamp returning false and trying to pass it to `date_format()` function. Also, fix undefined subdomain array key warnings by checking they exist. The most crucial fix is the PHP fatal error fix. 

See #478


Moppa5:
Merging this after a self-review.

PR #498, opened 2021-03-18
WIP admin ajax post logging

Wiljami:
Log admin ajax post data to `/data/log/admin-ajax.log`


Wiljami:
This isn't something we want seravo-plugin take care after all.

Issue #499, opened 2021-03-19
Unuseful "Technical contacts" link on Site Status page


Moppa5:
**Description**
I noticed that on Site Status page there is a link "Technical Contacts" which redirects user to upkeep page. However, for a user who is not aware of the contacts form / feature on upkeep page this is not very useful.

**Describe the solution you'd like**
We should hook the link to directly go to "Contacts" section on upkeep page and thus avoid any inconvenience or confusion. User should not scroll around and think about why the link redirected to this particular page. 

**Screenshots**
![Screenshot](https://user-images.githubusercontent.com/16706187/111789459-d5b2c900-88c9-11eb-9929-f5d6a52b37c6.png)

PR #500, opened 2021-03-19
Add id and use it on link to redirect directly (Closes: #499)

Moppa5:
Update "Technical Contacts" link on Site Status page to redirect directly to Contacts section on upkeep page.

#499


Moppa5:
Merging this after self review as it's very minor enhancement.

PR #504, opened 2021-03-23
Make speedtest button scroll to speedtest postbox

JoosuaKoskinen:
Appended `#seravo-postbox-speed-test` to speedtest button on admin bar. It took me a while to understand what the button does as the speedtest postbox was at the bottom of the page. The button now scrolls down to speedtest postbox.


Moppa5 (approved):
Looks great and is much more consistent and clear now. Feel free to merge.

PR #507, opened 2021-03-26
Fix last successful logins

JoosuaKoskinen:
The function should only list successful logins but
there were some failed logins too. Fixed now.


elguitar (approved):
LGTM! Feel free to merge

PR #508, opened 2021-03-26
Add dashboard widget to show site status

Moppa5:
Add a new dashboard widget and show plan details and monthly http requests up to 4 months. It also includes the PHP error count notify. There is also a yellow color indicating if there are more HTTP requests than plan allows.

Also, includes translations.

See #5 

#### Screenshots
Screenshot(1)


elguitar:
I think it looks wonderful! I am assigning someone outside of dev team to get an alternative opinion.

elguitar:
Hello @pekkakortelainen! Could you provide us with some thoughts on what could be useful information for this dashboard widget? This would be visible in the wp-admin dashboard front page to the customers.

pekkakortelainen:
Nice! I think the yellow color will bring a lot more attention to HTTP request tracking. Adding a sentence could help customers in need of a plan upgrade know what to do, e.g. "Contact help@seravo.com to upgrade your [plan](https://seravo.com/plans/)". 

Could we have multiple widgets like this in the future? Could it be useful to have separate widgets for e.g. PHP version (notify if at EOL), disk space? Adding too much information inside one box can lead to dismissing everything. We have these features already under Tools so the data is there.

Moppa5:
@pekkakortelainen, I was thinking about adding some text there as well. Some description text and the tip to contact help. 

Theoretically, we can have as many widgets as we want. There are currently some login notifications like for example the PHP error count (currently) as a separate widget, last login notification, some log file related notifications etc. Do you think that we should have for instance the php error count as a separate widget rather than including it on site status? I would keep the php error count on this site status widget for now, but for future there could be a separate widget for showing php error count and the mentioned php end of life? What do you think?

pekkakortelainen:
Sounds like a good plan! Site status as described above for now and separate widgets in future versions.

modules/login-notification.php
```diff
@@ -248,6 +246,98 @@ public static function display_admin_errors_notification() {
...
 
+    public static function display_site_status() {
+      $site_info = API::get_site_data();
+      $http_requests = $site_info['plan']['httplimit'];
+      echo '<h3>' . __('Plan details', 'seravo') . '</h3>';
+      echo '<div>' . __('Plan type: ', 'seravo') . $site_info['plan']['type'] . '</div>';
+      echo '<div>' . __('HTTP requests / month: ', 'seravo') . $http_requests . '</div><br>';
+
+      // logic ported to PHP from sitestatus page
+      if ( ! Helpers::is_production() ) {
```
JoosuaKoskinen, line 257: This should probably be called before any API calls

modules/login-notification.php
```diff
@@ -248,6 +246,98 @@ public static function display_admin_errors_notification() {
...
+        __('This feature is available only on live production sites.', 'seravo');
```
JoosuaKoskinen, line 258: `__`only returns the translations, use `_e`to print it. Also maybe `return` after this or the code below will be executed anyways.

modules/login-notification.php
```diff
@@ -248,6 +246,98 @@ public static function display_admin_errors_notification() {
...
+      }
+      ?>
+      <div>
+        <h3><?php _e('Monthly HTTP requests', 'seravo'); ?></h3>
+        <?php
+          $contact_email_url = '<a href="mailto:help@seravo.com">help@seravo.com</a>';
+          $msg = wp_sprintf(
+            /* translators:
+           * %1$s contact email link
```
JoosuaKoskinen, line 267: Multiline `translators` comments will cause warnings on `wp i18n`

modules/login-notification.php
```diff
@@ -248,6 +246,98 @@ public static function display_admin_errors_notification() {
...
+          </thead>
+          <tbody id="http-reports_table"></tbody>     
+      
+      <?php
+      $reports = glob('/data/slog/html/goaccess-*.html');
+      // Track max request value to calculate relative bar widths
+      $max_requests = 0;
+
+      foreach ( array_reverse($reports) as $report ) {
+        $total_requests_string = exec("grep -oE 'total_requests\": ([0-9]+),' $report");
```
JoosuaKoskinen, line 295: Haven't tested this code in practice yet but this looks a bit off. Does it work?

JoosuaKoskinen (changes requested):

modules/login-notification.php
Moppa5: It surely should. It's fixed now.

modules/login-notification.php
JoosuaKoskinen: `// translators: %1$s contact email link`

modules/login-notification.php
```diff
@@ -248,6 +246,104 @@ public static function display_admin_errors_notification() {
...
+        $bar_border_color = 'none';
+
+        if ( $total_requests > $http_requests ) {
+          // display warning
+          $bar_border_color = 'solid 0.5em #e8ba1b';
+        }
+        // generate table element of this data
+        echo '<tr><td><a href="tools.php?x-accel-redirect&report=' . $month_date . '.html" target="_blank"> ' .
+        $month_date . ' </a> </td> <td><div style="background: #44A1CB; border-right: ' . $bar_border_color .
+        '; color: #fff; padding: 3px; width: ' . $bar_css . '%; display: inline-block;"><div style="white-space: nowrap;">' .
```
JoosuaKoskinen, line 331: `width: ' . $bar_css . '%;`

`%` is appended after $bar_css even if it's `auto`

modules/login-notification.php
```diff
@@ -248,6 +246,104 @@ public static function display_admin_errors_notification() {
...
+
+        if ( $report_month_counter === 4 ) {
+          break;
+        }
+        $report_month_counter++;
+
+        $total_requests_string = exec("grep -oE 'total_requests\": ([0-9]+),' $report");
+        preg_match('/([0-9]+)/', $total_requests_string, $total_requests_match);
+        $total_requests = intval($total_requests_match[1]);
+        if ( $total_requests > $max_requests ) {
```
JoosuaKoskinen, line 304: `$max_requests` needs to be counted beforehand or the order affects the bar size:

site-status-requests-1
site-status-requests-2

modules/login-notification.php
```diff
@@ -248,6 +246,104 @@ public static function display_admin_errors_notification() {
...
+      <div class="http-requests_info_loading" style="padding: 0px;">
+        <table class="widefat striped" style="width: 100%; border: none;">
+          <thead>
+          <tr>
+            <th style="width: 25%;"><?php _e('Month', 'seravo'); ?></th>
+            <th style="width: 50%;"><?php _e('HTTP Requests', 'seravo'); ?></th>
+            <th style="width: 25%;"><?php _e('Report', 'seravo'); ?></th>
+          </tr>
+          </thead>
+          <tbody id="http-reports_table"></tbody>
```
JoosuaKoskinen, line 286: Is there a reason to close `tbody` here and not on line `337`?

JoosuaKoskinen (changes requested):

modules/login-notification.php
Moppa5: Not really sure since this code is ported from JS version. I will take a look at this and possible other excessive lines.

modules/login-notification.php
```diff
@@ -248,6 +246,104 @@ public static function display_admin_errors_notification() {
       echo '<div>' . $msg . '</div>';
     }
 
+    public static function display_site_status() {
+      if ( ! Helpers::is_production() ) {
+        return _e('This feature is available only on live production sites.', 'seravo');
+      }
+
+      $site_info = API::get_site_data();
```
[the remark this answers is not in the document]
Moppa5: I'm unsure whether this is ok `return _e( $site_info->get_error_message() );` ? Or should there be some other message like "failure fetching plan data" etc. ?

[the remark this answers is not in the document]
JoosuaKoskinen: No variables in `_e`. Probably best to print some failure message and log the real error message in php-error.log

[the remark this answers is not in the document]
Moppa5: Yeah noticed as well. I put it to `error_log` and made a error message print.

modules/login-notification.php
```diff
@@ -248,6 +246,104 @@ public static function display_admin_errors_notification() {
...
+          </thead>
+          <tbody id="http-reports_table"></tbody>     
+      
+      <?php
+      $reports = glob('/data/slog/html/goaccess-*.html');
+      // Track max request value to calculate relative bar widths
+      $max_requests = 0;
+      $report_month_counter = 0;
+
+      foreach ( array_reverse($reports) as $report ) {
```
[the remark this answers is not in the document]
Moppa5: There will be an error string indicating "The site has no HTTP request statistics yet".

modules/login-notification.php
Moppa5: Removed the tbody and it didn't affect anything. Presumably it had some functionality on JS side but not on this case

modules/login-notification.php
Moppa5: This should be now fixed on the wip commits.

elguitar (approved):
I don't have anything to nag about here :D @JoosuaKoskinen thoughts?

JoosuaKoskinen (approved):
LGTM!

PR #509, opened 2021-03-26
Clarify the translation process on README

Moppa5:
Update and clarify the translation process on README. Basically, add instructions how to fully update translations.

In general, documentation for making translations should be available fully rather than partially. This also seems to be something that people forget quite often and thus documenting it is beneficial for all.


JoosuaKoskinen (approved):
Much more useful :+1:

elguitar (approved):
Brilliant! We will merge this to include this into the next release :)

PR #511, opened 2021-03-29
Fix timezone error

JoosuaKoskinen:

- Fixes wp-login.log regex
-  Removes usage of `wp_timezone_string` so older core versions work


elguitar (approved):
Looks good :) Feel free to merge

Issue #517, opened 2021-03-31
More informative dashboard widgets


Moppa5:
**Description**
There has been discussion about wp-admin dashboard widgets. Especially it was requested that there could be more informative and notification stylish widgets for instance warning regarding usage of PHP version that is soon EOL, low disk space etc. More over the current PHP error count notification could be integrated with EOL PHP version warning. 

**Implementation**
Currently the #5 dashboard widget is implemented on login-notifications.php. However, in order for the login-notifications.php not becoming too large I'd propose that there would be some kind of dashboard_widgets class acting as a helper or completely for the widgets. This way we can implement the widgets and notifications separately as well. What do you think? 

**Additional context**
See #5 and #508 as examples

Moppa5:
There is now a class for dashboard widgets implemented on branch [`feature/dashboard-widgets`](https://github.com/Seravo/seravo-plugin/tree/feature/dashboard-widgets). As mentioned on this issue it might be better implementing these dashboard widgets separately from login notifications thus the new class. 

What do you think @JoosuaKoskinen and @elguitar?

JoosuaKoskinen:
Large files are no problem but for keeping the project structure nice and modular, it's a good change.

:+1: from me

Moppa5:
To sum up, this far dashboard widgets include
- PHP error count notification
- PHP EOL notification integrated with PHP error widget
- Low disk space notification if the plan is using more than 90% of the space

Moppa5:
Closing this issue as the foundations for extending and implementing new dashboard widgets have been made. We can make separate concrete issues about new features / widgets.

PR #519, opened 2021-04-14
Add widget class and some new dashboard widgets

Moppa5:
Add Dashboard_Widgets class for wp-admin dashboard widgets class. Extend dashboard widgets class by adding PHP end of life version and low disk space notifications. Also, move PHP error count widget from login notifications to the dashboard widgets class together with the new site status widget. Includes translations. 

PHP EOL needs reworking when the common config file / internal implementation is done. 

See #517

#### Manual testing steps?
To test end of life, error count and low disk space widgets there are some requirements:
- Some PHP errors on log (artificial or not)
- Switch to PHP 7.2
- `const LOW_DISK_SPACE_USAGE = 0.9;` change the value to something like 0.1 for instance. See the line [here](https://github.com/Seravo/seravo-plugin/blob/9d37bbb5afa32bbb256733ed189e8e1635f54283/modules/dashboard-widgets.php#L24)

Note that disk space usage is stored as a transient for 15 minutes, thus you need to delete transients and purge cache if you want to re-calculate and usage results. 

#### Screenshots
PHP EOL and warnings + disk space widget. For site status widget see #508 
Screenshot(3)


modules/dashboard-widgets.php
```diff
@@ -0,0 +1,301 @@
...
+      );
+
+      // Get total disk usage
+      $return_code = exec('du -sb /data ' . implode(' ', $exclude_dirs), $data_folder);
+      $data_size = 0;
+
+      if ( $return_code !== false ) {
+        $data_size = preg_split('/\s+/', $data_folder[0])[0];
+      }
+      $plan_details = API::get_site_data();
```
JoosuaKoskinen, line 93: I'm bit worried about this and the `exec` above. A lot of slow things are done every time accessing dashboard. I saw a difference up to 50-100ms in load time and we want to be as fast as possible. I believe the load time might increase even more on sites with a lot of individual files for `du`to check.

`get_site_data` is called multiple times on dashboard load and it takes the most time on my site. Maybe the resulting `['plan']['disklimit']` should be cached into transient?

modules/dashboard-widgets.php
```diff
@@ -0,0 +1,301 @@
...
+
+      if ( is_wp_error($plan_details) ) {
+        $plan_disk_limit = 0;
+      } else {
+        $plan_disk_limit = $plan_details['plan']['disklimit']; // in GB
+      }
+
+      if ( $plan_disk_limit !== 0 && $data_size !== 0 ) {
+        $data_size_human = (int) Helpers::human_file_size($data_size);
+        $relative_disk_space_usage = $data_size_human / ($plan_disk_limit * 1000);
```
JoosuaKoskinen, line 103: This relative disk space usage doesn't work. My site has 1.2GB in files with 1GB limit and the alert box didn't show up. It seems to assume `human_file_size` to always give the result in MB but it can give it any format. 

On my site, `human_file_size` returned `"1.2GB"` but as it's converted it into int it was `1` making this statement `$relative_disk_space_usage = 1 / (1 * 1000)` which is incorrect relative disk space usage.

See https://github.com/Seravo/seravo-plugin/blob/master/lib/helpers.php#L38

JoosuaKoskinen (changes requested):
LGTM otherwise!

modules/dashboard-widgets.php
Moppa5: Good point! Caching sounds a good solution to this. It needs some reworking for now.

modules/dashboard-widgets.php
Moppa5: I was actually afraid that the helper method wouldn't work in the desired way. Need to re-work the calculations.

modules/dashboard-widgets.php
Moppa5: Now added transients for 1 hour both in dashboard widgets and site status page.

modules/dashboard-widgets.php
Moppa5: Re-worked the calculations and it should now apply the MBs and calculate the relative usage properly.

Moppa5:
@JoosuaKoskinen, I updated and squashed the branch. I added transients for 15 minutes to store the disk space usage and not calculate it all the time. I changed the logic how the relative disk space is calculated and it should work now as intended. Also, there are some minor changes and improvements + translations as well.

JoosuaKoskinen:
> What do you think is the new implemented one excessive?

I don't think so, no. The built-in PHP nag wasn't introduced until core 5.1.0 which there's still a lot of sites running on.

We either need to check if wp core version is >= 5.1.0 and hide the new widget or just hide the core built-in one completely. I think we should do the latter one as the instruction links in the built-in one doesn't apply in our environment (we wan't to link to https://help.seravo.com and PHP-changer widget).

Moppa5:
@JoosuaKoskinen, Good points! I ended up hiding the WP PHP nag as mentioned that our is more Seravo specific. Transient case, calculations and the hiding are now done. Could you review whether we can merge the feature or not?

modules/dashboard-widgets.php
```diff
@@ -0,0 +1,328 @@
...
+            $bar_size = $total_requests / $max_requests * 100;
+          } else {
+            $bar_size = 1;
+          }
+
+          if ( $bar_size <= 10 ) {
+            $bar_css = 'auto';
+          } else {
+            $bar_css = strval($bar_size) . '%';
+          }
```
JoosuaKoskinen, line 304: Even with this, the bar is sometimes too small and the content overflows. Maybe just remove this and add `min-width: $bar_size%; width: auto;` to the bar css so it's always at least wide enough to fit the content.

JoosuaKoskinen (changes requested):
Looks really good! Feel free to merge after resolving that one change request.

PR #521, opened 2021-04-20
Add search-replace extension for multisites

JoosuaKoskinen:
Extends wp-cli `wp search-replace` to replace `DOMAIN_CURRENT_SITE` on multisites when needed.

Prevents multisites from breaking on automatic domain changes or manual search-replace.

```
$ wp option get home
https://joosua.fi
$ grep -F htdocs/wp-config.php -e DOMAIN_CURRENT_SITE 
define( 'DOMAIN_CURRENT_SITE', 'joosua.fi' );

$ wp search-replace 'joosua.fi' 'wordpress.joosua.me' --quiet
$ wp cache flush
Success: The cache was flushed.

$ wp option get home
https://wordpress.joosua.me
$ grep -F htdocs/wp-config.php -e DOMAIN_CURRENT_SITE 
define( 'DOMAIN_CURRENT_SITE', 'wordpress.joosua.me' );
```

### DIFF
--- a/modules/wp-cli.php
+++ b/modules/wp-cli.php
@@ -40,4 +40,70 @@ class Seravo_WP_CLI extends \WP_CLI_Command {
   }
 }
 
+/**
+ * Extends wp-cli search-replace feature.
+ *
+ * Will be called after running regular search-replace,
+ * even if --skip-plugins is specified.
+ */
+function search_replace_extension() {
+  // Positionals <old> <new> [<table>...]>
+  $positionals = \WP_CLI::get_runner()->__get('arguments');
+  $options = \WP_CLI::get_runner()->__get('assoc_args');
+
+  if ( count($positionals) != 3 ) {
+    // User gave specific tables to run search-replace on,
+    // nothing else should be replaced
+    return;
+  }
+
+  $from = $positionals[1];
+  $to = $positionals[2];
+
+  $dry_run = \WP_CLI\Utils\get_flag_value($options, 'dry-run');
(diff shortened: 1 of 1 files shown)


sjaks (approved):
Not affected:
image
Affected:
image

Seems to change the domain to correct value so that the correct domain is used everywhere.
Dry-run and regex exclusions work and a git commit is created for changes.

LGTM!

modules/wp-cli.php
```diff
@@ -40,4 +40,70 @@ public function updates( $args, $assoc_args ) {
...
+          file_put_contents($wp_config, $new_content);
+
+          if ( is_executable('/usr/local/bin/s-git-commit') && file_exists('/data/wordpress/.git') ) {
+            // Commit wp-config.php changes
+            exec('cd /data/wordpress/ && git add htdocs/wp-config.php && /usr/local/bin/s-git-commit -m "Update DOMAIN_CURRENT_SITE" && cd /data/wordpress/htdocs/wordpress/wp-admin');
+          }
+        }
+      }
+    }
+  }
```
elguitar, line 105: To avoid multiple nested ifs, where we go deeper and deeper, `if something; return` -pattern could be used instead.

JoosuaKoskinen: - [x] Less nested ifs

elguitar (commented):
LGTM!

How should we document this? Currently the user will not expect that the search-replace will do stuff that it would not normally do.

Issue #289, opened 2019-10-10
Shadow banners need some UX design


ottok:
Our current shadow banners need some UI design:
![image](https://user-images.githubusercontent.com/668724/66569839-0b03a500-eb75-11e9-8509-f98a33f4bca5.png)

What does the 'STAGING' actually tell the user? And the comment in the parenthesis does not add much value. We should have some UX design here to get this a smooth experience.

The same `NAME (comment)` pattern is in the warning banner:
![image](https://user-images.githubusercontent.com/668724/66569939-3be3da00-eb75-11e9-814c-0ea3f6a67681.png)

JoosuaKoskinen:
Now that `WP_ENV_COMMENT` is fixed, I think we could use just that. I don't see how the comment doesn't add much value. It's often more informative than "Testivarjo".

The comment is usually "Staging" or one requested by the customer. I agree that it could be improved, "STAGING (Staging)" is not useful for the user.

PR #522, opened 2021-04-21
Improve shadow banners (Closes: #289)

AatuYH:
Removes some of the unnecessary information from the banner that is shown when inside a shadow. Now the user only sees the shadow comment instead of the previous "staging(comment)".


Moppa5 (approved):
I have no shadows myself and can't deploy one easily. However, this is just a minor modification and implemented as designed on #289. I think you can feel free to merge this.

PR #524, opened 2021-05-12
Make purge cache notification dismissible again

Moppa5:
Excessive explicit buttons for dismissing purge cache notification caused not being able to actually dismiss the notification. As admin-notices implement the dismissing itself there's no need to explicitly add any buttons etc.


Moppa5 (commented):
Merging this after self-review as it's just a minor modification to fix the notification.

Issue #523, opened 2021-05-07
GUI for running wp-db-optimize


Moppa5:
**Is your feature request related to a problem? Please describe.**
We already have a widget / GUI for `wp-db-cleanup` on database page for cleaning some excessive database cruft and transients etc. I couldn't help but to think that could we have `wp-db-optimize` as well?

**Describe the solution you'd like**
Basically GUI for running the command `wp-db-optimize` would be fine. This could be integrated with the database cleanup feature that we have already implemented as they both have pretty similar functions (cleaning and optimizing). A separate widget works as well. 

What do you think, @Seravo/dev-team?

PR #529, opened 2021-06-01
Add GUI for wp-db-optimize (Closes: #523)

Moppa5:

Extend seravo plugin database tools by adding GUI for `wp-db-cleanup`. Widget is implemented so that the old database cleanup and the new optimization share the same postbox with own sections. This can be considered as extended feature for cleaning up the database. 

See #523 

#### Screenshots
Screenshot


Moppa5 (commented):
I fixed some localization issues and tested the feature in general. Merging this after a self-review as the functionality and changes are quite small. Possible improvements etc. can be done when refactoring and enhancing Seravo Plugin.

PR #530, opened 2021-06-02
Fix HTTP stats latest months logic

Moppa5:
WP-admin dashboard widget with HTTP stats had a bug in the logic. Instead of showing the latest 4 months as intended, it was showing the first 4 months of the year. This is just a minor fix. 

For the widget see #5


Moppa5 (commented):
Merging this after self-review as the changes only affect on the order rather than any other logic. It's quick fix.

PR #533, opened 2021-06-07
Check that filepath is valid when reading logs

Moppa5:
Ensure that logfiles can only be read under `/data/log` directory.


JoosuaKoskinen (commented):
This could be just: 
```
$valid = in_array($filepath, glob('/data/log/*));
$f = $valid ? @fopen($filepath, 'rb') : false;
```

No need for looping. LGTM anyways, feel free to merge!

Moppa5:
@JoosuaKoskinen, Good point! Converted the code to follow up that example as it's much more clearer / cleaner.

PR #546, opened 2021-06-16
Add common css file

Moppa5:
Add common css file that contains dashboard error widgets, tooltip and "Toggle more" wrapper related css styles. Result wrapper css might need re-working when applying the styles for the new design. 

There are still some files, like `password.css`, that contain only a few styles. Should we move those under the common as well?

### DIFF
--- a/style/common.css
+++ b/style/common.css
@@ -1,1 +1,113 @@
-.class {}
+/*
+ * Description: Common css file for any Seravo Plugin page to apply
+ * styles for postboxes, widgets etc.
+ */
+
+ /* Dashboard widgets and postboxes */
+
+  div#seravo-error-widget.postbox,
+  div#seravo-low-disk-space-widget.postbox {
+      border-left: 0.2em solid #e8ba1b;
+  }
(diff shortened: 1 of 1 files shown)


style/common.css
```diff
@@ -1 +1,115 @@
-.class {}
+/*
+ * Description: Common css file for any Seravo Plugin page to apply
+ * styles for postboxes, widgets etc.
+ */
+
+ .class {}
```
JoosuaKoskinen, line 6: This can be removed. It was there just because PHPCS doesn't allow empty files.

JoosuaKoskinen (changes requested):

JoosuaKoskinen:
> There are still some files, like password.css, that contain only a few styles. Should we move those under the common as well?

In my opinion, only CSS needed for pages under Tools menu should be included in common.css.

Moppa5:
Merging this as the .class {} has been removed. We can work with removing and moving css to the common file when we are converting the widgets individually.

Issue #538, opened 2021-06-11
Create helpers and unify the style


Moppa5:
**Description**
As discussed in an internal issue the Seravo Plugin has many ways of creating different components like texts for instance. We should discuss, design and implement a way to create these more dynamically. There were ideas considering some sort of JS library for creating JavaScript and AJAX posts more dynamically. We should investigate the CSS styling in Seravo Plugin as well since there could most likely be only one commons CSS for all the pages. 

These changes are sort of pending on the new design after cleanup, but we need to plan these a little beforehand.

Moppa5:
@JoosuaKoskinen, Could you clarify & wrap up here what kind of helpers should be implemented after the #545 is merged?

JoosuaKoskinen:
Some of the helpers needed under template.php:

[ ] paragraph
[ ] title
[ ] button
[ ] options field 
![settings](https://user-images.githubusercontent.com/42264063/122210851-c2a29b80-ceae-11eb-87a3-00d316060985.png)
[ ] result wrapper
![result](https://user-images.githubusercontent.com/42264063/122210992-ea91ff00-ceae-11eb-86e2-46d16b3befcb.png)
[ ] some basic table
![table](https://user-images.githubusercontent.com/42264063/122211118-0ac1be00-ceaf-11eb-9fed-3fac0215c496.png)
[ ] slider thingy (Tools >  Upkeep > Screenshots)
[ ] success/failure (with bool as input)
![success](https://user-images.githubusercontent.com/42264063/122211376-55dbd100-ceaf-11eb-8d75-5527440ebecf.png)
[ ] the spinner gif

They should have options to give id and class too if it might be needed.

Moppa5:
Options field needs some thinking and might be the best to be implemented when converting a widget that uses it. There's quite good guide here https://developer.wordpress.org/plugins/settings/settings-api/ available.

Moppa5:
@JoosuaKoskinen, Are you working with the wrapper thing? I would say that it's something we need next as there are quite many widgets using it.

Moppa5:
Closing this issue as most of the helpers are already created and those that are needed in future will therefore be created as well.

PR #547, opened 2021-06-16
Add multiple helpers to templates

Moppa5:
Helpers include this far: 
- Paragraph
- Text / div field
- Section title
- Tooltip
- Spinner
- Button
- Success/Failure

Check that the created components are ok and they meet our requirements. 

See #538

### DIFF
--- a/lib/postbox/template.php
+++ b/lib/postbox/template.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * File for Seravo postbox component templates.
  */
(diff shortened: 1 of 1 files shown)


lib/postbox/template.php
```diff
@@ -23,5 +24,84 @@ public static function error_paragraph( $message ) {
...
+      );
+    }
+
+    /**
+     * Display spinner image. In default it's hidden.
+     * @param string $id Id for this spinner.
+     * @return \Seravo\Postbox\Component Spinner component.
+     */
+    public static function spinner( $id ) {
+      return Component::from_raw('<div id="' . $id . '"><img src="/wp-admin/images/spinner.gif" hidden></div>');
```
JoosuaKoskinen, line 84: Could you remove the `hidden` and add option to add classes. Then you can make hidden spinner by giving `hidden` class and visible spinner by giving it nothing. Currently you can only make hidden spinners.

lib/postbox/template.php
```diff
@@ -23,5 +24,84 @@ public static function error_paragraph( $message ) {
...
+        $color = 'green';
+        $msg = sprintf('<b>%s</b>', __('Success!', 'seravo'));
+      } else {
+        $color = 'red';
+        $msg = sprintf('<b>%s</b>', __('Failure!', 'seravo'));
+      }
+
+      return Component::from_raw('<p style="color: ' . $color . ';">' . $msg . '</p>');
+
+    }
```
JoosuaKoskinen, line 104: Nothing wrong with this but maybe remove the first and last empty lines for consistency.

JoosuaKoskinen (changes requested):

JoosuaKoskinen:
Feel free to merge after resolving change requests.

lib/postbox/template.php
Moppa5: Added $class = 'spinner' default parameter and the class attribute for the div.

lib/postbox/template.php
Moppa5: Removed. The first empty line was due to the condition block as I like it better. However, they are both now removed.

PR #549, opened 2021-06-17
Add Ajax handler and postbox for auto-command

JoosuaKoskinen:

Adds AJAX handler and pre-made postbox for automatically sending AJAX request and running a command on page load.

Also fixes a couple of bugs.

Example usage here:
https://github.com/Seravo/seravo-plugin/blob/postbox/upgrade-backups-page/modules/backups.php#L35-L53

### DIFF
--- a/lib/postbox/ajax.php
+++ b/lib/postbox/ajax.php
@@ -151,8 +151,8 @@ if ( ! class_exists('Ajax_Handler') ) {
           if ( false === $response ) {
             // The data was not cached, call data_func
             $response = \call_user_func($this->ajax_func, $this->section);
-            if ( null !== $response ) {
-              // Cache new result unless it's null
+            if ( null !== $response && isset($response->data['success']) && $response->data['success'] === true ) {
+              // Cache new result unless it's invalid
               \set_transient($cache_key, $response, $this->data_cache_time);
             }
           }
@@ -193,7 +193,7 @@ if ( ! class_exists('Ajax_Response') ) {
     /**
      * @var bool[]|string[]|mixed[]|mixed Data to respond with.
      */
-    private $data = array();
+    public $data = array();
 
     public function __construct() {
       $this->data['success'] = false;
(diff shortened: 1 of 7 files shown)


lib/postbox/ajax.php
```diff
@@ -287,5 +287,131 @@ public static function exception_response() {
...
+     * @param bool $allow_failure Whether exit code other than 0 should respond with an error.
+     */
+    public function __construct( $section, $command = null, $cache_time = 300, $allow_failure = false ) {
+      parent::__construct($section);
+
+      $this->command = $command;
+      $this->allow_failure = $allow_failure;
+
+      $this->set_build_func(
+        function ( \Seravo\Postbox\Component $base, $section ) {
```
Moppa5, line 342: Here again, are this some rector related case?

rector.php
```diff
@@ -211,7 +211,7 @@
 
   // Type Declaration
   $services->set(\Rector\TypeDeclaration\Rector\ClassMethod\AddArrayReturnDocTypeRector::class);
-  $services->set(\Rector\TypeDeclaration\Rector\Property\CompleteVarDocTypePropertyRector::class);
+  //$services->set(\Rector\TypeDeclaration\Rector\Property\CompleteVarDocTypePropertyRector::class); Rector being Rector
```
Moppa5, line 214: Mystical comments :D

Moppa5 (changes requested):
Great job! I didn't test the new stuff quite extensively but test widgets on backup page seem to work normally and there are no major flaws. Possible arising issues can be fixed when converting the widgets as well.

However, I really think we should reconsider where we put the `extends PostBox` and similar. If we have for example 5 ajax commands, the file becomes quickly quite large and it gets difficult to grasp the contents of the file. I don't think there's any issue for example having a individual file for every module that extends Postbox or Ajax_Handler. What do you think about this? Personally I think it's not good in the terms of maintainability.

lib/postbox/component.php
```diff
@@ -53,9 +53,9 @@ public function add_children( $children ) {
       }
 
       foreach ( $children as $child ) {
-        if ( $child instanceof Component ) {
-          $this->children[] = $child;
-        }
+        // Add them with add_child so same checks
+        // are used for each child.
+        $this->add_child($child);
```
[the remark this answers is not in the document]
JoosuaKoskinen: It is and I still call it false positive. But it was quite useless check so removed it to make Rector happy.

rector.php
JoosuaKoskinen: Fixed

lib/postbox/ajax.php
JoosuaKoskinen: Fixed

lib/postbox/ajax.php
```diff
@@ -287,5 +287,131 @@ public static function exception_response() {
...
+      $response->set_error($error);
+      return $response;
+    }
+
+  }
+}
+
+namespace Seravo\Postbox\Ajax;
+
+use \Seravo\Postbox\Ajax_Handler;
```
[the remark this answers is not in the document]
JoosuaKoskinen: Fixed.

Moppa5 (approved):

PR #551, opened 2021-06-17
Move PHP files under src/

JoosuaKoskinen:

- Unify including of php, css and js files. They now use absolute paths. We need to make sure that doesn't cause issues.
- Move PHP files under new src/ folder.

Refactoring of Seravo Plugin. Src/ folder is the correct way to do things and not using it makes a lot of things difficult.


rector.php
```diff
@@ -16,13 +16,13 @@
   $parameters->set(
     Option::PATHS,
     array(
-      __DIR__ . '/lib',
-      __DIR__ . '/modules',
+      __DIR__ . 'src/lib',
+      __DIR__ . 'src/modules',
```
Moppa5, line 20: These causes issue when running `composer test`. Should they be changed to `/src/lib` and `/src/modules`, this seems to fix the issue. 

```
(seravo-plugin-2/src-folder) santeri@SakkeUbuntu:~/Desktop/seravo-plugin$ composer test
> vendor/bin/rector --dry-run

In Finder.php line 530:
                                                                              
  The "/home/santeri/Desktop/seravo-pluginsrc/lib" directory does not exist.  
                                                                              

process [-n|--dry-run] [-a|--autoload-file AUTOLOAD-FILE] [-o|--output-format [OUTPUT-FORMAT]] [--no-progress-bar] [--no-diffs] [--output-file OUTPUT-FILE] [--cache-debug] [--clear-cache] [--] [<source>...]

Script vendor/bin/rector --dry-run handling the test event returned with error code 1
```

rector.php
```diff
@@ -16,13 +16,13 @@
...
       __DIR__ . '/rector.php',
       __DIR__ . '/seravo-plugin.php',
     )
   );
-  $parameters->set(Option::SKIP, array( __DIR__ . '/lib/list-table.php' ));
+  $parameters->set(Option::SKIP, array( __DIR__ . 'src/lib/list-table.php' ));
```
Moppa5, line 25: Note this as well, should it be prefixed with `/src` instead?

Moppa5 (changes requested):
Great job! I didn't find any major issues when testing the new codebase. Running `composer test` gives quite a lot of stuff about `src/lib/list-table.php`. Should you fix it for this PR as well?

rector.php
JoosuaKoskinen: Fixed

rector.php
JoosuaKoskinen: Fixed

Moppa5 (approved):

Issue #543, opened 2021-06-13
Class loading by Composer autoloader


szepeviktor:
As Composer autoloader is given please load Seravo classes with it.

All the best!

Moppa5:
Thanks for the suggestion! We will be looking into this during the cleanup and refactoring.

szepeviktor:
You're welcome.

You get all my brains as a bonus :) https://github.com/szepeviktor/small-project/blob/master/MAIN-FILE-PARTS.md

Issue #537, opened 2021-06-11
Reorganize Seravo Plugin folder structure


Moppa5:
**Description**
As a part of the cleanup and Up-To-Date project one point was about Seravo Plugin folder structure. We should investigate and reorganize a proper and cleaner folder structure for Seravo Plugin. There are no clear plans yet for this as it's depending mostly on how the new implementation & design will be after the cleanup. 

However, we can discuss and document here about the changes.

JoosuaKoskinen:
@Moppa5 and I decided the folder structure for postboxes to be the following:
```
/lib
   |
   |__ /postbox
   |          |__ postbox.php
   |          |__ toolpage.php
   |          |__ handler.php
   |          |__ /component
   |          |            |__ component.php
   |          |            |__ templates.php
   |          |__ /templates
   |                       |__ auto_command.php ....
   |
   |__ /ajax
   |       |_____ handler.php
   |       |_____ response.php
           |_____ /templates
                           |__ auto_command.php ....
```

PR #552, opened 2021-06-17
Use Composer autoloading

JoosuaKoskinen:

Adds Composer autoloading. Only autoloads the new postbox classes under src/lib/postbox for now. The rest of the classes under /src/lib can be autoloaded later when the code allows for it.  

The two extra commits are because this change relies on https://github.com/Seravo/seravo-plugin/pull/551. Can be merged after https://github.com/Seravo/seravo-plugin/pull/551.

Part of https://github.com/Seravo/seravo-plugin/issues/543 and https://github.com/Seravo/seravo-plugin/issues/537

#### Manual testing steps?
- See that this code only works if `require_once SERAVO_PLUGIN_DIR . 'vendor/autoload.php';` is in seravo-plugin.php:
`new \Seravo\Postbox\Postbox::new('');`
- If autoloading doesn't work, a fatal error prevents any page from loading.


Moppa5 (approved):
Autoloading and changes in general look good to me! Rector / PHPStan "Access denied!" will be worked in another PR. Feel free to merge.

PR #553, opened 2021-06-18
Restructure postboxes and AJAX

JoosuaKoskinen:

- Organize postbox files better, separate ajax and postbox
- Improve namespaces
- Only one class per file

Temporarily moved global nonsense in new postbox-init.php.

**WIP**: Will update and improve the comments (file header comments -> class comments) a bit.

Part of https://github.com/Seravo/seravo-plugin/issues/537 and postbox refactori

Again, bit hard to review but the possible bugs will be found during postbox refactoring. Tested myself that all the classes are still available in the correct namespace.

It would be good to go through the postbox/ajax files and see if there's doc comments with old namespaces etc eg. `@param \Seravo\Postbox\Ajax_Handler`.


Moppa5 (approved):
Seems great! Structure is as discussed and no major flaws. As already mentioned we can work with possible issues later as the whole new design is still prone to change some parts. Feel free to merge.

PR #554, opened 2021-06-18
Use autoload on classes api, helpers & list-table (Related: #543)

Moppa5:
Use Composer autoload on more classes.


src/lib/api.php
```diff
@@ -1,74 +1,68 @@
...
-      $site = getenv('USER');
-      $ch = curl_init('http://localhost:8888/v1/site/' . $site . $api_query);
-      curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
-      curl_setopt($ch, CURLOPT_HTTPHEADER, array( 'X-Api-Key: ' . getenv('SERAVO_API_KEY') ));
-      $response = curl_exec($ch);
-      $httpcode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+/**
+ * Class API
+ *
+ * File for accessing and modifying site-related data.
```
JoosuaKoskinen, line 8: Maybe change the `File` to `Class` as this is the doc for the class.

JoosuaKoskinen (approved):
LGTM otherwise!

PR #562, opened 2021-06-23
Rewrite database page widgets (Related: 544)

Moppa5:
Rewrite 3/5 basic widgets on database page
- Database access
- Adminer link & info
- Database cleanup tools
- Misc changes/additions to helpers

I also added separate `$dryrun_command` as for example `wp-db-cleanup` needs to run `wp-backup` as well whereas with `--dry-run` it must not. This way any dry run command can be separated from the real execution command.


src/lib/ajax/templates/command-runner.php
```diff
@@ -63,17 +68,19 @@ function ( $section ) {
    * @return \Seravo\Ajax\AjaxResponse Response for the client.
    */
   public function ajax_command_exec( $section ) {
+    $exec_command = $this->command;
+
     if ( $this->command === null ) {
```
JoosuaKoskinen, line 73: Could you add these checks (and respond with error):
- If `is_dryrun_enabled()`, `$this->dryrun_command` cant be null
- If `$_GET['dryrun']` is true, `is_dryrun_enabled` must be true (I should have done this)

src/lib/ajax/templates/command-runner.php
```diff
@@ -107,6 +114,14 @@ public function set_command( $command, $cache_time = 300, $allow_failure = false
     $this->set_cache_time($cache_time);
   }
 
+  /**
+   * Set the special dry run command to be executed.
+   * @param string $command Command for exec.
+   */
+  public function set_dryrun_command( $command ) {
```
JoosuaKoskinen, line 121: Maybe `enable_dryrun` and `dryrun` variable should be removed now that we have dryrun_command? Bit unnecessary to write `$handler->set_dryrun_command(...); $handler->enable_dryrun(true);`.

We can just check if dryrun_command is null to see if dryrun whether dryrun is enabled.

src/lib/helpers.php
```diff
@@ -98,4 +98,31 @@ public static function sanitize_full_path( $file ) {
     return implode('/', $path);
   }
 
+  /**
+   * @return string Escaped adminer_url
+   */
+  public static function adminer_link() {
```
JoosuaKoskinen, line 104: As we talked, this could be just:
```
if ( ! self::is_production() ) {
  return esc_url(str_replace('//', '//adminer.', get_site_url()));
}
return esc_url(get_site_url(null, '.seravo/adminer'));
```

src/lib/postbox/component/template.php
```diff
@@ -46,6 +46,32 @@ public static function section_title( $content ) {
...
+  /**
+   * Generates link with icon.
+   * @param string $href Link URL.
+   * @param string $link_content Content to wrap the link around to.
+   * @return \Seravo\Postbox\Component Button with icon.
+   */
+  public static function button_link_with_icon( $href, $link_content ) {
+    return Component::from_raw(
+      '<a href="' . $href . '" class="button" target="_blank">' . $link_content . '
+      <span aria-hidden="true" class="dashicons dashicons-external" style="line-height: 1.4; padding-left: 3px;"></span> </a>'
```
JoosuaKoskinen, line 71: Is `style="line-height: 1.4; padding-left: 3px;"` something we could write in common.css instead? Inline CSS should be avoided and is one of the things I would like to get rid of with the refactoring.

src/modules/database.php
```diff
@@ -35,53 +42,110 @@ public static function load() {
...
-        'side'
+    public static function init_database_postboxes( Toolpage $page ) {
+      /**
+       * Database access info postbox
+       */
+      $database_access = new Postbox\InfoBox('database-access');
+      $database_access->set_title(__('Database Access', 'seravo'));
+      $database_access->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
+      $database_access->add_paragraph(
+        sprintf(
```
JoosuaKoskinen, line 66: In my opinion, this is too verbose. The `init_database_postboxes` is supposed to do basic postbox initialization, titles and stuff but now it's 30% infobox text. The postbox functions are designed to be very compact for the very same reason.

I don't like long lines but I think they could be a better solution. Or maybe something else, I don't know but I would prefer not to have these huge blocks of text and a lot of empty lines just having `)`, `),`, `);` etc.

src/modules/database.php
```diff
@@ -110,77 +174,63 @@ public static function enqueue_database_scripts( $page ) {
...
+        Template::paragraph(
         /* translators:
         * %1$s url to www.adminer.org
         */
-        printf(__('<a href="%1$s" target="_BLANK">Adminer</a> is a visual database management tool, which is simpler and safer than its competitor phpMyAdmin.', 'seravo'), 'https://www.adminer.org');
-        ?>
-      </p>
-      <p>
-        <?php
+          sprintf(__('<a href="%1$s" target="_BLANK">Adminer</a> is a visual database management tool, which is simpler and safer than its competitor phpMyAdmin.', 'seravo'), 'https://www.adminer.org')
```
JoosuaKoskinen, line 217: Unnecessary usage of `sprintf`. Adds the need for translators -comment which is now 3 useless lines.

src/modules/database.php
```diff
@@ -110,77 +174,63 @@ public static function enqueue_database_scripts( $page ) {
...
-          $siteurl .= '/';
-        }
-
-        // Inject subdomain
-        $adminer_url = str_replace('//', '//adminer.', $siteurl);
-
-      }
-
-      ?>
+          sprintf(__('At Seravo it can always be accessed at %1$s.', 'seravo'), '<code>example.com/.seravo/adminer</code>')
```
JoosuaKoskinen, line 226: Unnecessary sprintf

src/modules/database.php
```diff
@@ -110,77 +174,63 @@ public static function enqueue_database_scripts( $page ) {
...
-        <a href="<?php echo esc_url($adminer_url); ?>" class="button" target="_blank">
-          <?php _e('Open Adminer', 'seravo'); ?>
-          <span aria-hidden="true" class="dashicons dashicons-external" style="line-height: 1.4; padding-left: 3px;"></span>
-        </a>
-      </p>
-      <?php
+      $adminer_url = Helpers::adminer_link();
+      $button_wrapper = new Component('', '<p class="adminer_button">', '</p>');
+      $button_wrapper->add_child(Template::button_link_with_icon($adminer_url, __('Open Adminer', 'seravo')));
+      $base->add_child($button_wrapper);
```
JoosuaKoskinen, line 233: Each component has wrapper, content and possibly children. This $button_wrapper not has $wrapper and children, no content. Could be just:
```
$button = Template::button_link_with_icon(Helpers::adminer_link(), __('Open Adminer', 'seravo'));
$button->set_wrapper('<p class="adminer_button">', '</p>');
$base->add_child($button);
```

src/modules/database.php
```diff
@@ -35,53 +42,110 @@ public static function load() {
...
+      $page->register_postbox($adminer);
+
+      /**
+       * Search & Replace tool postbox
+       * To be implemented...
+       */
+      $search_replace = new Postbox\Postbox('database-search-replace');
+      $search_replace->set_title(__('Search-Replace Tool', 'seravo'));
+      $search_replace->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
+      $search_replace->set_build_func(array( __CLASS__, 'database_search_replace_postbox' ));
```
JoosuaKoskinen, line 105: You added these old functions in to postbox with new design. That's not how they are used and there's a change this causes problems.

If it doesn't cause problems, I guess it's okay as this is to be implemented.

src/modules/database.php
```diff
@@ -35,53 +42,110 @@ public static function load() {
...
-
+      $page->register_postbox($database_access);
+
+      /**
+       * Database adminer info postbox
+       */
+      $adminer = new Postbox\Postbox('database-adminer');
+      $adminer->set_title(__('Manage the Database with Adminer', 'seravo'));
+      $adminer->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
+      $adminer->set_build_func(array( __CLASS__, 'database_adminer_postbox' ));
```
JoosuaKoskinen, line 95: Postbox functions used to be named `database_adminer_postbox` etc because that function was the whole postbox. Now the function is only used to build the components for it and is only just a small part of the postbox functionality.

= would prefer `build_adminer_postbox` etc

src/modules/database.php
```diff
@@ -110,77 +174,63 @@ public static function enqueue_database_scripts( $page ) {
...
-      <?php
+    /**
+     * Build the database cleanup and optimize postbox.
+     * @param Component $base The postbox base component.
+     * @param Postbox $postbox To fetch the AJAX components from.
+     */
+    public static function database_cleanup_tools_postbox( Component $base, Postbox\Postbox $postbox ) {
+      $base->add_child(Template::section_title(__('Optimization', 'seravo')));
+      $base->add_child(
+        Template::paragraph(
```
JoosuaKoskinen, line 185: This is the absolutely correct way to build postboxes, awesome! :+1:

Still dont like the paragraph verbosity, maybe we should think if we can come up with something better?

src/modules/database.php
```diff
@@ -35,53 +42,110 @@ public static function load() {
...
+      $cleanup = new Postbox\Postbox('database-cleanup');
+      $cleanup->set_title(__('Database Cleanup Tool', 'seravo'));
+      $cleanup->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
+      $cleanup->set_build_func(array( __CLASS__, 'database_cleanup_tools_postbox' ));
+
+      // Initialize optimize section
+      $optimize = new ButtonCommand('optimize-db');
+
+      // define the db credentials for wp-db-optimize
+      $db_info = defined('DB_NAME') ? 'DB_NAME=' . DB_NAME . ' ' : '';
```
JoosuaKoskinen, line 120: I dont like that this function has so much stuff not related to the postbox initialization. I think max 10 lines for each postbox would be ideal. I know that in some cases it's not achievable but in this case I would do ajax initialization in other function.

JoosuaKoskinen (changes requested):

src/lib/ajax/templates/command-runner.php
```diff
@@ -63,17 +68,19 @@ function ( $section ) {
...
   public function ajax_command_exec( $section ) {
+    $exec_command = $this->command;
+
     if ( $this->command === null ) {
       return AjaxResponse::unknown_error_response();
     }
 
     if ( $this->is_dryrun_enabled() && isset($_GET['dryrun']) && $_GET['dryrun'] === 'true' ) {
-      $this->command = $command . ' --dry-run';
+      $exec_command = $this->dryrun_command . ' --dry-run';
```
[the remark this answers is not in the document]
Moppa5: I was also thinking this before making the PR. I will change the `--dry-run` logic to apply this.

src/lib/ajax/templates/command-runner.php
Moppa5: I will add a check for the dryrun_command not being null.

src/lib/ajax/templates/command-runner.php
Moppa5: I will add a check for the dryrun_command being null and error response.

src/lib/ajax/templates/command-runner.php
Moppa5: Agreed 👍🏼 . I will modify this.

src/lib/helpers.php
Moppa5: I don't remember that we talked about an exact solution and nevertheless the whole code was a bit weird to me. However, I will modify this.

src/lib/postbox/component/template.php
Moppa5: Good point! I will investigate and convert them to common.css

src/modules/database.php
Moppa5: I will fix this 👍🏼

src/modules/database.php
Moppa5: The old widgets are working just fine and I've tested them multiple times. I converted the implementation so that it's easy to start implementing them with the new design right away.

src/modules/database.php
Moppa5: I agree. I will change the postbox methods to build and perhaps rephrase them in general.

src/modules/database.php
Moppa5: Moved these to a helper method to not confuse too much stuff on the main init function.

JoosuaKoskinen (approved):
Awesome!

PR #563, opened 2021-06-23
Misc improvements to template helpers

Moppa5:
- Add optional class parameters
- Use common.css instead of styles inside element

### DIFF
--- a/style/common.css
+++ b/style/common.css
@@ -32,10 +32,20 @@
 
  /* Dashicon styles */
 
- .dashicons .dashicons-external {
-  line-height: 1.4;
-  padding-left: 3px;
- }
+  .dashicons .dashicons-external {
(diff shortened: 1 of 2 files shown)


style/common.css
```diff
@@ -30,6 +30,16 @@
       border-left: 0.2em solid #e8ba1b;
   }
 
+ /* Misc common elements */
+
+ p.success {
```
JoosuaKoskinen, line 35: Are these new classes indented different from others?

JoosuaKoskinen (changes requested):
Feel free to merge after resolving indents

style/common.css
Moppa5: Fixed also the new added dashicons stuff as they were intended differently as well.

Issue #544, opened 2021-06-14
Rewrite widgets with the new design and track down the progress


Moppa5:
**Description**
As a part of the Up-To-Date project it was mentioned that we could convert Seravo Plugin widgets one by one to apply the new design. This issue is here just so that we can reference and keep track of the rewrite process. 

**Sitestatus**
- [ ] Cache Status, contains "Toggle more" widget
- [ ] Site Information
- [ ] HTTP Request statistics
- [ ] Site checks, if possible ?
- [ ] Speed test
- [ ] Sanitize uploads
- [ ] Disk Usage
- [ ] Optimize Images
- [ ] Shadows

**Upkeep**
- [ ] Update Status
- [ ] Tests Status
- [ ] Changes Status
- [ ] Update tests
- [ ] Seravo Updates
- [ ] Screenshots ?
- [ ] Seravo Plugin Updater
- [ ] Change PHP Version

**Database**
- [ ] Database Access
- [ ] Manage the Database with Adminer
- [ ] Search-Replace Tool
- [ ] Database Cleanup Tool (optimize & cleanup)
- [ ] Database Size

**Backups**
- [ ] Backups, basic box
- [ ] Create a New Backup
- [ ] Files Excluded from the Backups
- [ ] Current Backups

**Security**
- [ ] Security (options)
- [ ] Last successful logins
- [ ] Check passwords
- [ ] Unnecessary themes
- [ ] Cruft Files
- [ ] Unnecessary plugins

**Domains & Logs*
- [ ] Domains & Mails bundled
- [ ] Logviewer ?

JoosuaKoskinen:
I think it's currently possible to rewrite these:
- [ ] Site Status -> Site information
- [ ] Upkeep -> Update Status (probably) 
- [ ] Upkeep -> Test Status
- [ ] Database -> Database Access
- [ ] Database -> Manage the Database with Adminer
- [ ] Backups -> Backups

After https://github.com/Seravo/seravo-plugin/pull/549 these should be possible:
- [ ] Backups -> Files Excluded from the Backups
- [ ] Backups -> Current Backups
- [ ] Security -> Last successful logins


There will be a second AJAX handler and 'options/settings' upcoming very soon, I believe over 50% of postboxes can be rewritten then. And after possibility to add different kinds of graphs and tables I expect it to be at least 3/4 postboxes.

Moppa5:
@JoosuaKoskinen, I can take a look at the following widgets next
- Site Information
[...]

PR #567, opened 2021-06-30
Rewrite some security page widgets

Moppa5:
Rewrite check passwords and last successful logins postboxes. Also add template for creating HTML tables. Add a description text and link to `wp-login.log` to last successful logins postbox as well. 

Related to #544


src/modules/security.php
```diff
@@ -91,7 +86,137 @@ public static function load() {
...
+      $page->register_postbox($passwords);
+
+      /**
+       * Last successful logins postbox
+       */
+      $logins = new Postbox\Postbox('logins-info');
+      $logins->set_title(__('Last successful logins', 'seravo'));
+      $logins->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
+
+      $last_logins = new Ajax\LazyLoader('retrieve-logins', 0);
```
Moppa5, line 115: Probably should have / put some cache time for these. I will check and apply these together with possible other change requests.

src/lib/postbox/component/template.php
```diff
@@ -70,6 +70,35 @@ public static function list_view( $element ) {
...
+    foreach ( $column_titles as $title ) {
+      $titles->add_child(Component::from_raw('<th class="' . $th_class . '">' . $title . '</th>'));
+    }
+    $main_table->add_child($titles);
+
+    foreach ( $all_rows as $rows ) {
+      $row = new Component('', '<tr>', '</tr>');
+
+      foreach ( $rows as $row_element ) {
+        $row->add_child(Component::from_raw('<td class="' . $td_class . '">' . $row_element . '</td>'));
```
JoosuaKoskinen, line 95: The cells used to have tooltips on hover. They need to be added back as most of the strings are cut and that's the only way to see them.

src/modules/security.php
JoosuaKoskinen: Yep, cache for 10 mins should be good, this isn't data that's supposed to be real time.

Also, could you create `Postbox\LazyLoader` and use that instead of `Postbox\Postbox` with `Ajax\LazyLoader`. It hasn't been created just because it hadn't been used yet. It basically just needs to be like the other premade Postboxes but only functions needed is constructor and:
```
public function set_ajax_func( $ajax_func ) {
  $handler = $this->ajax_handlers[$this->id];
  $handler->set_ajax_func($ajax_func);
}
```

style/common.css
```diff
@@ -46,6 +46,21 @@
       border-left: 0.2em solid #e8ba1b;
   }
 
+  .login_info_table {
```
JoosuaKoskinen, line 49: I don't think common.css should have postbox specific classes.

In current seravo-plugin-2/master, there's something like that already: https://github.com/Seravo/seravo-plugin/blob/seravo-plugin-2/master/style/common.css#L72

style/common.css
```diff
@@ -46,6 +46,21 @@
...
+    table-layout: fixed;
+    width: 100%;
+  }
+
+  .login_info_th {
```
JoosuaKoskinen, line 54: Fix the one above and then you can use `.result-table th {}`. Not good practice to have `_element` like that.

src/modules/security.php
```diff
@@ -91,7 +86,137 @@ public static function load() {
...
+     * Init postboxes on Security page.
+     * @param Toolpage $page Page to init postboxes.
+     */
+    public static function init_security_postboxes( Toolpage $page ) {
+      /**
+       * Check passwords postbox (Beta)
+       */
+      $passwords = new Postbox\SimpleCommand('check-passwords');
+      $passwords->set_title(__('Check passwords (Beta)', 'seravo'));
+      $passwords->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
```
JoosuaKoskinen, line 101: I think this Postbox and the one below `CAN_BE_ANY_ENV`, not just in production.

src/modules/security.php
```diff
@@ -28,8 +35,6 @@ public static function load() {
       add_action('admin_init', array( __CLASS__, 'register_security_settings' ));
       add_action('admin_enqueue_scripts', array( __CLASS__, 'register_security_scripts' ));
 
-      add_action('wp_ajax_seravo_security', 'Seravo\seravo_ajax_security');
-
       add_action('wp_ajax_seravo_check_passwords', 'seravo_ajax_check_passwords');
```
JoosuaKoskinen, line 38: Make sure you remove all old code not needed anymore, this could cause issues.

style/common.css
```diff
@@ -46,6 +46,21 @@
...
+    text-align: left;
+  }
+
+  .seravo-tooltip {
```
JoosuaKoskinen, line 58: There's replacement for this is master: https://github.com/Seravo/seravo-plugin/blob/seravo-plugin-2/master/style/common.css#L133

src/modules/security.php
```diff
@@ -91,7 +86,137 @@ public static function load() {
...
+      }
+      // Re-index the array after unsetting invalid lines
+      $login_data = array_values($login_data);
+
+      $response = new AjaxResponse();
+      $response->is_success(true);
+
+      if ( empty($login_data) ) {
+        $response->set_error(__('No login data available', 'seravo'));
+        $response->is_success(false);
```
JoosuaKoskinen, line 206: `is_success(false)` and `set_error` is meant for actually errors. In this there just wasn't anything to show.

JoosuaKoskinen (changes requested):

src/modules/security.php
Moppa5: Will do.

src/modules/security.php
Moppa5: Converted to use set_data and return response normally.

style/common.css
Moppa5: Agreed. Will convert.

src/lib/postbox/component/template.php
Moppa5: Absolutely will add them back. The title thing was actually something that I forgot to check afterwards.

src/lib/postbox/templates/lazy-loader.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Class LazyLoader
+ *
+ * LazyLoader is a pre-made postbox for executing
+ * command with spinner and showing the output.
+ */
+class LazyLoader extends InfoBox {
+
+  /**
+   * Constructor for SimpleForm. Will be called on new instance.
```
JoosuaKoskinen, line 16: Old comment

src/lib/postbox/templates/lazy-loader.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Seravo\Postbox;
+
+use \Seravo\Ajax;
+
+/**
+ * Class LazyLoader
+ *
+ * LazyLoader is a pre-made postbox for executing
+ * command with spinner and showing the output.
```
JoosuaKoskinen, line 11: Old comment, not executing command

JoosuaKoskinen (changes requested):

JoosuaKoskinen:
Feel free to merge after fixing comments

Issue #465, opened 2021-01-22
Implement a way to run backend operations without timing out


elguitar:
WordPress timeouts, if there is no output received from a AJAX call.

This is a huge annoyance when running large tasks such as shadow reset.

We should find a way to periodically receive output from PHP exec function. This can possibly help: https://stackoverflow.com/questions/1281140/run-process-with-realtime-output-in-php

Moppa5:
I think the `proc_open` is the right approach for this. I managed to convert `wp-test` which is run on upkeep page to use the `proc_open` and with the given pipelines etc. for the data. There's also a good example on [php.net](https://www.php.net/manual/en/function.proc-open.php) about the usage of `proc_open`. This example also contains notes about deadlock if the pipelines aren't properly closed before `proc_close`.

However, the implementation needs more designing. For instance, do we want just a way to run these on background or having stream data between AJAX and PHP also? Either we could use this process approach when required by implementing it on heavy operations, or we could implement similar to Helpers class a way to execute a PHP process and the command assigned to it by calling a function.

elguitar:
Realtime streaming is not needed between the front and back. It could just poll every 5 seconds or something like that.

A helper method for this would be pretty great, because this feature is needed in several places.

Moppa5:
PHP processes don't seem to be the working solution solely. We should investigate and try out some RestAPI solution that can be integrated to Seravo Plugin. WordPress has its own RestAPI and we should investigate it https://developer.wordpress.org/rest-api/.

PR #570, opened 2021-07-01
Add background commands with polling

JoosuaKoskinen:

Adds possibility to run commands on background and polling to see when it has finished.

Also added a test-page which is enabled in Seravo Plugin debug mode and there's an example usage of polling.

Needed for long operations and related to https://github.com/Seravo/seravo-plugin/issues/465.

You can try the example on test-page.

### DIFF
--- a/src/modules/test-page.php
+++ b/src/modules/test-page.php
@@ -1,0 +1,90 @@
+<?php
+/**
+ * File for test postboxes page. No need to
+ * translate anything here.
+ */
+
+namespace Seravo;
+
+use \Seravo\Ajax;
+use \Seravo\Postbox;
+use \Seravo\Postbox\Toolpage;
+use \Seravo\Postbox\Requirements;
+
+// Deny direct access to this file
+if ( ! defined('ABSPATH') ) {
+  die('Access denied!');
+}
+
+if ( ! class_exists('TestPage') ) {
+  class TestPage {
+
+    /**
+     * Initialize test-page.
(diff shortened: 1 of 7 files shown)


src/modules/test-page.php
```diff
@@ -0,0 +1,90 @@
...
+     * @param \Seravo\Postbox\Toolpage $page The page for postboxes.
+     */
+    public static function init_test_postboxes( Toolpage $page ) {
+      /**
+       * Polling test postbox
+       */
+      $poller_demo = new Postbox\SimpleForm('poller-test', 'normal');
+      $poller_demo->set_title('Polling Test');
+      $poller_demo->set_button_text('Click me');
+      $poller_demo->set_button_func(array( __CLASS__, 'long_polling_operation' ));
```
Moppa5, line 45: Just a comment that pop upped to my mind. Should we rather use `set_ajax_func` named method to specify ajax execution? In my opinion it would be better to use the same naming convention. At least to me `set_ajax_func` tells better what I'm doing vs `set_button_func`. What do you think?

Moppa5 (approved):
Looking good! There's one minor comment that is not necessarily this PR specific.

src/modules/test-page.php
JoosuaKoskinen: Yep, I have thought this myself too. I agree, the different function name is a bit confusing. Will change this in upcoming PR.

PR #572, opened 2021-07-02
Rewrite some upkeep page postboxes

Moppa5:
Rewrite Seravo Plugin update and Change PHP version postboxes. Also, add radio button to templates and rewrite PHP compatibility check to use the new AJAX poller.

Also, removed some excessive AJAX and JS files that could have caused issues with the new design. These files should be fully removed when the rewrite is complete on the given page.

### DIFF
--- a/src/modules/upkeep.php
+++ b/src/modules/upkeep.php
@@ -2,11 +2,13 @@
 
 namespace Seravo;
 
-use Seravo\Postbox\Postbox;
-use Seravo\Postbox\Requirements;
-use Seravo\Postbox\Template;
+use Seravo\Ajax\ButtonCommand;
+use \Seravo\Postbox;
 use Seravo\Postbox\Component;
-use Seravo\Postbox\Toolpage;
+use \Seravo\Postbox\Template;
+use \Seravo\Postbox\Toolpage;
+use \Seravo\Postbox\Requirements;
+
 
 if ( ! defined('ABSPATH') ) {
   die('Access denied!');
@@ -21,6 +23,12 @@ if ( ! class_exists('Upkeep') ) {
       add_action('wp_ajax_seravo_ajax_upkeep', 'Seravo\seravo_ajax_upkeep');
       add_action('wp_ajax_seravo_changes_since', 'Seravo\seravo_ajax_changes_since');
       add_action('admin_post_toggle_seravo_updates', array( __CLASS__, 'seravo_admin_toggle_seravo_updates' ), 20);
+
+      $page = new Toolpage('tools_page_upkeep_page');
+      self::init_upkeep_postboxes($page);
+
+      $page->enable_ajax();
+      $page->register_page();
 
       // TODO: check if this hook actually ever fires for mu-plugins
       register_activation_hook(__FILE__, array( __CLASS__, 'register_view_updates_capability' ));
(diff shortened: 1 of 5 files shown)


JoosuaKoskinen:
> remove the old check PHP compatibility as it needs to be re-written so that it doesn't timeout
It could be rewritten already as we have to poller merged.

src/modules/upkeep.php
```diff
@@ -63,33 +67,50 @@ public static function load() {
...
-     */
     public static function init_upkeep_postboxes( Toolpage $page ) {
+      /**
+       * Seravo Plugin Updater
+       */
+      $seravo_plugin_update = new Postbox\Postbox('seravo-plugin-updater');
+      $seravo_plugin_update->set_title(__('Seravo Plugin Updater', 'seravo'));
+      $seravo_plugin_update->set_requirements(array( Requirements::CAN_BE_ANY_ENV => true ));
+
+      $update_button = new SimpleCommand('seravo-plugin-update', 'wp-seravo-plugin-update &');
```
JoosuaKoskinen, line 80: The other postboxes and AJAX handlers use namespace prefix, this one should too.
`Ajax\SimpleCommand`

src/modules/upkeep.php
```diff
@@ -187,7 +208,196 @@ public static function register_scripts( $page ) {
...
+
+    /**
+     * Build Seravo Plugin Update postbox.
+     * @param Component $base Postbox base component.
+     * @param Postbox $postbox Current postbox to build for.
+     * @param mixed $data Data returned by data function.
+     */
+    public static function build_seravo_plugin_update_postbox( Component $base, Postbox\Postbox $postbox, $data ) {
+      if ( ! isset($data['current_version']) || ! isset($data['upstream_version']) ) {
+        $base->add_child(Template::error_paragraph(__('No upstream or current Seravo Plugin version available, please try again later')));
```
JoosuaKoskinen, line 369: Translate missing domain. Add `'seravo'`.

src/modules/upkeep.php
```diff
@@ -187,7 +208,196 @@ public static function register_scripts( $page ) {
 
         wp_localize_script('seravo_upkeep', 'seravo_upkeep_loc', $loc_translation);
       }
+    }
+
+    /**
+     * Build Change PHP version postbox.
+     * @param Component $base Postbox base component.
+     * @param Postbox $postbox Current postbox to build for.
+     * @param mixed $data Data returned by data function.
```
JoosuaKoskinen, line 217: You are not using $data param

src/modules/upkeep.php
```diff
@@ -187,7 +208,196 @@ public static function register_scripts( $page ) {
...
+
+      if ( $polling === false ) {
+        if ( array_key_exists($php_version, $php_version_array) ) {
+
+          if ( $php_version === $current_version ) {
+            $response = new AjaxResponse();
+            $response->is_success(true);
+            $response->set_data(
+              array(
+                'output' => Template::error_paragraph(__('The selected PHP version is already in use.', 'seravo'))->to_html(),
```
JoosuaKoskinen, line 331: Another `Template->to_html`

src/modules/upkeep.php
JoosuaKoskinen, line 80: `wp-seravo-plugin-update` is just ran on background? What happens after the update is done? Is anything outputted to user? I'm not completely sure how this is supposed to work and this is bit hard to test.

JoosuaKoskinen (changes requested):
Also, none of the old postboxes worked anymore because JavaScript seems to be broken.

JoosuaKoskinen:
Was the Javascript fixed yet? Still couldn't get it to work.

Moppa5:
@JoosuaKoskinen Should be fixed by now. At least I could get the old postboxes working. Are you sure you have pulled the new work?

JoosuaKoskinen (approved):

PR #574, opened 2021-07-05
Add FancyForm AJAX handler and postbox

JoosuaKoskinen:

Adds `FancyForm` AJAX handler and postbox. FancyForm is like SimpleForm but
with wrapper output, titles and status colors. Also adds test-page postbox for it
and renames `set_button_func` to `set_ajax_func`.

The JS is currently bit messy but  I will clean up that later.

The test-page postbox is a good start.


src/lib/postbox/handler.php
```diff
@@ -143,6 +143,8 @@ public function enqueue_postboxes_scripts() {
           'server_invalid_response' => __('Error: Something unexpected happened! Server responded with invalid data.', 'seravo'),
           'server_timeout' => __("Error: Request timeout! Server didn't respond in time.", 'seravo'),
           'server_error' => __("Error: Oups, this wasn't supposed to happen! Please see the php-error.log.", 'seravo'),
+          'show_more' => __('Show more', 'seravo'),
+          'show_less' => __('Show less', 'seravo'),
```
Moppa5, line 147: Just a comment that the original had "Toggle details" for both cases (opening / closing). Is this something that we want to move from?

src/lib/ajax/templates/fancy-form.php
```diff
@@ -0,0 +1,144 @@
...
+    if ( $this->build_form_func !== null ) {
+      \call_user_func($this->build_form_func, $form);
+    }
+
+    if ( $this->dryrun_button_text !== null ) {
+      $buttons = Template::side_by_side(
+        array(
+          Template::button($this->dryrun_button_text, $section . '-dryrun-button', 'button-primary'),
+          Template::button($this->button_text, $section . '-button', 'button-primary', true),
+        )
```
Moppa5, line 84: `side_by_side` takes Components left and right, not array?

src/lib/ajax/templates/fancy-form.php
```diff
@@ -0,0 +1,144 @@
...
+   */
+  const STATUS_YELLOW = 'rgb(232, 186, 27)';
+  /**
+   * @var string Green status color. Used for success.
+   */
+  const STATUS_GREEN = 'rgb(3, 129, 3)';
+  /**
+   * @var string Red status color. Used for failure.
+   */
+  const STATUS_RED = 'red';
```
Moppa5, line 28: Are these color constants something we want to the PHP files or should we rather use CSS and apply them with JS / PHP?

Moppa5 (changes requested):
Looking good! Tested the simple `wp-test` widget. See the` side_by_side `and consider the other two comments, they are not necessarily something that must be applied 😃.

src/lib/postbox/handler.php
JoosuaKoskinen: Toggle details is not the only one used originally, both are used. In my opinion, `show more` / `show less` is better.

src/lib/ajax/templates/fancy-form.php
JoosuaKoskinen: Fixed

Moppa5 (approved):
As far as I'm concerned feel free to merge. Those other comments are most likely resolved. The color thing is something to consider but the choice is yours.

PR #577, opened 2021-07-07
Rewrite site checks postbox

Moppa5:
Rewrite Site checks postbox with the new FancyForm. 
- Misc modifications and additions to CSS
- Add method for getting site checks result in HTML
- Remove old excessive code


style/common.css
```diff
@@ -62,16 +88,16 @@ div#seravo-low-disk-space-widget.postbox {
...
-p.failure {
-  color: red;
+p.failure, h3.failure {
+  color: red !important;
 }
 
-p.warning {
-  color: orange;
+p.warning, h3.warning {
+  color: orange !important;
```
Moppa5, line 100: `!important` was added so that no parent div or other can override these specific classes.

style/common.css
```diff
@@ -178,7 +204,7 @@ table.widefat {
   border-left: solid 0.5em #e8ba1b;
   box-shadow: 1px 1px 1px 1px rgba(0,0,0,0.12);
   display: block;
-  margin-top: 1em;
+  margin-top: 2em;
```
Moppa5, line 207: I don't think this was intended. Will check with possible other changes.

style/sitestatus.css
```diff
@@ -126,21 +126,13 @@ th, td {
...
-.seravo-site-check-issue-box {
-  padding: 10px;
-  border: 1.5px solid#ccd0d4;
-  border-left: solid 0.3em red;
-  border-bottom: none;
+/* Override the overflow-y on common css */
+#site-checks-output.seravo-result-wrapper-output {
+  overflow-y: initial;
+  padding: 2em;
 }
```
Moppa5, line 133: These styles were added to the `sitestatus.css` because with `overflow-y: scroll;` the tooltips on site checks postbox didn't work properly and were cut. If there's a better way, I'm certainly open to implement.

src/modules/sitestatus.php
```diff
@@ -815,33 +819,27 @@ public static function speed_test() {
...
-            </a>
-          </div>
-        </div>
-      </div>
-      <?php
+    /**
+     * AJAX function for Site checks postbox.
+     * @return \Seravo\Ajax\AjaxResponse
+     */
+    public static function run_site_checks() {
```
JoosuaKoskinen, line 826: The work itself is done in separate file, awesome!

style/sitestatus.css
JoosuaKoskinen: I sadly can't come up with a better solution. Let's go with this for now.

JoosuaKoskinen (changes requested):
LGTM! Just check the `em` you found yourself.

style/common.css
Moppa5: In this specific case I think these should never be overridden because there's no reason to use these styles if one doesn't need their properties. I wasn't able to find the root cause for the CSS that overrides these but remember something about `#dashboard-widgets` so might be related to seravo-postbox CSS or WordPress CSS in general as well. However, I think these can stay with `!important` as personally I have intended these styles not to be overridden.

PR #578, opened 2021-07-07
Add ability to use charts in new postboxes

JoosuaKoskinen:

Adds the ability to use charts in new postboxes. Also adds
JS event so custom JS is easier to use. 

There's a test postbox on test-page.


src/lib/postbox/toolpage.php
```diff
@@ -44,6 +44,24 @@ function() {
...
+  public function enable_charts() {
+    add_action(
+      'admin_enqueue_scripts',
+      function( $page ) {
+      if ( $page !== $this->screen ) {
+          return;
+      }
+
+        wp_enqueue_script('apexcharts-js', SERAVO_PLUGIN_URL . 'js/lib/apexcharts.js', '', \Seravo\Helpers::seravo_plugin_version(), true);
+        wp_enqueue_script('seravo-charts', SERAVO_PLUGIN_URL . 'js/charts.js', array( 'jquery' ), \Seravo\Helpers::seravo_plugin_version(), false);
```
Moppa5, line 60: There will be only one chart file to attach on the page? Do we put the old charts code there as well?

Moppa5 (commented):
Great work! Might you have any other examples regarding the custom JS with the new event? Currently there's just the chart but I understand that one can add any custom JS easily? Also, note the one comment about `chart.js` file and it's usage.

src/lib/postbox/toolpage.php
Moppa5: As we discussed the charts can be placed into one file.

Moppa5 (approved):

PR #580, opened 2021-07-12
Reorganize AJAX JavaScript

JoosuaKoskinen:

The JavaScript was a bit nasty so it's now
little better.

Also added cache bursting on `SERAVO_PLUGIN_DEBUG` mode and JS events
`seravoAjaxError` and `seravoAjaxRequest`.

Old code shouldn't be affected.


js/lib/ajax/ajax-handler.js
```diff
@@ -0,0 +1,230 @@
...
+  var dryrun_button = jQuery(section).find('#' + section_name + '-dryrun-button');
+  var spinner = jQuery('#' + section_name + '-spinner').closest('.seravo-spinner-wrapper');
+  var output = jQuery('#' + section_name + '-output');
+
+  function on_success(response) {
+    // Show output
+    spinner.hide();
+    output.html('<hr>' + response['output'] + '<hr>');
+    output.show();
+
```
Moppa5, line 81: Would you consider adding on this PR
Screenshot(14)

Basically adding again the optional `<hr>` tag usage as for example object cache nag looks ugly with these and of course the `.hideable` elements.

js/common.js
```diff
@@ -0,0 +1,86 @@
+"use strict";
+
+jQuery(document).ready(
+  function () {
+
+    jQuery('.seravo-show-more-wrapper').click(on_show_more_wrapper_click);
+
```
Moppa5, line 7: Not necessarily an issue but these empty lines look weird here.

Moppa5 (commented):
Looking good! I think the JS structure is now even more clearer. I had only one minor comment and some additions from `rewrite-cache-status` branch, which I would like you to consider. 

Anyway, after fixing the phpcs errors etc. I think you can feel free to merge this.

Moppa5 (approved):

JoosuaKoskinen:
This has been merged in https://github.com/Seravo/seravo-plugin/commit/76f72aadb68db256b7c3b37d1370481bc2399405.

PR #581, opened 2021-07-14
Add nag bars for postboxes

JoosuaKoskinen:

Adds nag bars that support SimpleForms in them. The bar can contain any
component but it's meant for paragraphs and SimpleForm components with
only button on them.

If the nag bar has SimpleForm button in it, the bar is removed on 'success' response.

There's a test postbox on test-page.


Moppa5 (approved):
Looks great and this is very nice addition for the postboxes in general. Feel free to merge 🎉

PR #583, opened 2021-07-15
Rewrite speed test postbox

Moppa5:
Rewrite the speed test postbox and functionality with the new design / Simpleform.
- Add quick toggle for clearing the location / page input
- Add `speedtest.js` for executing the speed test related AJAX requests etc.
- Move chart creation to the common `chart.js`


js/speedtest.js
```diff
@@ -0,0 +1,135 @@
+// phpcs:disable PEAR.Functions.FunctionCallSignature
```
JoosuaKoskinen, line 1: Is this actually needed?

js/speedtest.js
```diff
@@ -0,0 +1,135 @@
...
+// Number of curl commands the speed test uses for both cached and non-cached
+const speedNumberOfTests = 10;
+// Speed test data for the chart
+let speedData = [];
+// Speed test data for the chart
+let speedDataCached = [];
+// Visual chart of the speed test
+let speedChart;
+
+jQuery(document).ready(function ($) {
```
JoosuaKoskinen, line 13: Don't place everything in `jQuery(document).ready()`. Only put this part in it:
```
// Check if there is a 'speed_test_target' key in the url
let searchParams = new URLSearchParams(window.location.search);
if (searchParams.has('speed_test_target')) {
  run_speed_test();
}

$('#speed-test-button').click(run_speed_test);
$('#clear-url').click(
  function() {
    $("#speed-test-url").val('');
});
  ```
  
  Functions don't need to be defined there. Just useless indents.

JoosuaKoskinen (changes requested):

js/speedtest.js
Moppa5: It is as without there are some phpcs nags that phpcbf can't fix. I can re-check this again when working with the other request.

js/speedtest.js
Moppa5: Sure. This is most likely a habit that comes from the page specific js files where nearly everything is placed inside the `jQuery(document).ready...` However, I understand the point and will "fix".

js/speedtest.js
Moppa5: After moving the functions out of the `jQuery` part it seems that the issues are fixed. Will remove.

JoosuaKoskinen (approved):
LGTM & FFTM

PR #584, opened 2021-07-16
Rewrite screenshots postbox

Moppa5:
Rewrite the screenshots postbox with the new Postbox design. 
- Add screenshots.js for the slider logic
- Fix Seravo Updates postbox email nag text missing (_There must be at least one contact email_)


src/lib/postbox/toolpage.php
```diff
@@ -50,6 +50,7 @@ function( $page ) {
           'server_error' => __("Error: Oups, this wasn't supposed to happen! Please see the php-error.log.", 'seravo'),
           'show_more' => __('Show more', 'seravo'),
           'show_less' => __('Show less', 'seravo'),
+          'email_fail' => __('There must be at least one contact email', 'seravo'),
```
JoosuaKoskinen, line 53: This translation is only used by upkeep.js so it should be added to `seravo_upkeep` script, not `seravo-ajax`

JoosuaKoskinen (changes requested):
LGTM and FFTM after resolving one requested change.

src/lib/postbox/toolpage.php
Moppa5: Let's go with that then.

PR #585, opened 2021-07-19
Rewrite database size postbox

Moppa5:
Rewrite the database size postbox with the new design. 

- Remove old AJAX & JS files
- Remove the old "Toggle details" wrapper implementation
- Remove reports-chart.js completely
- Remove lots of excessive css
- New seravo-container element


src/modules/database.php
```diff
@@ -19,26 +20,29 @@
...
+      add_action(
+        'admin_enqueue_scripts',
+        function( $page ) {
+          wp_register_style('seravo_database', SERAVO_PLUGIN_URL . 'style/database.css', '', Helpers::seravo_plugin_version());
+
+          if ( $page === 'tools_page_database_page' ) {
+            wp_enqueue_style('seravo_database');
+          }
+        }
+      );
```
Moppa5, line 45: Removed the `public static` method as it felt quite excessive.

JoosuaKoskinen: My PR does the opposite :D

Moppa5: We can go with that `enqueue_scripts` scripts thing.

src/modules/database.php
```diff
@@ -281,65 +282,157 @@ public static function execute_search_replace() {
...
-        </div>
-        <?php
-      endif; // end database info
+    /**
+     * @param Component $base The base component to add child elements.
+     * @param Postbox\Postbox $postbox The postbox to add the components.
+     */
+    public static function build_database_size( Component $base, Postbox\Postbox $postbox ) {
+      $base->add_child($postbox->get_ajax_handler('db-info')->get_component());
+      $base->add_child(Component::from_raw('<hr><b>' . __('Table sizes', 'seravo') . '</b>'));
```
JoosuaKoskinen, line 291: Is there a reason for using `no_hr` and using one here instead?

JoosuaKoskinen: Oh, I get it, because it would be between the chart and title otherwise. All good.

style/database.css
```diff
@@ -80,73 +37,17 @@ label.from_label {
...
   padding: 3px;
 }
 
-.database_detail_table {
+.database-detail-table {
   padding: 10px 0px;
 }
 
-.database_detail_table, .database_detail_table td {
+.database-detail-table, .database-detail-table td {
```
JoosuaKoskinen, line 49: Could some class in common.css be used instead? These definitions look quite generic and we are trying to get rid of as many \*page\*.css files as possible.

js/charts.js
```diff
@@ -175,8 +176,10 @@ function generate_database_bars( json_data ) {
     },
     tooltip: {
       enabled: true,
+      shared: true,
+      intersect: false,
       custom: function({series, seriesIndex, dataPointIndex, w}) {
-        return '<div class="arrow_box">' +
+        return '<div class="arrow-box">' +
```
JoosuaKoskinen, line 182: I'm not sure I get how this works. Isn't `arrow-box` defined in `sitestatus.css`?

JoosuaKoskinen (changes requested):
The "Table sizes" chart is quite long (can't even fit the whole chart in one screenshot). Maybe it should have scrollbar or maybe just print the 10 largest tables?

js/charts.js
Moppa5: `arrow-box` is defined both in `sitestatus.css` and `database.css`

style/database.css
Moppa5: To me it seems that `result-table` on `common.css` is good enough for this. Will modify.

js/charts.js
Moppa5: Moved to `common.css` under the tooltips as it's chart custom tooltip.

JoosuaKoskinen (approved):

PR #586, opened 2021-07-19
Rewrite creating pages

JoosuaKoskinen:
- Move pages under `src/module/pages/` to seperate them from real modules.
- Composer autoload pages.
- Make pages be instances of `\Seravo\Postbox\Toolpage` instead of creating an Toolpage instance separately.
- Add requirements/access management for Toolpages.
- Make Toolpages take care of submenu registering instead of need for `toolbox.php`. 
- Fix invalid script registering and unify them.

This is going to create conflicts, needs to be merged carefully.


Moppa5 (commented):
Looking great overall 🎉! It seems that Database size widget is not working as the disk bars charts are not loading and spinners are spinning forever. However, #585 should take care of rewriting the widget and as we discussed it would be nice to merge these new widgets before this massive PR.

I will probably re-review this when everything is fully ready. However, I didn't come across any major flaws that we should take care of. The new `pages/` structure is much better, and to me it seems that the pages are getting simpler (there's no excessive loading and stuff done). Great job!

JoosuaKoskinen:
> It seems that Database size widget is not working as the disk bars charts are not loading and spinners are spinning forever

That was because I forgot to enable charts for the page (`$this->enable_charts();`). Should be fixed now. 

> I will probably re-review this when everything is fully ready

Yep, the PR is probably going to change a bit as we merge old PR's first.

src/lib/postbox/toolpage.php
```diff
@@ -8,24 +8,76 @@
  * Toolpage is a simple way to handle a postbox screen as
  * it takes care of the common features like nonces.
  */
```
Moppa5, line 10: Should we update the class comment in to something like "Toolpage is a base class for building pages that provide a simple way to handle a postbox screen --". For some reason the comment feels a little outdated regarding this new abstract design.

seravo-plugin.php
```diff
@@ -294,78 +293,50 @@ public static function load_all_modules() {
...
-     * Hide some functionality in multisites from normal admins
+    // OLD AJAX FILES
+    require_once SERAVO_PLUGIN_SRC . 'lib/database-ajax.php';
+    require_once SERAVO_PLUGIN_SRC . 'lib/domains-ajax.php';
+    require_once SERAVO_PLUGIN_SRC . 'lib/domain-tables.php';
+    require_once SERAVO_PLUGIN_SRC . 'lib/cruftfiles-ajax.php';
+    require_once SERAVO_PLUGIN_SRC . 'lib/cruftplugins-ajax.php';
+    require_once SERAVO_PLUGIN_SRC . 'lib/cruftthemes-ajax.php';
+    require_once SERAVO_PLUGIN_SRC . 'lib/sitestatus-ajax.php';
+    require_once SERAVO_PLUGIN_SRC . 'modules/check-site-health.php';
```
Moppa5, line 304: Site Health module is not an old AJAX file nor has it anything to do with AJAX?

Moppa5 (changes requested):
Great work! Consider those two changes / questions and after that I think you can feel free to merge this 🎉

src/lib/postbox/toolpage.php
JoosuaKoskinen: Comment changed.

seravo-plugin.php
JoosuaKoskinen: Added a comment for it.

PR #588, opened 2021-07-21
Rewrite cruftfiles postbox

Moppa5:
Rewrite the cruftfiles postbox with the new design.

- Autoloadable CruftRemover PHP class
- Common JS file for all cruft remover widgets
- Some css style changes & additions
- Move cruft files listing under `seravo-container` so that it overflows rather than consuming the whole page

#### Manual testing steps?
One can use for example the following bash script for generating some mock cruft files when testing.
```
#!/bin/bash
for ((index = 0; index <32; index++))
do
touch /data/wordpress/$index.sql
done
```


js/cruftremover.js
```diff
@@ -0,0 +1,172 @@
...
+          }
+        );
+      }
+    }
+  );
+}
+
+/**
+ * Remove the given cruft entries on a section with animation.
+ * @param {array} remove_list List of the content to remove.
```
JoosuaKoskinen, line 62: \@Param comments, awesome!

src/modules/pages/security.php
```diff
@@ -391,24 +409,68 @@ public static function get_last_successful_logins() {
...
-      <div id="cruftfiles_status_loading">
-        <?php _e('Searching for files...', 'seravo'); ?>
-        <img src="/wp-admin/images/spinner.gif">
-      </div>
-    </div>
-    </p>
-    <?php
+  /**
+   * Build func for cruft files postbox.
+   * @param $base The base of the postbox to add elements.
```
JoosuaKoskinen, line 414: The `$base` is missing type.

style/cruftremover.css
```diff
@@ -15,6 +15,19 @@
   margin-bottom: 20px;
 }
 
+.cruft-delete-button {
```
JoosuaKoskinen, line 18: Why doesn't it use `.button`?

src/modules/pages/security.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Seravo;
 
+use Seravo\Ajax\AjaxHandler;
```
JoosuaKoskinen, line 5: Please make it `\Seravo\...`

JoosuaKoskinen (changes requested):
The code is good! Just a couple of design notes: 

- The button looks different than other buttons in the plugin
cruftfile1
- The postbox is really tall, could be much shorter as there's the scrollbar
- There's a weird line at bottom right corner:
cruftfile2
- Maybe use WordPress popup like this instead of `confirm`?
cruftfile3

style/cruftremover.css
Moppa5: Good point. Maybe too much reverse engineering and didn't even think about that. I will try to fit the consistent button as other widgets have as well :)

js/cruftremover.js
```diff
@@ -0,0 +1,193 @@
...
+    // Init checkbox functionality
+    init_cruft_selects(section);
+
+    section.find('.button-primary').click(
+      function () {
+        //event.preventDefault();
+        /*var is_user_sure = confirm(cruftremover_l10n.confirm);
+        if (! is_user_sure) {
+          return;
+        } */
```
Moppa5, line 137: Should remove these...

JoosuaKoskinen (changes requested):
- The "The selected files have..." is on bit weird place, almost touching the table. Is the text even needed? We could show the "Not removed" text on failure but on success customer will see the entry disappear from the table.
cruftfiles
- The weird line from previous review is still there

LGTM otherwise

PR #589, opened 2021-07-21
Add wrapper for WordPress settings API

JoosuaKoskinen:

Adds an easy to use wrapper for settings. Also added
a postbox for ease making of setting postboxes.

This is also adds some improvements to the settings
system like email-list field, pre-made sanitizers and
better notification system.

Using settings should take much less effort now.

There's once again a test postbox on test-page. There's also 3 PRs using
this feature: https://github.com/Seravo/seravo-plugin/pull/592, https://github.com/Seravo/seravo-plugin/pull/591, https://github.com/Seravo/seravo-plugin/pull/590.


js/common.js
```diff
@@ -47,7 +54,81 @@ function on_show_more_wrapper_click(event) {
...
+
+  if (emails === undefined) {
+    return;
+  }
+
+  emails.split(',').forEach(
+    function (email) {
+    if (email === "") {
+        return;
+    }
```
Moppa5, line 101: Just a small inconvenience as the if statement is not in line with the function.

seravo-plugin.php
```diff
@@ -299,7 +299,6 @@ public static function load_all_modules() {
     require_once SERAVO_PLUGIN_SRC . 'modules/check-site-health.php';
 
     // OLD AJAX FILES
-    require_once SERAVO_PLUGIN_SRC . 'lib/database-ajax.php';
```
Moppa5, line 302: We should really pay attention that every unnecessary module is completely removed everywhere, both when coding and reviewing 👍🏼.

src/lib/postbox/component/settings.php
```diff
@@ -0,0 +1,465 @@
...
+class Settings {
+
+  /**
+   * @var string Field type for textfield.
+   */
+  const FIELD_TYPE_STRING = 'string';
+  /**
+   * @var string Field type for checkbox.
+   */
+  const FIELD_TYPE_BOOLEAN = 'boolean';
```
Moppa5, line 20: Could this just be `FIELD_TYPE_CHECKBOX`? It was really confusing at first what does boolean mean here.

src/lib/postbox/component/settings.php
```diff
@@ -0,0 +1,465 @@
...
+          $type = 'array';
+          $build_func = array( __CLASS__, 'build_email_list_field' );
+          $sanitizer = $sanitizer === null ? function( $value ) use ( $name, $default ) {
+            $default = empty(get_option($name)) ? ($default === null ? array() : $default) : get_option($name);
+            return self::sanitize_email_list_field($value, $default);
+          } : $sanitizer;
+          break;
+        default:
+          return;
+      }
```
Moppa5, line 132: Consider providing this part as a private method for instance.

src/lib/postbox/component/settings.php
```diff
@@ -0,0 +1,465 @@
...
+            return self::sanitize_number_field($value, $default);
+          } : $sanitizer;
+          break;
+        case self::FIELD_TYPE_EMAIL_LIST:
+          $type = 'array';
+          $build_func = array( __CLASS__, 'build_email_list_field' );
+          $sanitizer = $sanitizer === null ? function( $value ) use ( $name, $default ) {
+            $default = empty(get_option($name)) ? ($default === null ? array() : $default) : get_option($name);
+            return self::sanitize_email_list_field($value, $default);
+          } : $sanitizer;
```
Moppa5, line 128: These methods are not static thus `self::` seems excessive.

src/lib/postbox/component/settings.php
```diff
@@ -0,0 +1,465 @@
...
+    $input = new Component('', '<tr>', '</tr>');
+    $input->add_child(Component::from_raw('<td><input type="email" placeholder="' . $placeholder . '"/></td>'));
+    $input->add_child(Template::button(__('Add', 'seravo'), $field['id'] . '-button', 'button email-list-add-button')->set_wrapper('<td>', '</td>'));
+
+    $table = new Component('', '<table class="email-list-input">', '</table>');
+    $table->add_child($hidden_field);
+    $table->add_child($input);
+
+    return $table;
+  }
```
Moppa5, line 362: This class is becoming a rather god class-like here. If there will ever be more to build, this will start to look like a mess. Maybe own helper for generating Settings components or one static method that could utilize the switch case as well. Basically could just become then `public static function build_settings_field( $field, $type)`. However, it tends to get the same mess issue with the one static method as well so I would really consider moving these into own helper class or something.

src/lib/postbox/component/settings.php
```diff
@@ -0,0 +1,465 @@
...
+      $this->add_notification('invalid-email', __("Invalid emails weren't saved", 'seravo'));
+    }
+
+    if ( empty($valid_emails) ) {
+      $this->add_notification('no-valid-email', __('At least one valid email required', 'seravo'));
+      return $default;
+    }
+
+    return $valid_emails;
+  }
```
Moppa5, line 447: Here to note again. I don't really think that the Settings class should have this much responsibility regarding sanitizing the fields as well. Settings class should provide only the most crucial parts regarding using the WordPress API effortlessly.

src/lib/postbox/postbox.php
```diff
@@ -278,6 +282,31 @@ public function get_ajax_handler( $section ) {
     return null;
   }
 
+  /**
+   * Adds a setting section for the postbox. The same instance
+   * shouldn't be added to multiple postboxes without cloning.
```
Moppa5, line 287: Could you please elaborate what does "without cloning" mean here?

Moppa5 (changes requested):
Great work overall 🎉. I had a couple of small nags and notes but one more larger as well. In general, it seems to me that Settings class is now taking too much responsibility of all kind of things like generating and sanitizing the components as well, which in my opinion is not a good design.

src/lib/postbox/component/settings.php
Moppa5: To be clear enough, I mean the `self::sanitize_*` functions.

js/common.js
JoosuaKoskinen: Fixed

seravo-plugin.php
JoosuaKoskinen: Yep

src/lib/postbox/component/settings.php
JoosuaKoskinen: WordPress API uses the name `boolean` for it. It means the data type is a boolean. The user of this wrapper doesn't need to know what type of input there is, they just need to tell what kind of value they want from user and checkbox is only logical in this case. I can change it to checkbox if you want to but then we probably need to change most of the other field type names too.

src/lib/postbox/component/settings.php
JoosuaKoskinen: Fixed

src/lib/postbox/component/settings.php
JoosuaKoskinen: I don't know, I count these as crucial parts. This is the most basic validation every field will **always** have. The forms may have more validation in addition (see image optimization PR image resolution limits for example) but these checks must always exists.

We always have to see that the is exactly as requested or anyone could input arbitrary data in wp_options table. And this is just the bare minimum. Checking checkbox really sent checkbox data, seeing that number is a number, email is email etc.

I don't see why these would be done separately every time, it increase the change for a mistake by a million times. And moving these to another class would be kind of pointless as the only thing using them would be Settings class internals anyway. 

 We tell Settings class "Ask user whether he wants to enable a feature, the value should be a boolean" and Settings class should do that for us. That's an easy to use wrapper.

src/lib/postbox/postbox.php
JoosuaKoskinen: https://www.php.net/manual/en/language.oop5.cloning.php

src/lib/postbox/component/settings.php
Moppa5: We can stick with `FIELD_TYPE_BOOLEAN` then for the sake of the consistency.

src/lib/postbox/component/settings.php
JoosuaKoskinen: Fixed

Moppa5 (approved):
As discussed about the design, this might be currently the best effort we can have so I'd stay we can stick with it 🎉 and improve it later. Feel free to merge this 👍🏼

PR #591, opened 2021-07-21
Rewrite optimize images postbox

JoosuaKoskinen:

Rewrites optimize images postbox to use the new settings wrapper.

This PR is based on https://github.com/Seravo/seravo-plugin/pull/589 and https://github.com/Seravo/seravo-plugin/pull/590, they must be merged first. The two extra commits are from them.


Moppa5 (changes requested):
Looks great overall! However, I stumbled upon some kind of a bug / inconvenience as when saving for example "Optimize images" settings the "Settings saved" message appears on both the current and "Sanitize Uploads" postbox. I don't think this is intended so could you debug? After that you can feel free to merge 🎉

PR #592, opened 2021-07-21
Rewrite security settings postbox

JoosuaKoskinen:

Rewrites security settings postbox to use the new settings wrapper.

This PR is based on https://github.com/Seravo/seravo-plugin/pull/589, that must be merged first. The extra commit is from that PR.

### DIFF
--- a/src/lib/postbox/component/template.php
+++ b/src/lib/postbox/component/template.php
@@ -186,6 +186,14 @@ class Template {
   }
 
   /**
+   * Display fake checkbox used to tell features that can't be disabled.
+   * @return \Seravo\Postbox\Component Disabled fake checkbox.
+   */
+  public static function fake_checkbox() {
+    return Component::from_raw('<input type="checkbox" checked="on" disabled="disabled"/>');
+  }
+
+  /**
    * Display datetime picker as HTML input element.
    * @param string $label   The label text.
    * @param string $id      Unique id of the input.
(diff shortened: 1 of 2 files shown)


src/lib/postbox/component/template.php
```diff
@@ -185,6 +185,14 @@ public static function checkbox_with_label( $label, $name, $checked = false ) {
     return $wrapper;
   }
 
+  /**
+   * Display fake checkbox used to tell features that can't be disabled.
+   * @return \Seravo\Postbox\Component Disabled fake checkbox.
+   */
+  public static function fake_checkbox() {
+    return Component::from_raw('<input type="checkbox" checked="on" disabled="disabled"/>');
+  }
```
Moppa5, line 194: `fake_checkbox()` name feels weird here. Could you come up with some better named function? If not, then let's stick with this.

`disabled_checkbox()` feels rather inconvenient as well or what you think?

Moppa5 (approved):
The settings seem to work and looks clean. I had only one comment about the `fake_checkbox()` to consider. By all means feel free to merge 🎉

src/lib/postbox/component/template.php
JoosuaKoskinen: I can't really come up with a better name. `disabled_checkbox` is not good as the checkbox is not only disabled, it doesn't have a name etc.

PR #594, opened 2021-07-23
Rewrite cruft plugins postbox

Moppa5:
Rewrite the cruft plugins postbox with the new design.
- Remove and move old AJAX code to CruftRemover class
- Remove old CSS from `cruftremover.css`
- Move some functions to `cruftremover.js`
- Add spinner and proper error messages if removing a plugin(s) fails

You can easily generate cruft plugins for example with this bash script
```
#!/bin/bash
wp plugin install w3-total-cache
wp plugin install wp-super-cache
wp plugin install better-wp-security
wp plugin install wordfence
wp plugin install jetpack
wp plugin install backup
wp plugin install wp-cerber
wp plugin install all-in-one-wp-migration
```

This PR is depending on #588 so it should be merged first.


js/cruftremover.js
```diff
@@ -170,3 +209,159 @@ function cruft_files(event, response) {
     status_section.html('<b>' + error + '</b>');
   }
 }
+
+/**
+ * Build and initialize cruft files functionality.
+ * @param {event} event Event that triggered the function call. Default seravoAjaxSuccess.
+ * @param {array} response Array containing possible cruft files found.
+ */
+function cruft_plugins(event, response) {
```
JoosuaKoskinen, line 218: Okay this is not a general cruft remover thing at all but a huge postbox specific function to take care of almost everything. Is there a reason this cant be done with one common function?

src/lib/cruftremover.php
```diff
@@ -245,12 +245,12 @@ public static function list_cruft_files() {
     foreach ( $list_known_dirs as $dirname ) {
       exec('ls -d ' . $dirname, $cruft_found);
         if ( ! empty($cruft_found) ) {
-          foreach ( $cruft_found as $key => $cruft_dir ) {
-            if ( self::only_has_whitelisted_content($cruft_dir, $white_list_files, $white_list_dirs) ) {
-              unset($cruft_found[$key]);
+        foreach ( $cruft_found as $key => $cruft_dir ) {
```
JoosuaKoskinen, line 248: Looks like you broke the indentation

js/cruftremover.js
```diff
@@ -79,13 +80,51 @@ function remove_cruft_entries(remove_list, table, section ) {
...
     }
   );
 }
 
+/**
+ * Sorts / categorizes plugins based on inactive, foolish plugin etc.
+ * @param {array} dataArray Array to sort and categorize
+ * @returns Sorted array with plugins categorized.
+ */
+function sort_data_array(dataArray) {
```
JoosuaKoskinen, line 97: Could the data just be sorted on backend?

src/modules/pages/security.php
```diff
@@ -315,6 +319,8 @@ public static function get_security_settings() {
   public static function build_last_logins( Component $base, Postbox\Postbox $postbox ) {
     $base->add_child(Template::paragraph(__('This tool can be used to retrieve last 10 successful logins. For more details and full login log see <a href="tools.php?page=logs_page&logfile=wp-login.log" target="_blank">wp-login.log</a>.', 'seravo')));
     $base->add_child($postbox->get_ajax_handler('logins-info')->get_component());
+    $base->add_child(Component::from_raw('<div id="cruft-plugins-status"></div>'));
```
JoosuaKoskinen, line 322: This makes no sense

src/modules/pages/security.php
```diff
@@ -473,25 +479,72 @@ public static function remove_cruft_files() {
...
+            }
+          }
+        }
+      }
+    }
+
+    $response->is_success(true);
+    $response->set_data(
+      array(
+        'data' => $call_section,
```
JoosuaKoskinen, line 542: What is this call_section and why is just passed back without using it?

JoosuaKoskinen (changes requested):
This still reminds me a lot of the old design, with cruft*.php files and cruft*.js just combined together. I would like to see the JS be completely common code, it shouldn't need to even know what type of cruft remover it is. The postbox could just have `data-cruft-type=plugin` etc and JS would send it to one common AJAX handler which would then decide what function should be called to remove them:
```
public static function remove_cruft() {
  $cruft_type = isset($_POST['cruft-type']) ? $_POST['cruft-type'] : '';
  $cruft = isset($_POST['cruft']) ? $_POST['cruft'] : '';
  if ( $cruft_type === 'plugin' ) {
    return CruftRemover::remove_cruft_plugins($cruft);
  } else if ( $cruft_type === 'file' ) {
    return CruftRemover::remove_cruft_files($cruft);
  } else if ( $cruft_type === 'theme' ) {
    return CruftRemover::remove_cruft_themes($cruft);
  } else {
    return // *ajax error response*
  }
}
``` 

- The spinner is not very aligned:
spinner
- The cruftfiles and cruftplugins output are very different. Could you remove the text still shown after removing plugins? And the `<hr>` at the end.
cruftplugins2

js/cruftremover.js
Moppa5: I will investigate and work with the more common way.

src/modules/pages/security.php
Moppa5: True, added to a wrong function :D. However, when modifying in to more general style there should be something like `div class="cruft-status">`

src/modules/pages/security.php
Moppa5: The call section is removed for now, need to work out something else for it as well.

Moppa5:
Waiting now for us to re-implement the `cruftremover.js` with a better design.

Moppa5:
Closing this as #613 takes care of the cruft rewrite.

PR #596, opened 2021-07-28
Rename and autoload classes

Moppa5:
Rename classes with CamelCase naming and autoload the classes. Excluded Domains related classes as agreed. Please test each modification carefully. I've personally tested Seravo Plugin functionality and didn't find any issues. There might be some semantic issues how `seravo-plugin.php` should handle some permissions etc. but please elaborate and I will fix those as well. 

**Rename and autoload the following**

- Site health
- Security restrictions
- Login notifications
- Dashboard widgets
- Purge cache
- Seravotest auth bypass
- Speed test
- Third party fixes
- Site status page

**Autoload the following classes**

- Canonical domain and https
- Check default email
- Check https
- Fixes
- HideUsers
- InstanceSwitcher
- NoIndex
- OptimizeOnUpload
- Passwords
- Sanitize on uploads
- Svg support
- Wp login log
- Wp plugin log
- Wp user log
- Check PHP version

#543


seravo-plugin.php
```diff
@@ -212,101 +213,100 @@ public static function load_all_modules() {
...
+      PurgeCache::load();
+    }
 
     /*
      * Add a speed test button to the WP adminbar
      */
-    require_once SERAVO_PLUGIN_SRC . 'modules/speed-test.php';
+    if ( getenv('WP_ENV') === 'production' ) {
+      SpeedTest::load();
+    }
```
Moppa5, line 235: For example these are the parts I'm concerned. Is this the right way of handling or should we use something else?

src/modules/check-default-email.php
```diff
@@ -1,62 +1,54 @@
...
-      <?php
-      $siteurl = get_option('siteurl');
-      // translators: $s user's website url
-      $link = sprintf(wp_kses(__('Warning: A generic admin email was detected in the <a href="%s/wp-admin/options-general.php">site settings</a>. Please update it.', 'seravo'), array( 'a' => array( 'href' => array() ) )), esc_url($siteurl));
-      echo $link;
-      ?>
+  /**
+   * @return void
+   */
+  public static function _seravo_show_email_warning() { ?><div class="notice notice-error">
```
Moppa5, line 42: This is weird, might need to re-check that the code is aligned properly.

Moppa5: Checked these and fixed many of them. More will be fixed if found on review.

PR #598, opened 2021-07-30
Rewrite shadows postbox

Moppa5:
Rewrite the Shadows postbox with the new design.

- Rename `sitestatus.js` to `shadows.js`
- Move excessive code and css styles


js/shadows.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+
+    /**
+     * Hide the success / error container
+     */
+    jQuery('.closebtn').click(
+      function() {
+        jQuery(this).parent().hide();
+      }
+    );
```
Moppa5, line 56: Probably could be reduced as well.

Moppa5: Seems that jQuery doesn't let you do that Uncaught `TypeError: (intermediate value).apply is not a function`. Will stick with the current implementation.

src/modules/pages/sitestatus.php
```diff
@@ -235,6 +215,20 @@ public static function init_postboxes( Toolpage $page ) {
...
+    /**
+     * Shadows postox
+     */
+    $shadows = new Postbox\Postbox('shadows');
+    $shadows->set_title(__('Shadows', 'seravo'));
+    $shadows->set_build_func(array( __CLASS__, 'build_shadows' ));
+    $shadows->set_requirements(array( Requirements::CAN_BE_PRODUCTION => true ));
+    $shadows->set_data_func(array( __CLASS__, 'get_shadows_data' ));
+    // Init AJAX handler for resetting shadows
+    $remove_shadow = new Ajax\AjaxHandler('reset-shadows');
```
JoosuaKoskinen, line 228: Maybe shadow_reset would be a better name, shadow reset doesn't remove shadow

src/modules/pages/sitestatus.php
```diff
@@ -819,132 +813,127 @@ public static function get_site_info() {
...
+
+    if ( $polling === true ) {
+      $response = new AjaxResponse();
+      $response->is_success(true);
+      return $response;
+    }
+
+    if ( $polling === false ) {
+      // run the shadow reset here
+      if ( isset($_POST['shadow']) && $_POST['shadow'] !== '' ) {
```
JoosuaKoskinen, line 916: Return type from this function is `AjaxResponse|mixed`, that should tell that something is wrong as ajax funcion must always return AjaxResponse only.

The mixed is because you are trying to return `false`. If $polling is false but $_POST['shadow'] doesn't exists, nothing is done and `false` is returned because of the last line `return  $polling;`

JoosuaKoskinen (changes requested):
Couldn't do a full review as the code doesn't run:
```
Parse error: syntax error, unexpected token "<<" in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/src/modules/pages/sitestatus.php on line 831
```

src/modules/pages/sitestatus.php
Moppa5: Yeah sure. This was left since I first had all of these as "remove_shadows" for some arbitrary reason.

JoosuaKoskinen (changes requested):
- Not fan of the design and I don't think it's that useful to rewrite a postbox and use exactly the same design. Takes much less effort to make something with good design than to first rewrite a postbox with old design and then later rewrite with new design. 
- My site has 2 shadows. The issues I had trying this:
   - Spinner for every shadow starts spinning on reset button click
   - The buttons are not disabled and can be pressed multiple times
   - Clicking reset caused a loop and my browser started making 100 AJAX requests a second.
   
IMO, this design is not good and the implementation is quite buggy. I think this should be rewritten again, this feels a bit worse than the old one. I leave it up to you to decide what you want to do with this. Didn't review the code, will do that when re-review is requested.

Moppa5:
Let's see first of all what those bugs are, and then continue working with the new widget design. I agree that the current shadows design is not the best one.

js/shadows.js
```diff
@@ -0,0 +1,87 @@
...
+        var shadow_id = jQuery(this).closest('tbody').attr('id');
+        var shadow_reset_status = jQuery('#' + shadow_id).find('#shadow-reset-status');
+        var reset_button = jQuery(this);
+        var alert_success = jQuery('.alert#alert-success');
+        var alert_error = jQuery('.alert#alert-error');
+        var primary_domain = jQuery('#shadow-primary-domain');
+        var closebtn = jQuery('.closebtn');
+
+        tb_show(shadow_loc.confirm, '#TB_inline?width=600&height=120&inlineId=remove-shadow-modal');
+        jQuery('#remove-shadow-modal-proceed').click(on_confirm);
```
JoosuaKoskinen, line 17: Okay, the issue with AJAX is that you add the click event for the modal again every time .reset is click

JoosuaKoskinen (changes requested):
The design looking much better! The one change request is for the AJAX bug.

Also I think all the reset buttons should be disabled when one reset is in progress so nobody resets multiple shadows at once. That could cause bit load, not good to have multiple pollers making a request every couple of seconds and user wouldn't see the success/error/instruction banner because it would disappear as the other reset finishes.

js/shadows.js
Moppa5: Damn it jQuery 😄

Moppa5: @JoosuaKoskinen, All shadow buttons are now disabled if one reset is running. I moved the `seravo_ajax_request` related parts from the `.click` handler thus the bug should be gone.

js/shadows.js
```diff
@@ -0,0 +1,96 @@
...
+    jQuery('.reset').click(
+      function() {
+        // Get the details of this shadow
+        shadow_id = jQuery(this).closest('tbody').attr('id');
+        shadow_reset_status = jQuery('#' + shadow_id).find('#shadow-reset-status');
+        reset_button = jQuery(this);
+        alert_success = jQuery('.alert#alert-success');
+        alert_error = jQuery('.alert#alert-error');
+        primary_domain = jQuery('#shadow-primary-domain');
+        closebtn = jQuery('.closebtn');
```
Moppa5, line 56: These could be moved. Also, maybe the success / nag banner could hide if user starts shadow reset again?

JoosuaKoskinen (changes requested):
There's some issue with the success banner. The GIF below explained:
1. Ran reset for shadow with no domain -> got correct success banner
2. Ran reset for shadow with domain -> got correct success banner
3. Ran reset for the first shadow again -> got the banner meant for the 2. shadow

shadow

Also, the jQuery is currently begging for bugs:
- The variables on top are polluting the JS environment, they don't stay in file scope. And the names are really common (closebtn / reset_button / alert_error etc). If some script on some site uses variables with the same name, there's collision. Done correct, there shouldn't be need for variables on top of file.
- You are not finding the elements relative to other elements. You add click listener for every element with eg. `.reset` or `.closebtn` classes. Very likely to cause issues with some sites, maybe all sites.

So instead of doing everything like this:
```
var shadow_id;
var shadow_reset_status;
var reset_button;
var alert_success;
var alert_error;
var primary_domain;
var closebtn;

jQuery(document).ready(
  function() {
    ...
    shadow_id = jQuery(this).closest('tbody').attr('id');
    shadow_reset_status = jQuery('#' + shadow_id).find('#shadow-reset-status');
    reset_button = jQuery(this);
    alert_success = jQuery('.alert#alert-success');
    alert_error = jQuery('.alert#alert-error');
    primary_domain = jQuery('#shadow-primary-domain');
    closebtn = jQuery('.closebtn');
    ...
  }
);
```
Should be more like this:
```
jQuery(document).ready(
  function() {
     jQuery('.shadow-section').each(init_shadow_section); 
  }
);

function init_shadow_section(section) {
  ...
  // If all of these event need to be in a variable,
  // bit like this:
  var shadow_id = jQuery(section).find('tbody').attr('id');
  var shadow_reset_status = jQuery('#' + shadow_id).find('#shadow-reset-status');
  var reset_button = jQuery(section).find('.reset-button');
  var alert_success = jQuery(section).find('.alert#alert-success');
  var alert_error = jQuery(section).find('.alert#alert-error');
  var primary_domain = jQuery(section).find('#shadow-primary-domain');
  var closebtn = jQuery(section).find('.closebtn');
  ...
}
```

I don't mind this for now as I'm going to run ESLint on all our JS and do some refactoring, you may merge after fixing the banner issue. Up to you once again.

js/shadows.js
```diff
@@ -0,0 +1,102 @@
+'use strict';
+
+jQuery(document).ready(
+  function() {
+    jQuery('.shadow-section').each(
+      function() {
+        init_shadow_section(this);
```
JoosuaKoskinen, line 7: Could this just be `each(init_shadow_section)`

js/shadows.js
```diff
@@ -0,0 +1,102 @@
...
+    jQuery('.shadow-closebtn').click(
+      function() {
+        jQuery(this).parent().hide();
+      }
+    );
+
+    /**
+     * Open / fold the row containing additional information
+     */
+    jQuery('#shadow-table td.open-folded').on(
```
JoosuaKoskinen, line 20: I think you should add these .shadow-closebtn and open-folded click listeners in `init_shadow_section`, that's kind of what it's for.

js/shadows.js
```diff
@@ -0,0 +1,102 @@
...
+ * Init shadow tbody sections.
+ * @param {Object} section Shadow tbody section.
+ */
+function init_shadow_section(section) {
+  var section_element = jQuery(section);
+  var shadow_id = section_element.attr('id');
+  var shadow_reset_status = section_element.find('.reset-status');
+  var reset_button = section_element.find('.reset');
+  var alert_success = jQuery('.alert#alert-success');
+  var alert_error = jQuery('.alert#alert-error');
```
JoosuaKoskinen, line 55: You still don't find the elements relative to `section`, this jQuery applies to **all** elements with `.alert` -class and I guarantee that Seravo Plugin shadow reset postbox is not the only one using the class `.alert`.

js/shadows.js
```diff
@@ -0,0 +1,102 @@
...
+  var primary_domain = jQuery('#shadow-primary-domain');
+  var closebtn = jQuery('.shadow-closebtn');
+
+  function on_success(response) {
+    shadow_reset_status.empty();
+    jQuery('.reset').prop('disabled', false);
```
JoosuaKoskinen, line 61: Again disabling every `.reset` element there is, Seravo Plugin shadow reset postbox is not the only one using this class

js/shadows.js
```diff
@@ -0,0 +1,102 @@
...
+    var data = {
+      shadow: shadow_id
+    }
+    seravo_ajax_request('post', 'shadows', 'reset-shadows', on_success, on_error, data);
+    jQuery('.reset').prop('disabled', true);
+    shadow_reset_status.html("<img src=\"/wp-admin/images/spinner.gif\">");
+  }
+
+  function confirm() {
+    jQuery('#remove-shadow-modal-proceed').click(execute);
```
JoosuaKoskinen, line 95: You add the click listeners every time confirm is called.

Moppa5:
Big thanks to @JoosuaKoskinen for working with much better JS for the shadows. I added the cursor pointer to the alert close button, and removed excessive ids etc. Success & error banners should work as expected.

src/modules/pages/sitestatus.php
```diff
@@ -108,23 +107,15 @@ public static function enqueue_scripts( $screen ) {
       return;
     }
 
-    \wp_enqueue_script('seravo-site-status-js', SERAVO_PLUGIN_URL . 'js/sitestatus.js', array(), Helpers::seravo_plugin_version());
+    \wp_enqueue_script('seravo-shadows-js', SERAVO_PLUGIN_URL . 'js/shadows.js', array(), Helpers::seravo_plugin_version());
```
Moppa5, line 110: Probably should add `jquery` similar to the speedtest-js.

src/modules/pages/sitestatus.php
```diff
@@ -816,132 +811,134 @@ public static function get_site_info() {
...
+      $alert_success->add_child(Template::button('&times', 'shadow-closebtn', 'shadow-closebtn'));
+      $alert_success->add_child(Template::paragraph('<b>' . \__('Shadow reset successfully!', 'seravo') . '</b>'));
+      $shadow_reset_alert = new Component('', '<div class="shadow-reset-sr-alert alert">', '</div>');
+      $shadow_reset_alert->add_child(Template::paragraph(\__('Because this shadow uses a custom domain, <strong>please go to the shadow and run search-replace there with the values below</strong> for the shadow to be accessible after reset: ', 'seravo')));
+      $shadow_reset_alert->add_child(Template::paragraph(\__('<strong>From:</strong> ', 'seravo') . \str_replace(array( 'https://', 'http://' ), '://', \get_home_url()) . \__('<br><strong>To:</strong> ', 'seravo') . '://<span id="shadow-primary-domain"></span>'));
+      $shadow_reset_alert->add_child(Template::paragraph(\__('When you\'re in the shadow, run search-replace either on "Tools --> Database --> Search-Replace Tool" or with wp-cli. Instructions can be found from <a href="https://help.seravo.com/en/docs/151" target="_BLANK">documentation</a>.', 'seravo')));
```
JoosuaKoskinen, line 837: If the shadow has domain, it will redirect to the production site after reset until search-replace has been done so the "Tools --> Database --> Search-Replace Tool" will not work. WP-CLI is the only way to do this.

JoosuaKoskinen (changes requested):
Looking good! :tada: Feel free to merge after making the small text change

PR #601, opened 2021-08-03
Enhance Seravo Plugin Updater postbox

Moppa5:
Enhance the already re-written Seravo Plugin Updater postbox.

- Add a proper error handling for failed command
- Error log installation issues and recommend making an issue on Github
- Reload the page if installation run

The reload after installation is now done by using `location.reload()` as AJAX function not returning response / trying to redirect somewhere else will cause JS side `on_error`. Not sure if it's the best implementation, but please let me know how we could fix it for better 👍🏼.

### DIFF
--- a/src/modules/pages/upkeep.php
+++ b/src/modules/pages/upkeep.php
@@ -5,6 +5,7 @@ namespace Seravo\Page;
 use \Seravo\Helpers;
 use \Seravo\Shell;
 use \Seravo\API;
+use \Seravo\Compatibility;
 
 use \Seravo\Ajax;
 use \Seravo\Ajax\AjaxResponse;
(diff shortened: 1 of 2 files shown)


src/modules/pages/upkeep.php
```diff
@@ -415,13 +416,13 @@ public static function build_seravo_plugin_update_postbox( Component $base, Post
...
-    $base->add_child(Template::paragraph(__('Upstream version: ', 'seravo') . $data['upstream_version']));
+    $base->add_child(Template::paragraph(__('Current version: ', 'seravo') . '<b>' . $data['current_version'] . '</b>'));
+    $base->add_child(Template::paragraph(__('Upstream version: ', 'seravo') . '<b>' . $data['upstream_version'] . '</b>'));
 
     if ( $data['current_version'] == $data['upstream_version'] ) {
-      $base->add_child(Template::paragraph(__('The currently installed version is the same as the latest available version.', 'seravo'), 'success'));
+      $base->add_child(Template::paragraph(__('Seravo Plugin installation is up to date.', 'seravo'), 'success bold'));
     } else {
-      $base->add_child(Template::paragraph(__('There is a new version available', 'seravo'), 'warning'));
+      $base->add_child(Template::paragraph(__('There is a new version available', 'seravo'), 'warning bold'));
```
Moppa5, line 425: I think the bold texts work here much better, what you think? Also the text "version is up to date" was changed to something more understandable than "The version is the same as the upstream version" which rather adds complexity.

Screenshot(25)

JoosuaKoskinen (approved):
LGTM, feel free to merge

PR #607, opened 2021-08-06
Add strict PHPStan rules and run CI on oldest and latest WordPress stubs

JoosuaKoskinen:

- Add strict PHPStan rules. Changes mainly were:
    - Strict return check for `preg_match` and like. It may return `0` or `false` and loose comparison sees both as `false`.
    - Replace usage of `empty()` with strict comparison, `empty` is bad as means we don't know the type but we always must know the variable type.
    - Use strict modes of `in_array`, `base64_decode` etc.
    - Strict return check for `strpos`. It may return `0` (string was found) or `false` (not found) and loose comparison sees both as `false`. These were a big mistake, there's even a big banner on PHP docs warning about this. Thanks to PHPStan this shouldn't happen again.
- Remove email-list setting type as obsolete.
- Make CI run PHPStan on latest WordPress and WP-CLI stubs too so we know if some function was removed/changed in latest version.

### DIFF
--- a/src/modules/noindex-domain-alias.php
+++ b/src/modules/noindex-domain-alias.php
@@ -30,11 +30,11 @@ class Noindex {
     if ( isset($_SERVER['HTTP_HOST']) ) {
 
       if (
-        preg_match('/^.*\.wp-palvelu\.fi$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.seravo\.fi$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.seravo\.com$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.wp\..*$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.dev\..*$/', $_SERVER['HTTP_HOST'])
+        preg_match('/^.*\.wp-palvelu\.fi$/', $_SERVER['HTTP_HOST']) === 1 ||
+        preg_match('/^.*\.seravo\.fi$/', $_SERVER['HTTP_HOST']) === 1 ||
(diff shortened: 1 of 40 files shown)


src/modules/noindex-domain-alias.php
```diff
@@ -30,11 +30,11 @@ public static function maybe_hide_domain_alias( $output, $public ) {
...
-        preg_match('/^.*\.wp-palvelu\.fi$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.seravo\.fi$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.seravo\.com$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.wp\..*$/', $_SERVER['HTTP_HOST']) ||
-        preg_match('/^.*\.dev\..*$/', $_SERVER['HTTP_HOST'])
+        1 === preg_match('/^.*\.wp-palvelu\.fi$/', $_SERVER['HTTP_HOST']) ||
+        1 === preg_match('/^.*\.seravo\.fi$/', $_SERVER['HTTP_HOST']) ||
+        1 === preg_match('/^.*\.seravo\.com$/', $_SERVER['HTTP_HOST']) ||
+        1 === preg_match('/^.*\.wp\..*$/', $_SERVER['HTTP_HOST']) ||
+        1 === preg_match('/^.*\.dev\..*$/', $_SERVER['HTTP_HOST'])
```
Moppa5, line 37: Should these rather be `preg_match(...) === 1` for the clarity.

Moppa5 (changes requested):
Great! These will really help us in the future development. I had one rather minor "nag" about comparison order. However, there are the Rector warnings / errors that we agreed you'll fix for this PR as well. I think that after doing those you can merge this 👍🏼

PR #608, opened 2021-08-10
Add missing Rector rules

JoosuaKoskinen:

- Add script for comparing local and upstream Rector rules
- Add all Rector rules to rector.php (disabled are commented)
- Enable some new rules and disable some old rules

Should be quite easy to review, it's 99% preslash changes


Moppa5 (approved):
Looking good! I didn't find any issues while quickly going through the modifications and Seravo Plugin in general.

Issue #606, opened 2021-08-04
Replace exec usage with Compatibility::exec()


JoosuaKoskinen:
**Describe the bug**

Same case as with https://github.com/Seravo/seravo-plugin/issues/604.

The `exec` return type has changed on PHP 8.0 and `Compatibility::exec()` will take care of that. All usage of `exec` should be replaced with `Compatibility::exec()` and checked if the result is `false`.

Issue #604, opened 2021-08-04
Replace substr usage with Compatibility::substr()


JoosuaKoskinen:
**Describe the bug**

Up to PHP 8.0, `substr` used to return empty string on failure. Since PHP 8.0, it returns `false`. We have a function (coming in https://github.com/Seravo/seravo-plugin/pull/602) which can handle both of the 2 different cases.

Currently we don't check our substr for failures at all. Eg:
`$diff_txt = file_get_contents(substr($screenshot, 0, -4) . '.diff.txt');`
will evaluate to:
`$diff_txt = file_get_contents('false.diff.txt');`
on PHP 8.0 failure and to
`$diff_txt = file_get_contents('.diff.txt');`
on failure on versions before 8.0.

The compatibility function forces us to check if we had a failure or not.

The error was found by PHPStan but it doesn't complain about it anymore as it thinks we want to have `false.diff.txt` etc.

PR #610, opened 2021-08-10
Ensure exec & substr compatibility

Moppa5:
Ensure that `exec` and `substr` are used properly in Seravo Plugin. I don't think we should call the day off / close those issues yet, but this is a good step forward.

Related to #604 and #606

### DIFF
--- a/src/modules/pages/upkeep.php
+++ b/src/modules/pages/upkeep.php
@@ -417,7 +417,7 @@ class Upkeep extends Toolpage {
    * @return void
    */
   public static function build_seravo_plugin_update_postbox( Component $base, Postbox\Postbox $postbox, $data ) {
-    if ( ! isset($data['current_version']) || ! isset($data['upstream_version']) ) {
+    if ( ! isset($data['current_version']) || ! isset($data['upstream_version']) || $data['upstream_version'] === false ) {
       $base->add_child(Template::error_paragraph(\__('No upstream or current Seravo Plugin version available, please try again later', 'seravo')));
       return;
     }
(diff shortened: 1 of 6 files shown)


src/modules/pages/upkeep.php
```diff
@@ -443,8 +444,10 @@ public static function get_seravo_plugin_update() {
...
     if ( $upstream_version === false || $upstream_version === '' ) {
-      $upstream_version = \exec('curl -s https://api.github.com/repos/seravo/seravo-plugin/tags | grep "name" -m 1 | awk \'{gsub("\"","")}; {gsub(",","")}; {print $2}\'');
-      \set_transient('seravo_plugin_upstream_version', $upstream_version, 10800);
+      $upstream_version = Compatibility::exec('curl -s https://api.github.com/repos/seravo/seravo-plugin/tags | grep "name" -m 1 | awk \'{gsub("\"","")}; {gsub(",","")}; {print $2}\'');
+      if ( $upstream_version !== false ) {
+        \set_transient('seravo_plugin_upstream_version', $upstream_version, 10800);
+      }
     }
 
     $data['upstream_version'] = $upstream_version;
```
JoosuaKoskinen, line 454: The $data['upstream_version'] might be `false` on error, it should be handled in `build_seravo_plugin_update_postbox`.

From: `if ( ! isset($data['current_version']) || ! isset($data['upstream_version']) ) {`
To: `if ( ! isset($data['current_version']) || ! isset($data['upstream_version']) || $data['upstream_version'] === false ) {`

JoosuaKoskinen (approved):

src/modules/pages/upkeep.php
Moppa5: Sure!

PR #611, opened 2021-08-16
Unify module loading and module permission handling

JoosuaKoskinen:

- Adds trait `\Seravo\Module\Module` to unify modules a bit
- Converts most of the modules to use it
- Rewrites the modules, some only a bit and some a bit more
- Renames and restructures same of the modules

Stuff that got left out due to time limitations:
- Dashboard widgets (+ it has rewrite PR open so not good to mess with it)
- canonical-domain-and-https should be converted in to module
- security-restrictions should be converted in to module


Moppa5 (approved):
Looking good to me. Feel free to merge after resolving the CI issue.

Moppa5 (approved):

PR #613, opened 2021-08-17
Rewrite cruftfiles, cruftplugins and cruftthemes

JoosuaKoskinen:

- Rewrite cruftfiles, cruftplugins and cruftthemes


src/modules/pages/security.php
```diff
@@ -224,16 +120,86 @@ public static function init_postboxes( Toolpage $page ) {
...
-    $cruft_files->set_ajax_func(array( __CLASS__, 'list_cruft_files' ));
-
-    $remove_cruft_files = new AjaxHandler('remove-cruft-files');
-    $remove_cruft_files->set_ajax_func(array( __CLASS__, 'remove_cruft_files' ));
-    $cruft_files->add_ajax_handler($remove_cruft_files);
-    $page->register_postbox($cruft_files);
+    $cruft_files2 = self::create_cruft_box('cruftfiles', 'files');
+    $cruft_files2->set_title(\__('Cruft Files', 'seravo'));
+    $cruft_files2->add_paragraph(\__('Find and delete any extraneous and potentially harmful files taking up space in the file system. Note that not everything is necessarily safe to delete.', 'seravo'));
+    $page->register_postbox($cruft_files2);
```
Moppa5, line 126: These most likely shouldn't have the `2` suffix any longer. Note the other as well.

src/lib/cruftremover.php
```diff
@@ -287,6 +287,283 @@ function( $item ) use ( $crufts ) {
...
+      'backup_plugins' => array(
+        'title' => __('Unnecessary Backup Plugins:', 'seravo'),
+        'description' => __('Backups of your website are automatically run on the server on a daily basis. Any plugins creating additional backups are redundant and will unnecessesarily fill up your data storage space.', 'seravo'),
+        'plugins' => array(
+          'updraftplus',
+          'backwpup',
+          'jetpack',
+          'duplicator',
+          'backup',
+          'all-in-one-wp-migration',
```
Moppa5, line 342: This is both in `backup_plugins` and `foolish_plugins` and for some reason when trying to remove it => `Some files failed to remove!` => even though it gets removed. Shouldn't the message string be `Failed to remove some plugins?`

Moppa5 (changes requested):
Looking good! I went through the code rather quickly and tested that the functionality works as expected. There were some minor bugs. 

It can get rather confusing when the delete button and select all feature are overflowing together with the content (see the screenshot). Could this be changed so that only the content overflows and not the selector tools? I think this is issue only with themes and files as it's totally ok for the whole plugins section to overflow. What do you think?

Screenshot(38)

js/cruftremover.js
```diff
@@ -1,189 +1,193 @@
...
-    cruft_view_section.append('<br><button class="button-primary" type="button">' + cruftremover_l10n.delete + '</b>');
-    if ( filecount < 30 ) {
-      section.find('.less-than').hide();
+Seravo.Cruftbox = function(section, type) {
+
+  this.section = section;
+  this.type = type;
+  this.postbox = null;
+  this.cruftArea = null;
+  this.dataFields = null;
```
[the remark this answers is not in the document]
JoosuaKoskinen: The camelCase is because this is the first JS file using WordPress ESLint rules. Couldn't rewrite all JS files with these rules yet but I think all new JS should be written with them.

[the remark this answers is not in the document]
Moppa5: Alright, I thought something like this as well 👍🏼.

src/modules/pages/security.php
JoosuaKoskinen: True, fixed!

src/lib/cruftremover.php
Moppa5: I mean `all-in-one-wp-migration`

JoosuaKoskinen: The message is because I want to use one string for all the cruft removers and plugins/themes are files too. Maybe should change it to `Failed to remove some cruft!`

JoosuaKoskinen: The remove failed because you can't remove the same plugin twice, I removed it from `foolish_plugins`

PR #614, opened 2021-08-18
Update translations

Moppa5:
Update the translations up-to-date

- Remove Rector slashes `\` from translation strings
- Translate strings that were not translated
- Update and fix some translations with up-to-date data


src/lib/postbox/component/template.php
```diff
@@ -46,10 +46,11 @@ public static function text( $content, $class = 'postbox-text', $id = '' ) {
    * Display section title on widget.
    * @param string $content The given title to display.
    * @param string $class   Optional class for the title element.
+   * @param string $id      Optional id for the title element.
```
Moppa5, line 49: At least technical contacts section needed this for the redirect `#contacts` to work properly. Might be useful with other templates as well when we have postboxes with multiple sections.

JoosuaKoskinen (approved):

Issue #624, opened 2021-08-19
Current backups shows error before any backups are made


JoosuaKoskinen:
**Describe the bug**
Only the following is shown when there's no backups yet:
```
Error: Command wp-backup-status 2>&1 failed to execute and returned with status 1. Try running it manually.
```

Should check if `/data/backups/data` exists and show a proper message.

JoosuaKoskinen:
Fixed in #634

PR #628, opened 2021-08-20
Improve current backups error handling (Closes: #624)

Moppa5:
Add proper error handling for both missing directory `/data/backups` or `/data/backups/data` but also include handling for failed command `wp-backup-status`. Also, includes one translated string so it should be good to go 👍🏼 .

#624


JoosuaKoskinen (approved):
LGTM

Issue #615, opened 2021-08-19
Format or filter check passwords output


Moppa5:
**Description**
If command `wp-check-passwords` finds a weak password, the line / user with weak password has red color. However, when running this on Seravo Plugin the output is something like this
`[31mtest_account         Weak password found. Change immediately![0m`

These characters should be formatted or filtered away similarly to `wp-test` output. The check passwords feature on Seravo Plugin is labeled as beta so there might be some other enhancements to do as well.

Issue #625, opened 2021-08-19
wp-check-haveibeenpwned return code 1 in Vagrant


JoosuaKoskinen:
**Describe the bug**
Not exactly Seravo Plugin issue but I think we should remove the `wp-check-haveibeenpwned` return code check from passwords.php as it's useless and currently causing issues.

In production env:
```
PRODUCTION [@joosuakoskinen:~] $ wp-check-haveibeenpwned --json b03fe988f8d55b87dbd2223ee1a8a37a0f882d6e 2>&1
{"found": true}
PRODUCTION [@joosuakoskinen:~] $ $?
-bash: 0: command not found
```

Randomly in development env:
```
DEVELOPMENT [@wordpress:~/wp-content/mu-plugins/seravo-plugin] seravo-plugin-2/master(+1/-0)* 127 ± wp-check-haveibeenpwned --json b03fe988f8d55b87dbd2223ee1a8a37a0f882d6e 2>&1
{"found": true}
DEVELOPMENT [@wordpress:~/wp-content/mu-plugins/seravo-plugin] seravo-plugin-2/master(+1/-0)* 1 ± $?
-bash: 1: command not found
```

JoosuaKoskinen:
Fixed in #629

PR #629, opened 2021-08-20
Enhance password checks (Closes: #615, #625)

Moppa5:

- Format/Filter the wp-check-passwords output
- Remove return code from the wp-check-haveibeenpwned

Mostly related to #615, but because both are related to password checking also #625 minor fix included.

### DIFF
--- a/src/modules/pages/security.php
+++ b/src/modules/pages/security.php
@@ -5,9 +5,10 @@ namespace Seravo\Page;
 use \Seravo\Helpers;
 use \Seravo\CruftRemover;
 
+use \Seravo\Ajax;
 use \Seravo\Ajax\AjaxHandler;
 use \Seravo\Ajax\AjaxResponse;
-
+use \Seravo\Compatibility;
 use \Seravo\Postbox;
 use \Seravo\Postbox\Settings;
 use \Seravo\Postbox\Component;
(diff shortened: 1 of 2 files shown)


src/modules/pages/security.php
```diff
@@ -142,6 +151,24 @@ public static function init_postboxes( Toolpage $page ) {
...
+   */
+  public static function check_passwords() {
+    $check_passwords = Compatibility::exec('wp-check-passwords', $output, $ret_val);
+
+    if ( $check_passwords === false ) {
+      return Ajax\AjaxResponse::command_error_response('wp-check-passwords', $ret_val);
+    }
+    // Filter and format the color chars from output.
+    $pattern = '/\x1b\[[0-9;]*m/';
+    $output = \preg_replace($pattern, '', $output);
```
JoosuaKoskinen, line 166: `preg_replace` returns `null` on error

src/modules/pages/security.php
```diff
@@ -5,9 +5,10 @@
 use \Seravo\Helpers;
 use \Seravo\CruftRemover;
 
+use \Seravo\Ajax;
 use \Seravo\Ajax\AjaxHandler;
 use \Seravo\Ajax\AjaxResponse;
-
+use Seravo\Compatibility;
```
JoosuaKoskinen, line 11: This should be prefixed with `\` like the others

JoosuaKoskinen (changes requested):
Feel free to merge after these changes

src/modules/pages/security.php
Moppa5: Yup. Most likely VSC added this automatically and I forgot to re-check that.

src/modules/pages/security.php
Moppa5: Good point!

Issue #631, opened 2021-08-24
Empty php error log causes "Too many errors to count" on site checks


Moppa5:
**Describe the bug**
I stumbled upon this "Too many errors to count" first when running site checks. Basically `Logs::get_week_error_count()` is not making any difference between abnormally large, missing or empty log file. They are all interpret as an invalid log file and this information is used on site checks to determine that log file has too many errors to count. 

**To Reproduce**
1. Go to `/data/log/`
2. Make sure `php-error.log` is empty (e.g. create new with `touch`)
3. Try to run site checks
4. See "Too many errors to count" message on potential issues

**Expected behavior**
On this case the empty log file should be handled as 0 PHP errors rather than a potential issue. 

**Screenshots**
![Screenshot(29)](https://user-images.githubusercontent.com/16706187/130575018-dc47f233-a009-4dd3-9419-8e40685b84ba.png)

PR #632, opened 2021-08-24
Fix site checks php error count (Closes: #631)

Moppa5:
Fix site checks php error count being "Too many PHP errors to count" when log file is empty. Also includes the following enhancements:

- Add PHP end of life to potential issues
- Translations for the end of life notification
- Replace wp_srintf with sprintf instead

#631


JoosuaKoskinen (approved):
Is php-error.log ever created without writing to it? LGTM either way as extra checks don't hurt.

Issue #626, opened 2021-08-19
Misc issues in development and shadow envs


JoosuaKoskinen:
**Describe the bug**
- Opening security page in dev env logs following:
```
[19-Aug-2021 14:13:00 UTC] PHP Warning:  Undefined array key 0 in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/src/lib/cruftremover.php on line 387
[19-Aug-2021 14:13:00 UTC] PHP Warning:  foreach() argument must be of type array|object, null given in /data/wordpress/htdocs/wp-content/mu-plugins/seravo-plugin/src/lib/cruftremover.php on line 389
```

- I think all postboxes on database page should be shown in development and shadow envs, currently only search-replace postbox is shown. The adminer link must be different in dev env though.

- "Backups" postbox should be hidden in development env as it's just information about automatic backups and which apply in development env.

Moppa5:
Cruftremover seems to have many instances of "blindly executing" something in some path. We should go through the whole `cruftremover.php` file for possible issues. 

As the adminer link is now fetched using the Helpers method, it can quite easily be enhanced so that development gets correct link.

Moppa5:
In addition, I stumbled upon an issue on cruft files listing when running locally on development (docker). At least files `/data/wordpress/.vagrant/mariadb.bak` and `/data/wordpress/.vagrant/backup-dump.sql` cause the cruft files listing to emit 
`Error: Something unexpected happened! Server responded with invalid data.`

`Error:  TypeError: category.cruft.forEach is not a function`
Cruftremover.js on line 95: `category.cruft.forEach((cruft) => {`

I couldn't quite find a solution to this but it seems to be somehow related to method as both of the files had empty mod_date.
```
public static function add_file_information( $file ) {
    if ( $file !== Helpers::sanitize_full_path($file) ) {
      return array(
        'size' => '0B',
        'mod_date' => null,
        'filename' => $file,
      );
    }
```

@JoosuaKoskinen, Do you have any thoughts on this?

JoosuaKoskinen:
Issues mentioned here have been fixed in https://github.com/Seravo/seravo-plugin/tree/seravo-plugin-2/fix-cruftremover. But now found that the confirmation modal doesn't show up. No javascript errors or anything.

JoosuaKoskinen:
> But now found that the confirmation modal doesn't show up. No javascript errors or anything.

[...]

PR #633, opened 2021-08-24
Misc fixes and improvements

Moppa5:
Misc minor fixes and improvements. Remove some excessive lines etc.

- Hide backups info on development
- Make all database postboxes visible on staging & development
- Adminer link when on development

Doesn't include any fixes for cruftremover as it might be better to debug those issues separately.

Related to #626

### DIFF
--- a/src/lib/helpers.php
+++ b/src/lib/helpers.php
@@ -165,7 +165,7 @@ class Helpers {
    * @return string Escaped adminer_url
    */
   public static function adminer_link() {
-    if ( ! self::is_production() ) {
+    if ( self::is_development() ) {
       return \esc_url(\str_replace('//', '//adminer.', \get_site_url()));
     }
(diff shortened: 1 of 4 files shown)


src/lib/helpers.php
```diff
@@ -165,7 +165,7 @@ public static function sanitize_full_path( $file ) {
    * @return string Escaped adminer_url
    */
   public static function adminer_link() {
-    if ( ! self::is_production() ) {
+    if ( self::is_development() ) {
       return \esc_url(\str_replace('//', '//adminer.', \get_site_url()));
     }
     return \esc_url(\get_site_url(null, '.seravo/adminer'));
```
Moppa5, line 171: @JoosuaKoskinen, If the Vagrant works, could you please test this and if it doesn't work, provide a fix? I haven't been quite successful with the Vagrant myself.

JoosuaKoskinen (approved):
Looks good!

Issue #617, opened 2021-08-19
Slack webhook test button not working


JoosuaKoskinen:
**Describe the bug**

The Slack webhook test button doesn't work. Seems to be because JS uses ID `#slack-webhook-test` but the button only has class `.slack-webhook-test`.

JoosuaKoskinen:
Fixed in #634

Issue #623, opened 2021-08-19
WP CLI extensions not working in development env


JoosuaKoskinen:
**Describe the bug**
WP CLI doesn't work in Vagrant and only the following is printed:
```
DEVELOPMENT [@wordpress:~/wp-content/mu-plugins/seravo-plugin] seravo-plugin-2/master 1 ± wp seravo updates
Error: Callable ["Seravo\\Module\\SeravoCLI","updates"] does not exist, and cannot be registered as `wp seravo updates`.
```

The updates command shouldn't be registered at all on development env.

JoosuaKoskinen:
Fixed in #634

Issue #619, opened 2021-08-19
Customer managed domain zones can't be sniffed


JoosuaKoskinen:
**Describe the bug**
On domains page, it used be possible to sniff zone of a domain if it's managed by customer. That doesn't work anymore on finnish sites because Seravo Plugin uses `$domain['management'] === 'Customer'` check but `$domain['management']` is translated as `'Asiakas'` earlier. This bug has probably existed for a while.

JoosuaKoskinen:
Fixed in #634

PR #634, opened 2021-08-24
More misc fixes

JoosuaKoskinen:
Closes: #617, #624, #623, #619

- Fixes Slack webhook
- Fixes wp-backup-status ran before first backup
- Removes WP CLI 'seravo updates' from dev and staging envs.
- Fix zone sniff on finnish sites.

### DIFF
--- a/src/modules/pages/backups.php
+++ b/src/modules/pages/backups.php
@@ -142,7 +142,7 @@ class Backups extends Toolpage {
    * @return Ajax\AjaxResponse
    */
   public static function get_backups_status() {
-    if ( ! \file_exists('/data/backups') || ! \file_exists('/data/backups/data') ) {
+    if ( ! \file_exists('/data/backups') || ! \file_exists('/data/backups/data') || ! \file_exists('/data/backups/exclude.filelist') ) {
       return Ajax\AjaxResponse::error_response(__('No backups data found under <code>/data/backups/</code>', 'seravo'));
     }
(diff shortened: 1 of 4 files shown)


src/modules/pages/backups.php
```diff
@@ -133,7 +133,7 @@ function ( Component $base, Postbox\Postbox $postbox ) {
    * @return Ajax\AjaxResponse
    */
   public static function get_backups_status() {
-    if ( ! \file_exists('/data/backups') || ! \file_exists('/data/backups/data') ) {
+    if ( ! \file_exists('/data/backups') || ! \file_exists('/data/backups/data') || ! \file_exists('/data/backups/exclude.filelist') ) {
```
Moppa5, line 136: I'm not sure whether the missing `exclude.filelist` should cause any issues? It seems that `wp-backup-status` just outputs that the file is missing. Did you run into problems with this?
```
--> Comparing /data to latest backup...
cat: /data/backups/exclude.filelist: No such file or directory
...
```

Moppa5 (approved):
Looking good! I had one comment about the `exclude.filelist`, but if you have had issues with it missing then I think we are good to go.

src/modules/pages/backups.php
JoosuaKoskinen: `exclude.filelist` is created on first `wp-backup`. If it doesn't exists, backups neither do

Moppa5: Well alright. Still find it rather excessive as we are already checking here whether the backups exist at all. However, this is no big deal and we can stick with that if you want to.

Issue #635, opened 2021-08-24
Fix default postbox positioning


JoosuaKoskinen:
**Describe the bug**
The default postbox positions, on a new site be before user has moved them, is not good. The postboxes should be put evenly in all columns so full page width is used.

Security page: First column is double the height of second column (both have 3 postboxes but one is huge) Only two columns are used, use 4 if possible (instructions below).

Database page: All postboxes are in the first column, rest of them are empty. At least two columns should be used but 4 would be the best if it's possible.

Upkeep page: The page has two columns but only the first one is used by default.

Site status page: Two of the postboxes are in the second column, the rest are in the first column. Should be balanced and use all 4 if possible.

**Instructions**
A page can be reset to initial order with `wp user meta delete joosua seravo-postbox-order_tools_page_security_page`.
The position parameter can be given in `\Seravo\Postbox\Postbox` constructor. Eg. `new Postbox('somebox', 'side');`

The position system is part of the postbox handler which hasn't been rewritten yet and is implemented bad. It should work anyways with position values `normal` (1st column) and `side` (2nd column). The values `column3` and `column4` do exists but not sure if they work. They should be tried anyways.

PR #638, opened 2021-08-30
Fix default postbox positioning (Closes: #635)

Moppa5:
Fix the default positioning of the postboxes. Use `column3` and `column4` when appropriate and "balance" the postboxes overall.

#635


JoosuaKoskinen (approved):
Looks much better! There's only one issue, four columns wide pages are converted as two wide on smaller displays and `side`, `column3` and `column4` are all on the right side. Will fix that when postbox handler is fixed later.

PR #640, opened 2021-09-10
Create separate vendor dir for dev requirements

JoosuaKoskinen:

* Set `vendor-dev` directory as the default composer vendor path. See next section for reasoning.
* Update 'svg-sanitize' and 'composer/installer' as they were quite old and new Seravo-Plugin release is coming.
* Disable composer `platform_check.php` because it:
   - Increases overhead of every page load.
   - Is useless to us, Seravo-Plugin must support all PHP versions and modules available in our environment.
   - Could break a site. Very unlikely but it's literally designed to `die()`.
* Makes the scripts in composer.json bit prettier and more flexible.

Seravo-Plugin has `vendor` dir commited so `wp-seravo-plugin-update` doesn't have to run `composer install` and we can be sure the package versions are correct, working and secure. But composer is also used to manage all the development tools required.

Normally you would have `vendor` in `.gitignore` but we can't do that. We also can't just `.gitignore` dev packages only because `vendor` has bunch of common composer files which are commited and change on every composer operation.

This was real pain in Seravo-Plugin workflow as any changes, git rebases, commits, local tests, etc other actions required installing and uninstalling of dev tools, all kinds of playing with git and `git status` was always huge list.  Also no tools could determine what actually exists in production environment and this already caused some bugs like usage of `hrtime()` from PHP 7.3 and `str_contains()` from PHP 8.0 as they were defined by some dev packages.

Default vendor dir is now `vendor-dev` and the real vendor dir commited and used in production is `vendor`. The `vendor-dev` is .gitignored and Seravo-Plugin can now be developed without hassle.

The two vendor dirs is not really the "composer way of doing things" but neither is commiting vendor dir. This should work until nicer solution is possible. 

Code review isn't really needed as not real changes there. Diff is mostly the normal composer file changes from updating packages.


sjaks (approved):
Tested and works :+1: Do you think if it'd be good to mention the use of two vendor dirs in the README somewhere? :)

PR #641, opened 2021-09-10
Fix hrtime() PHP error and small misc mistakes

JoosuaKoskinen:

- Install polyfill-php73:
    - Seravo-Plugin uses `hrtime()` which is not something all sites have (see #640). This package adds it for us.
- Fix misc linter errors:
     - Couple of `@return` phpdoc comments
     - A few useless escapes in regexes
     - Bit of invalid HTML
     - Add PHP extensions to composer requirements, doesn't really affect anything but it's always to keep track of what Seravo-Plugin requires to work even if our environment always has them.

The first commit is #640 which this PR depends on. It should be merged first. 

Again, about 10 real line changes, other are from installing composer package and the first commit.


sjaks:
Lint fixes look alright and the inclusion of polyfill fixes the issue. This doesn't pass CI since additional packages are required. Otherwise, LGTM!

PR #643, opened 2021-09-14
Fix bugs found on production sites

JoosuaKoskinen:

- Limit HTTP report count on site status page
- Only load Seravo postbox scripts/styles on pages they are needed (tool pages and partly dashboard)


sjaks (approved):
Only reviewed code. If the script and styles assets actually are loaded at the right moment after this fix, then feel free to merge :)

PR #645, opened 2021-09-24
Rewrite disk usage fetching

JoosuaKoskinen:

The old design ran 'du /data' every time the 15min/60min transient expired.
The command can take up to a minute causing a page to load really slow.

New design stores the last known disk size in a non-expiring transient.
When the main transient expires, it takes the last know size and
runs a background command to update the non-expiring transient from disk.

Found on a production site.

### DIFF
--- a/src/lib/shell.php
+++ b/src/lib/shell.php
@@ -58,7 +58,7 @@ class Shell {
    * @param string $command Command to run.
    * @return mixed|bool Pid of the program.
    */
-  public static function backround_command( $command ) {
+  public static function background_command( $command ) {
     $output = array();
     \exec('(' . $command . ') > /dev/null & echo $!', $output);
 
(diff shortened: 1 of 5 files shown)


sjaks (approved):
Looks good! Have you tested how this works on a new site? Are widgets generated fluently on the first access when `get_disk_space_usage` is run for the first time?

JoosuaKoskinen:
> Have you tested how this works on a new site? Are widgets generated fluently on the first access when `get_disk_space_usage` is run for the first time?

If the data doesn't exist yet, it's fetched immediately:
https://github.com/Seravo/seravo-plugin/blob/eb6184320bea0a0d482f14609628bccaa39e812f/src/modules/dashboard-widgets.php#L192-L195

+ New change: Renamed function `backround_command` to `background_command`.

PR #646, opened 2021-10-02
Make Composer scan all files in src/

szepeviktor:

Simplify Composer autoloader's configuration.

@JoosuaKoskinen

Big win! 🎉

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -822,11 +822,7 @@
     "prefer-stable": false,
     "prefer-lowest": false,
     "platform": {
-        "php": "^5.6 || ^7.0 || ^8.0",
-        "ext-curl": "*",
-        "ext-fileinfo": "*",
-        "ext-json": "*",
-        "ext-zlib": "*"
```
szepeviktor, line 829: I do not know where it came from...

PR #649, opened 2021-10-02
Improve Composer config

szepeviktor:
https://getcomposer.org/doc/articles/scripts.md#referencing-scripts

Composer knows its scripts :)
Please see the diff.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -76,18 +76,18 @@
     ]
   },
   "scripts": {
-    "phpcs:check": "@php ./vendor-dev/bin/phpcs -n",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -76,18 +76,18 @@
     ]
   },
   "scripts": {
-    "phpcs:check": "@php ./vendor-dev/bin/phpcs -n",
+    "phpcs:check": "phpcs -n",
     "phpcs:fix": "@php ./vendor-dev/bin/phpcbf -n || true",
```
szepeviktor, line 80: Here we have shell scripting.

PR #653, opened 2021-12-21
Parse PHP versions from /etc/php/versions

JoosuaKoskinen:

Automate PHP version recommendations and warnings by parsing the default and EOL versions from `/etc/php/versions`.

The PHP upgrade postbox is still hard coded as we only want to show specific versions there so needs a bit more thinking.

#### Manual testing steps?
- Set a site to use PHP `7.3`
- Login to the site (notifications only shown right after login)
- See a banner at the top recommending `7.4` or `8.0` (depends on `PHP_DEFAULT_VERSION`)
- See "PHP warnings and errors" postbox below warning about `7.3` being EOL
- Upgrade PHP and see these messages disappear

### DIFF
--- a/src/lib/helpers.php
+++ b/src/lib/helpers.php
@@ -89,7 +89,15 @@ class Helpers {
    * @return string Recommended PHP version.
    */
   public static function get_recommended_php_version() {
-    return '7.4';
+    $php_config = self::parse_php_config();
+
+    if ( $php_config === array() || ! isset($php_config['PHP_DEFAULT_VERSION']) ) {
+      // Couldn't parse the config. Return '0.0' which should hide any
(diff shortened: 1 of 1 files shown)


sjaks (commented):
Code looks good to me. I wasn't able to get the notification to pop up.
1. I checkout to this branch on my testing site
2. Switched to PHP 7.3
3. Logged out and back in
4. No banner was seen

Wiljami (approved):
Code looks good and tested using my own site. It functioned as described in the initial post:
Changed to 7.3 and got following banner:
image
Changed to 8.0 and the banner would no longer appear when logging in.

sjaks (approved):

JoosuaKoskinen:
Merged in https://github.com/Seravo/seravo-plugin/pull/665.

PR #655, opened 2022-02-14
Make it possible to disable the plugin using a constant or an environment variable

joppuyo:
Running the Seravo MU Plugin in a local development environment can cause various issues since the plugin will try to write log files into a location that might not exist in the development environment.

This PR introduces a PHP constant and an environment variable called SERAVO_PLUGIN_DISABLED which can be used to conditionally disable the plugins in environments where it's not desirable to have the plugin enabled.

This means you can add the following to your `wp-config.php` to disable the plugin:

```php
define('SERAVO_PLUGIN_DISABLED', true);
```

You can also define the same as environment variable for the same effect

```env
SERAVO_PLUGIN_DISABLED='true'
```

### DIFF
--- a/seravo-plugin.php
+++ b/seravo-plugin.php
@@ -80,6 +80,15 @@ class Loader {
   private static $domain = 'seravo';
 
   public function __construct() {
+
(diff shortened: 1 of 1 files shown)


Jaska:
How about this change @simoke ?

PR #661, opened 2022-08-30
Timeout disk space usage exec after 3 seconds

JoosuaKoskinen:
Temporary fix to prevent disk space usage calculations from taking too long.

It now timeouts after 3 seconds and on failure, disables the feature until transients are cleared.

### DIFF
--- a/src/modules/dashboard-widgets.php
+++ b/src/modules/dashboard-widgets.php
@@ -193,14 +193,10 @@ class DashboardWidgets {
         // Last known disk space usage transient doesn't exist either. This shouldn't happen unless object-cache
         // was emptied (if using one) or the site is new. We are forced to fetch the data disk usage now.
         $data_folder = array();
-        $exec = Compatibility::exec('du -sb /data ' . \implode(' ', $exclude_dirs), $data_folder);
+        $exec = Compatibility::exec('timeout 3 du -sb /data ' . \implode(' ', $exclude_dirs), $data_folder);
         if ( $exec === false || $data_folder === array() ) {
           // Couldn't get the disk usage
-          return array(
-            'relative_usage' => 0.0,
-            'disk_usage' => 0,
-            'plan_limit' => 0,
-          );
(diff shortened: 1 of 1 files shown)


simoke (approved):
Tested on my own site. First, modified `du` to sleep 120 seconds before executing anything:
```
PRODUCTION [root@simokemppainen-ecce6a:/usr/bin] 130 # ls -l du*
-rwxrwxr-x 1 root root     35 Aug 31 15:38 du
-rwxr-xr-x 1 root root 100568 Jan 18  2018 du.real
PRODUCTION [root@simokemppainen-ecce6a:/usr/bin] # cat du
#!/bin/bash
sleep 120
du.real "$@"
```
The tried to log in -> timeout, 504 error (as expected).

Uploaded the fixed version of `dashboard-widgets.php` and tried again -> worked! The only "bad" behaviour is that in Site status view, the spinner will spin forever without giving any messages to user.
image


However, acute problem is fixed.

simoke:
In addition, after a few minutes, got the message:
image

JoosuaKoskinen:
> Tested on my own site. First, modified `du` to sleep 120 seconds before executing anything:
> 
> ```
> PRODUCTION [root@simokemppainen-ecce6a:/usr/bin] 130 # ls -l du*
> -rwxrwxr-x 1 root root     35 Aug 31 15:38 du
> -rwxr-xr-x 1 root root 100568 Jan 18  2018 du.real
> PRODUCTION [root@simokemppainen-ecce6a:/usr/bin] # cat du
> #!/bin/bash
> sleep 120
> du.real "$@"
> ```
> 
> The tried to log in -> timeout, 504 error (as expected).
> 
> Uploaded the fixed version of `dashboard-widgets.php` and tried again -> worked! The only "bad" behaviour is that in Site status view, the spinner will spin forever without giving any messages to user. image
> 
> However, acute problem is fixed.

simoke:
I suppose that this can be merged :)

PR #665, opened 2022-11-25
Parse PHP versions from local file

JoosuaKoskinen:
Parse available, recommended and EOL versions from /etc/php/versions.

The code is mostly same as in https://github.com/Seravo/seravo-plugin/pull/653 so the review from there should so OK here too. The EOL notification should be reviewed again.

### DIFF
--- a/src/lib/helpers.php
+++ b/src/lib/helpers.php
@@ -85,22 +85,6 @@ class Helpers {
   }
 
   /**
-   * Get the recommended PHP version.
-   * @return string Recommended PHP version.
-   */
-  public static function get_recommended_php_version() {
-    return '7.4';
-  }
-
-  /**
-   * Get the current end of life PHP version.
-   * @return string Current end of life PHP version.
-   */
-  public static function get_eol_php_version() {
-    return '7.2.34';
-  }
-
-  /**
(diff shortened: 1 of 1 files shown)


Wiljami:
Tested this on a fresh site. Wp-admin panel would end up just displaying only:
```
"7.3"
```
on the screen (it being the only source code as well) with this seravo-plugin change.

Wiljami:
Tested this on a fresh site. Wp-admin panel would end up just displaying only:
```
"7.3"
```
on the screen (it being the only source code as well) with this seravo-plugin change.

JoosuaKoskinen:
> Tested this on a fresh site. Wp-admin panel would end up just displaying only:
> 
> ```
> "7.3"
> ```
> 
> on the screen (it being the only source code as well) with this seravo-plugin change.

Forgot some debug code, fixed now.

Wiljami (dismissed):
Tested.

`get_recommended_php_version` returns 8.0 and `get_eol_php_version` returns 7.3

Which looks good to me and with the above fix, the site is no longer broken due to debug.

JoosuaKoskinen:
@Wiljami can you approve again. Nothing in code has changed, just rebased.

PR #671, opened 2023-04-13
Fix PHP version changing

JoosuaKoskinen:
All hard coded PHP version numbers have been removed. The latest versions are now on the list. Version with EOL date passed is not listed unless it's in use. The EOL date is now translated. EOL date is shown only if it has passed or is in less than 12 months. 

Old:
image

New:
image


frimro (dismissed):
Can confirm that 8.1 and 8.2 are now available.

image

frimro (approved):

frimro (approved):

PR #675, opened 2023-05-03
Reset shadows using the Container API

JoosuaKoskinen:
Instead of the Seravo Plugin's built-in poller, use Container API's `/task/` and `/shadow-reset/` endpoints to do it.

To be fixed at later time:
- Trying to reset an shadow (eg. from an other browser tab) while another shadow reset is in progress, will show a very generic error message to the user.
- The `SWD` and `Container` API classes are ugly, will replace the with some common `HTTPClient` system later.

#### Manual testing steps?
- Have a site running with a Container API and at least one shadow site.
- Try breaking the "Shadows" (Varjot) feature on the site status (Sivuston tila) page.


frimro (approved):
Works as expected.

PR #679, opened 2023-05-04
Fix the disk usage calculations and clarify the disk info 

JoosuaKoskinen:
- Fixed the usage calculation for the chart (notice how the chart in old image is at 100% and not 93%).
- Added disk usage percentage to make it clear which color in the chart is the usage.
- Changed "Levytilaa käytössä" to "Levytilaa käytetty" so it's not confused as the disk limit.
- Translated all the untranslated strings to Finnish.

Customer feedback.

#### Screenshots
Old:
image

New:
image


frimro (approved):
Looks much better now!

PR #686, opened 2023-05-17
Add Container API client

JoosuaKoskinen:

Adds the base for Container API Client and the possibility to run background tasks with polling.

Container API is not used anywhere yet.

### DIFF
--- a/src/lib/ajax/handler.php
+++ b/src/lib/ajax/handler.php
@@ -2,6 +2,7 @@
 
 namespace Seravo\Ajax;
 
+use Seravo\API\Container;
 use \Seravo\Postbox\Component;
 
 /**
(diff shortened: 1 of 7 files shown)


src/lib/ajax/handler.php
```diff
@@ -213,22 +214,68 @@ public function get_cache_time() {
...
+
+    // Has the task completed?
+    if ( $response['status'] === 'completed' ) {
+      return true;
+    }
+
+    // Is the task still running?
+    if ( $response['status'] === 'started' || $response['status'] === 'created' ) {
+      return AjaxResponse::require_polling_response($id, 'task');
+    }
```
frimro, line 268: There's also `scheduled` status, which is after the container API has successfully scheduled the task to the task manager. So I guess you could add three of them to an array and check is `$response['status']` one of them.

JoosuaKoskinen: Fixed as you suggested.

frimro (approved):
Tested this with both #681 #685, using a container image without the latest features and one with them. Works as expected.

PR #689, opened 2023-05-30
Allow object-cache.php to have a custom path

JoosuaKoskinen:
Instead of assuming `object-cache.php` to be under `/data/wordpress/htdocs/wp-content/`, use `WP_CONTENT_DIR` constant to determine to correct path.

#### Manual testing steps?
- Have a custom path for `wp-content` dir.
- Navigate to dashboard > tools > site status.
- In the caching postbox, see that object-cache is OK.
- Run `mv object-cache.php object-cache.php.off`
- Install object-cache back with the install button.
- See that object-cache.php is installed and the `.off` file removed.

### DIFF
--- a/src/modules/pages/sitestatus.php
+++ b/src/modules/pages/sitestatus.php
@@ -39,11 +39,6 @@ class SiteStatus extends Toolpage {
   const IMAGE_MIN_SIZE = 500;
 
   /**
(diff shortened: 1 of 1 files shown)


ypcs (approved):
Tested, works as expected.

frimro (approved):

PR #705, opened 2023-07-20
Log all API errors

ypcs:

This makes it much easier to trace why certain features have failed.

See: <https://www.php.net/manual/en/function.error-log.php>

### DIFF
--- a/src/lib/api/container.php
+++ b/src/lib/api/container.php
@@ -105,7 +105,9 @@ class Container {
 
     // Only accept HTTP status codes indicating success.
     if ( $status < 200 || $status >= 300 ) {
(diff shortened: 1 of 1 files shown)


frimro (commented):
We should probably pass the response contents themselves to the error function, as it now passes the HTTP status only. The actual error can be obtained from `detail` in the response JSON.

```console
[20-Jul-2023 07:55:34 UTC] Container API error on post to '/wordpress/php-compatibility-check/' failed: HTTP status code 400
```

frimro (approved):
Timestamp is probably redundant, but can confirm that the API error is now logged correctly

PR #707, opened 2024-04-12
Add WP Start to site plans in Site Status

pekkakortelainen:
This modifies the array of site plans used on Site Status page to include the new WP Start plan.


slagsbulte (approved):
Code change looks good to me. We might also want to add Dev plan there, but I guess it's not required necessary just now.

PR #710, opened 2024-08-15
Fix API error with php_compatibility_check

frimro:
- Fix API error with php_compatibility_check
- Bump version to 1.10.2

Currently SP sends bad JSON if no version is specified, resulting in API error.


ypcs (approved):
Looks good.

PR #713, opened 2025-04-28
Fix 'Call to undefined function Seravo\rmdir_recursive()' error

mhgx:
Add missing `self::` to rmdir_recursive() call.


simoke (approved):

frimro:
Thank you for the contribution and fix!

frimro:
This is available in 1.10.4

